PR #18, opened 2015-09-30
Intialise itemStore with an action

flashbackzoo:
@assertchris 

Based on feedback from @unclecheese


assertchris:
looking good, much cleaner!

PR #21, opened 2015-10-02
More minor fixes

unclecheese:
* Fix `Folder::find_or_make` bug where root `assets/` folder was not being removed.
* Tidy up JS functions


code/AssetGalleryField.php
```diff
@@ -216,7 +216,7 @@ protected function getData($filters = array()) {
 	 */
 	protected function getFolder($folder = null) {
 		if ($folder) {
-			return Folder::find_or_make($folder);
+			return Folder::find_or_make(preg_replace('/^'.ASSETS_DIR.'\//', '', $folder));
```
assertchris: Would something like `str_ireplace` be simpler here?

public/src/store/editorStore.js
```diff
@@ -6,9 +6,7 @@ import CONSTANTS from '../constants';
 var _fields = [];
 
 function create(data) {
-	var fieldExists = _fields.filter((field) => { return field.name === data.name; }).length > 0;
-
-	if (fieldExists) {
+	if(_fields.some(field => field.name === data.name)) {
```
assertchris: <3

assertchris:
Oh! I just read the description of this. I'm not sure we want the assets path removed - that makes it impossible to browse from that folder, which means all uploads need to be in a subfolder. @tractorcow opinions?

tractorcow:
Yep, we don't want any references to assets at all, really.

`Folder::find_or_make` will trust any data it's given. Case in point, it should be perfectly valid to create an `assets/assets` folder.

tractorcow:
Also, the "Filename" property of File DataObject will no longer contain the assets folder anyway, so even fewer reasons to care about it. ;D

PR #22, opened 2015-10-02
Numerous improvements to reduce verbosity

unclecheese:
This PR introduces several changes to the Flux implementation that reduces verbosity, most notably through less reliance on constants and redundant boilerplate functions.

* New `ActionSet` class allows you to create actions declaratively and assign a listener, eliminating the need to explicitly invoke `Dispatcher.dispatch()` in every single action. Example:

```js
myActions.createAction('create', data => {
    // This will be called before any stores are notified. Good time to start fire an XHR.
});

// The store will call this in response to the action: 
MyStore.onCreate(data) // on{ActionName}
```

You can create asynchronous actions using `createAsyncAction`. The action you create will automatically get a `.completed` function that stores can listen to independently.

```js
myActions.createAsyncAction('create', data => {
    runAjaxRequest(data, myActions.create.completed);
});

// The store will call this in response to the action: 
MyStore.onCreate(data);
MyStore.onCreateComplted(serverResponse);
```

* No more XHRs in stores. Only in actions. Now multiple stores can respond to the outcome of a server call
* Created layer of abstraction for the API (api/API.js)
* Create a `BaseStore` class that is `ActionSet` aware, so dispatcher registrations become trivial, and responses to actions are more opinionated, `onActionName` and `onActionNameCompleted`.
* Heaps of minor JavaScript syntactical changes

### DIFF
--- a/code/AssetGalleryField.php
+++ b/code/AssetGalleryField.php
@@ -162,6 +162,7 @@ class AssetGalleryField extends FormField {
 		$count = 0;
 
 		if($folder->hasChildren()) {
+
 			/** @var File[]|SS_List $files */
 			$files = $folder->myChildren();
 
@@ -216,7 +217,7 @@ class AssetGalleryField extends FormField {
 	 */
 	protected function getFolder($folder = null) {
 		if ($folder) {
-			return Folder::find_or_make($folder);
+			return Folder::find_or_make(preg_replace('/^'.ASSETS_DIR.'\//', '', $folder));
 		}
 
 		$path = $this->config()->defaultPath;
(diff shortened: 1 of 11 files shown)


code/AssetGalleryField.php
```diff
@@ -216,7 +217,7 @@ protected function getData($filters = array()) {
 	 */
 	protected function getFolder($folder = null) {
 		if ($folder) {
-			return Folder::find_or_make($folder);
+			return Folder::find_or_make(preg_replace('/^'.ASSETS_DIR.'\//', '', $folder));
```
assertchris: This change is duplicated in https://github.com/open-sausages/silverstripe-assets-gallery/pull/21...

assertchris:
https://github.com/open-sausages/silverstripe-assets-gallery/pull/26

PR #26, opened 2015-10-08
Simplified everything

assertchris:
@flashbackzoo, @scott1702, @unclecheese 

How about this?


flashbackzoo:
Hey @assertchris 

One general comment and a few specific things...

### General thing

#### Architecture

Your refactor makes the code less verbose - it did seem like overkill having Flux there for a reasonably simple application. However it has made the life-cycle (input -> update store -> re-render) less obvious to the reader (well - to me anyway :smile: ). You have to explicitly pass the store down into components, then emit events on that store, which listens to itself (which seems...weird). It takes a fair bit of digging around the codebase to figure out what's going on. One advantage of using Flux is it's an established pattern, there are lots of examples available online, which makes the architecture much easier for newcomers. There's a lot less project specific knowledge required.

In saying that - I think it's more important at this stage to get the feature finished than to go back and forward refactoring the architecture. What you've done makes a lot of sense, it just takes a bit of time to figure out what's going on, so I'm happy to merge it from that perspective. Nice work! :smile: 

### Specific things

#### FileStore.addListeners()

Can we call this inside `FileStore.create` so you don't have to remember to call it when creating a new `FileStore`?

#### Order or arguments when extending objects

See https://github.com/open-sausages/silverstripe-assets-gallery/commit/5a83bb37fb9766bef4183f781db94f83d3a0c130#diff-040679c2d855f91a8f2e0e5ed17124b2R125

Currently you can't override defaults. `data.limit` would be overridden by `defaults.limit`. Is that intended?

#### Removing store listeners

I _think_ we should be using [emitter.removeAllListeners](https://nodejs.org/api/events.html#events_emitter_removealllisteners_event) in [FileStore.removeEventListeners](https://github.com/open-sausages/silverstripe-assets-gallery/commit/5a83bb37fb9766bef4183f781db94f83d3a0c130#diff-040679c2d855f91a8f2e0e5ed17124b2R31). [emitter.removeListeners](https://nodejs.org/api/events.html#events_emitter_removelistener_event_listener) will remove a max of 1 listener. So if `FileStore.addListeners` has been called multiple times, there will still be listeners hanging around.

There are [listeners added by the gallery component](https://github.com/open-sausages/silverstripe-assets-gallery/commit/5a83bb37fb9766bef4183f781db94f83d3a0c130#diff-740508765e3ec6774dd8142a6cc22e88R21). Should these be removed when the component is unmounted?

assertchris:
@flashbackzoo 

> FileStore.addListeners()

Yeah, moved it into the constructor and removed from `main.js`.

> Order or arguments when extending objects

Good catch!

> Removing store listeners

Not sure how to do that well, though it does need to be done... Lemme try a thing.

assertchris:
@flashbackzoo how about https://github.com/open-sausages/silverstripe-assets-gallery/commit/f30b5450d1c538dd7299d4fc64442ba042c7f29e?

flashbackzoo:
Looks good!

PR #29, opened 2015-10-12
Fix scrollbar on galleryview when there are no images

scott1702:
thumbnail images now cover
fix to 'load more' breaking when deleting images
@assertchris


scott1702:
Added styles specific to the insert-media dialog based on https://github.com/open-sausages/silverstripe-framework/tree/insert-media-gallery

scott1702:
Rebased @assertchris

PR #42, opened 2015-10-15
Add accessibility for item actions

scott1702:
:wrench:Code cleanup


flashbackzoo:
Looking good. There's no way to tab through the files and drill down into a folder. Maybe it should go...

tab -> file get focus
tab -> delete options get focus
tab -> edit gets focused

scott1702:
Updated based on feedback

public/src/component/file-component.js
```diff
@@ -46,29 +52,64 @@ class FileComponent extends BaseComponent {
 		return thumbnailClassNames;
 	}
 
+	getItemClassNames() {
+		var itemClassNames = 'item ' + this.props.category;
+
+		itemClassNames += this.state.focussed ? ' focussed' : '';
```
assertchris: Please expand this to an if statement - that's consistent with the rest of the js...

public/src/component/file-component.js
```diff
@@ -46,29 +52,64 @@ class FileComponent extends BaseComponent {
...
+	handleKeyDown(event) {
+		event.stopPropagation();
+
+		//if event doesn't come from the root element, do nothing
+		if (event.target !== React.findDOMNode(this)) {
+			return;
+		}
+
+		//If space or enter is pressed
+		if (event.keyCode === 32 || event.keyCode === 13) {
```
assertchris: seems this would be better as `<FileComponent selectKeys={[32, 13]} ... />`

public/src/component/file-component.js
```diff
@@ -46,29 +52,64 @@ class FileComponent extends BaseComponent {
...
+
+	handleBlur() {
+		this.setState({
+			'focussed': false
+		})
 	}
 
 	render() {
-		return <div className={this.getItemClassNames()} data-id={this.props.id} onClick={this.onFileNavigate}>
+		return <div className={this.getItemClassNames()} data-id={this.props.id} tabIndex="0" onClick={this.onFileNavigate.bind(this)} onKeyDown={this.handleKeyDown.bind(this)}>
```
assertchris: The improvements PR removed `fn.bin(this)` from render. Those binds make new references each time the component is rendered (which can happen often). Use `this.bind(['fn', ...])` in the constructor and then `onClick={this.fn}` will already be bound (as seen [here](https://github.com/open-sausages/silverstripe-assets-gallery/blob/master/public/src/component/file-component.js#L10) and [here](https://github.com/open-sausages/silverstripe-assets-gallery/blob/master/public/src/component/file-component.js#L66))...

public/src/component/file-component.js
```diff
@@ -46,29 +52,64 @@ class FileComponent extends BaseComponent {
...
 			<div className={this.getThumbnailClassNames()} style={this.getThumbnailStyles()}>
 				<div className='item__actions'>
 					<button
 						className='item__actions__action item__actions__action--remove [ font-icon-trash ]'
 						type='button'
-						onClick={this.onFileDelete}>
+						onClick={this.onFileDelete.bind(this)}
+						onFocus={this.handleFocus.bind(this)}
+						onBlur={this.handleBlur.bind(this)}>
```
assertchris: as above...

PR #50, opened 2015-10-20
API AssetGalleryField hooks into AssetAdmin search

tractorcow:
See https://github.com/open-sausages/silverstripe-cms/pull/3 for CMS change.

The frontend code will need to be updated; When running the 'search' ajax command, you'll need to either discard or include the `?q=` querystring.

This update ensures that search results are applied correctly to the asset gallery view.

Rather than re-implementing the entire search logic in AssetGalleryField, pre-filtered search results are expected to be injected (yay DI).

Removed all the redundant re-implementation of this search logic.

### DIFF
--- a/code/AssetGalleryField.php
+++ b/code/AssetGalleryField.php
@@ -13,6 +13,44 @@ use SS_HTTPResponse;
 use SS_List;
 
 class AssetGalleryField extends FormField {
+
+	/**
+	 * @var SS_List
+	 */
+	protected $source = null;
+
+	/**
+	 * Create a new AssetGalleryField
+	 *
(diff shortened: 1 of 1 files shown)


assertchris:
Meant to hit comment.

I don't think it's a good idea to merge this without the client side code being fixed and tested.

tractorcow:
Yeah, let's close this and get @flashbackzoo to fix the frontend code before we re-open.

flashbackzoo:
Yup - I can make the changes.

assertchris:
This change does couple the new field with CMS severely. As in no hope of using it with just the ORM, in a non ss project. Don't like that part of it...

PR #56, opened 2015-10-28
Fix double highlight on selection

scott1702:
Before 
image

After 
image


assertchris:
This doesn't seem like a good idea. Shouldn't we rather try to style the focus to something less obvious instead of preventing it altogether?

PR #61, opened 2015-11-09
Link GridField to GalleryField

scott1702:
This PR also assumes the changes at https://github.com/open-sausages/silverstripe-framework/tree/insert-media-gallery and https://github.com/open-sausages/silverstripe-cms/tree/integrate-asset-gallery-view

This PR is a series of commits which take advantage of the new cmsEvents introduced in https://github.com/open-sausages/silverstripe-reactjs-common/pull/4


flashbackzoo:
This is no longer required because the GalleryField has grown up and become a CMS section.

PR #65, opened 2015-11-24
File edit callback

chillu:
Required for operation of the component within AssetAdmin.
Can't be handled by events since it needs to replace behaviour rather than augment it.
The callback shouldn't be used by other implementations, it's likely going to be replaced
by React subclasses or a more integrated approach to state changes (shared Flux store between list and gallery view)


chillu:
Required by https://github.com/open-sausages/silverstripe-cms/commit/8009b20d4e0d0b1d929b3ed50bfe976497b700b3

PR #67, opened 2015-11-25
Add notice when there are no files in a folder

scott1702:
image


public/src/component/gallery-component.js
```diff
@@ -177,6 +177,14 @@ export default class extends BaseComponent {
 		$select.change(() => React.addons.TestUtils.Simulate.click($select.find(':selected')[0]));
 	}
 
+	getNoItemsNotice() {
+		if (this.state.count < 1) {
+			return <p className="no-item-notice">No items found</p>;
```
chillu: Needs to be localised (through `i18n.js`)

PR #77, opened 2016-02-08
Re-enabled CI

chillu:
Requires some weird Travis hacks, but should serve us well for the next month or so until the underlying `framework` and `cms` changes required by this new module can be merged back.

See https://github.com/silverstripe-labs/silverstripe-travis-support/pull/27

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,26 @@
 language: php
 
-php:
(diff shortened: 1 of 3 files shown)


chillu:
Whee, tests passing now - @flashbackzoo

PR #80, opened 2016-02-15
Enforce permissions

chillu:


code/AssetGalleryField.php
```diff
@@ -46,6 +53,54 @@ class AssetGalleryField extends FormField {
...
+
+	/**
+	 * @param string $name
+	 * @param string $title
+	 * @param SS_List $dataList
+	 */
+	public function __construct($name, $title = null, SS_List $dataList = null) {
+		parent::__construct($name, $title, null);
+
+		if($dataList && !$dataList->dataClass() instanceof File) {
```
tractorcow: `->dataClass()` returns a string, which is never an `instanceof File`, eh?

tractorcow: Also you might want an `$dataList instanceof DataList`, since `SS_List` interface doesn't promise a `dataClass()` method. :)

code/AssetGalleryField.php
```diff
@@ -46,6 +53,54 @@ class AssetGalleryField extends FormField {
...
+			throw new InvalidArgumentException('AssetGalleryField requires a DataList based on File');
+		}
+
+		if(!$dataList) {
+			$dataList = File::get();
```
tractorcow: Should the default be root files, or all files?

code/AssetGalleryField.php
```diff
@@ -74,10 +129,14 @@ public function fetch(SS_HTTPRequest $request) {
 			$this->httpError(400);
 		}
 
-		$files = File::get()->filter('ParentID', $params['id']);
+		$files = $this->getList()->filter('ParentID', $params['id']);
```
tractorcow: If we are filtering, I suggest promoting all references in this class from `SS_List` to `SS_Filterable`.

code/AssetGalleryField.php
```diff
@@ -74,10 +129,14 @@ public function fetch(SS_HTTPRequest $request) {
...
 
 		if ($files) {
 			foreach($files as $file) {
+				if(!$file->canView()) {
```
tractorcow: Question here; Can this mechanism do paging? If we are looping over all files in a folder, it WILL break on large folders. I suggest a start / limit mechanism passed in via querystring to simulate paging.

We could build it into the backend, at least, and fall back to loading full pages if omitted from the request.

code/AssetGalleryField.php
```diff
@@ -153,40 +212,35 @@ public function search(SS_HTTPRequest $request) {
 	 * @return SS_HTTPResponse
 	 */
 	public function update(SS_HTTPRequest $request) {
-		$id = $request->getVar('id');
-		$file = File::get()->filter('id', (int) $id)->first();
-
-		$code = 500;
+		parse_str($request->getBody(), $vars);
+		$id = $vars['id'];
+		$file = $this->getList()->filter('id', (int) $id)->first();
```
tractorcow: Please use correct case for `ID` field in filter.

code/AssetGalleryField.php
```diff
@@ -153,40 +212,35 @@ public function search(SS_HTTPRequest $request) {
 	 * @return SS_HTTPResponse
 	 */
 	public function update(SS_HTTPRequest $request) {
-		$id = $request->getVar('id');
-		$file = File::get()->filter('id', (int) $id)->first();
-
-		$code = 500;
+		parse_str($request->getBody(), $vars);
+		$id = $vars['id'];
```
tractorcow: Should we be doing a bit more validation on request handlers? It wouldn't hurt to have an InvalidArgumentException or two if $id isn't valid.

code/AssetGalleryField.php
```diff
@@ -153,40 +212,35 @@ public function search(SS_HTTPRequest $request) {
...
-
-		if ($file) {
-			$title = $request->getVar('title');
-			$basename = $request->getVar('basename');
-
-			if (!empty($title)) {
-				$file->Title = $title;
-			}
+		if (!$file) {
+			return (new SS_HTTPResponse(json_encode(['status' => 'error']), 500))
```
tractorcow: Shouldn't this be 404? Or is it a react thing that requires it being treated as a server error?

code/AssetGalleryField.php
```diff
@@ -203,29 +258,28 @@ public function delete(SS_HTTPRequest $request) {
 
 		if($fileIds) {
 			foreach ($fileIds as $id) {
-				if ($file = File::get()->filter("id", (int) $id)->first()) {
+				if ($file = $this->getList()->filter("id", (int) $id)->first()) {
```
tractorcow: As above, please use correct case for `ID`.

tractorcow: Also, `filter('ID', $fileIDs)` works fine and will result in much fewer SQL queries.

code/AssetGalleryField.php
```diff
@@ -203,29 +258,28 @@ public function delete(SS_HTTPRequest $request) {
...
 					array_push($files, $file);
 				}
 			}
 		}
 
-		if(count($files)) {
-			foreach ($files as $file) {
-				$file->delete();
-			}
+		if (!count($files)) {
```
tractorcow: 404 please :D

code/AssetGalleryField.php
```diff
@@ -203,29 +258,28 @@ public function delete(SS_HTTPRequest $request) {
...
+		if (!min(array_map(function($file) {return $file->canDelete();}, $files))) {
+			return (new SS_HTTPResponse(json_encode(['status' => 'error']), 401))
+				->addHeader('Content-Type', 'application/json');
+		}
 
-			$response->setBody(json_encode(array(
-				'status' => 'could not find the file',
-			)));
+		foreach($files as $file) {
+			$file->delete();
```
tractorcow: Can we do ->archive() for now? These files are versioned, so we can't leave live records lying around.

When we come to do changesets, we can revert this back to ->delete(), since then we can use the changesets feature for propegating file deletions into the live site.

code/AssetGalleryField.php
```diff
@@ -302,6 +356,10 @@ protected function getData($filters = array()) {
 			}
 
 			foreach($files as $file) {
+				if(!$file->canView()) {
```
tractorcow: filterByCallback?

tractorcow:
General feedback on the entire feature; Have you considered publishing behaviour at all? We may need to be ready to add ->doPublish(), ->doArchive(), ->doUnpublish() etc, each of which has their respective permissions canPublish(), canArchive, canUnpublish() and so on.

I'll accept that this is a consideration that could be left to a later date, however files are ALREADY versioned, so I'd expect that at a minimum this gallery doesn't leave files in an inconsistent state (e.g. unable to be removed from live site).

code/AssetGalleryField.php
tractorcow: Should this list also be requesting specifically stage records? Versioned::get_by_stage?

chillu: It's no different than the previous implementation, so I'll ignore it for this PR. Raised it with Scott, agree its a bug - most likely a regression in the initial `fetch()` call rather than PHP.

code/AssetGalleryField.php
chillu: I've actually changed them to `DataList`, since the implementation relies on `filterAny()` which isn't defined in any of these interfaces.

code/AssetGalleryField.php
chillu: Haha yes, I knew this would come up ;) Yeah existing issue, I just made it worse by also calling `canView`. Let's park this for later.

code/AssetGalleryField.php
chillu: I'll add it to the TODO list, not related to this PR

code/AssetGalleryField.php
chillu: Done

code/AssetGalleryField.php
chillu: As discussed on Slack, changed to 404 now

chillu:
I don't think file versioning has been considered when writing this form field, since it was all a bit theoretical three months ago when most of the PHP was written by Chris.

code/AssetGalleryField.php
```diff
@@ -46,6 +58,55 @@ class AssetGalleryField extends FormField {
...
+	/**
+	 * @param string $name
+	 * @param string $title
+	 * @param DataList $dataList
+	 * @throws InvalidArgumentException
+	 */
+	public function __construct($name, $title = null, DataList $dataList = null) {
+		parent::__construct($name, $title, null);
+
+		if($dataList && !is_subclass_of($dataList->dataClass(), 'File')) {
```
tractorcow: File isn't a subclass of File. Did you mean is_a?

code/AssetGalleryField.php
```diff
@@ -74,10 +135,15 @@ public function fetch(SS_HTTPRequest $request) {
 			$this->httpError(400);
 		}
 
-		$files = File::get()->filter('ParentID', $params['id']);
+		// TODO Limit results to avoid running out of memory (implement client-side pagination)
```
tractorcow: Accepted solution. :D

code/AssetGalleryField.php
tractorcow: Parking accepted.

PR #81, opened 2016-02-15
PSR-2

chillu:
Based on #80 for easier handling (merge that one first). See https://www.silverstripe.org/software/addons/supported-modules-definition/


chillu:
Rebased - @tractorcow could you merge please?

PR #82, opened 2016-02-15
Re-enabled Behat tests

chillu:
They'll be horribly broken at the moment, we'll get them fixed after @flashbackzoo's routing fixes are merged. At some point we'll also need to get the new `insert-an-image.feature` tests over from `framework`, since they would've been adapted to work with TinyMCE 4 (pull request pending on that one)


scott1702:
@chillu looks like this breaks the npm test command for some reason.

PR #83, opened 2016-02-16
Routing

chillu:
Creating a PR for easier review


chillu:
As discussed with David offline, requires fixes to popstate-based navigation of assets detail view (transition from admin/pages -> admin/assets -> admin/assets/edit and back again)

PR #87, opened 2016-03-02
Adds support for scaffolding forms based on a form field schema

flashbackzoo:
Opening this now for peer review.

Limitations:
- Only TextField and HiddenField types are supported.
- Field values on scaffolded forms are not connected to state.
- Form actions, for example submission, don't post back data.


javascript/src/components/form-builder/__tests__/form-builder-test.js
```diff
@@ -0,0 +1,27 @@
+jest.dontMock('react');
```
chillu: `__tests__` as a folder notation (with double underscores) isn't what we discussed in the architecture research doc, right? Any reason for it? It's no less publicly accessible than without the underscores :)

javascript/src/components/form-action/index.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import SilverStripeComponent from 'silverstripe-component';
+
+class FormActionComponent extends SilverStripeComponent {
+
+    render() {
+        return (
+            <button type='button' class='btn btn-primary'>{this.props.label}</button>
```
chillu: Can we make HTML attributes a prop with defaults? Start with best practices, right?

javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+        if (this.isFetching === true) {
+            return this.formSchemaPromise;
+        }
+
+        if (schema === true) {
+            headerValues += 'schema';
+        }
+
+        if (state === true) {
```
chillu: Will fail when `schema === false`. I'd suggest `[].join(',')`

javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,149 @@
...
+            switch (Component.name) {
+                case 'TextFieldComponent':
+                    return <Component
+                        key={i}
+                        extraClass={field.extraClass}
+                        label={field.leftTitle}
+                        name={field.name}
+                        value={field.value} />;
+                default:
+                    return null;
```
chillu: Heading says "If the field type doesn't exist a TextField will be used as a default.". That doesn't seem to be the case though, only the two types explicitly mentioned in `fakeInjector` will default to `TextFieldComponent`.

javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * Maps a list of schema fields to their React Component.
+     * If the field type doesn't exist a TextField will be used as a default.
+     *
+     * @param array fields
+     *
+     * @return array
+     */
+    mapFieldsToComponents(fields) {
+        return fields.map((field, i) => {
```
chillu: How would this be used by nested form field components like `TabSet`?

javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,149 @@
...
+        const formProps = {
+            actions: schema.actions,
+            attributes: schema.attributes,
+            data: schema.data,
+            fields: schema.fields,
+            mapFieldsToComponents: this.mapFieldsToComponents
+        };
+
+        return (
+            <div>
```
chillu: Why the `<div>` when we have a single node in the return value?

javascript/src/components/form/README.md
```diff
@@ -0,0 +1,37 @@
+# FormComponent
+
+The FormComponent is used to render forms in SilverStripe. The only time you should need to use `FormComponent` directly is when you're composing custom layouts. Forms can be automatically generated from a schema using the `FormBuilder` component.
+
+This component should be moved to Framework when dependency injection is implemented.
```
chillu: What's the rationale for moving the component later, rather than placing it in `framework` (and the `integrate-react` branch) in the first place? You seem to have avoided the DI issue by using fakes?

javascript/src/components/form/index.js
```diff
@@ -0,0 +1,59 @@
...
+        });
+    }
+
+    render() {
+        const attr = this.props.attributes;
+        const fields = this.props.mapFieldsToComponents(this.props.fields);
+        const actions = this.getFormActionComponents();
+
+        return (
+            <div id={attr.id} className={attr.class} encType={attr.enctype} method={attr.method} action={attr.action}>
```
chillu: Huh, no `<form>` tag?

javascript/src/components/form/README.md
```diff
@@ -0,0 +1,37 @@
...
+}
+```
+
+### data
+
+Ad hoc data passed to the front-end from the server.
+
+### fields (required)
+
+A list of field objects to display in the form. These objects should be transformed to Components using the `this.props.mapFieldsToComponents` method.
```
chillu: Good idea, composes quite nicely!

javascript/src/components/text-field/index.js
```diff
@@ -1,35 +1,53 @@
...
+
     constructor(props) {
         super(props);
 
         this.handleChange = this.handleChange.bind(this);
     }
+
     render() {
+        var inputProps = {
+            className: 'text ',
```
chillu: Can we use `[].join(' ')` here to avoid the ugly space?

javascript/src/components/text-field/index.js
```diff
@@ -1,35 +1,53 @@
...
+class TextFieldComponent extends SilverStripeComponent {
+
     constructor(props) {
         super(props);
 
         this.handleChange = this.handleChange.bind(this);
     }
+
     render() {
+        var inputProps = {
```
chillu: This should probably be a getter, similar to the `FormField` API - keep it DRY between the dozen components

javascript/src/state/schema/reducer.js
```diff
@@ -0,0 +1,33 @@
...
+
+        case ACTION_TYPES.SET_SCHEMA:
+            if (state.forms.length === 0) {
+                return deepFreeze(Object.assign({}, state, { forms: [action.payload] }));
+            }
+
+            // Replace the form which has a matching `schema.id` property.
+            return deepFreeze(Object.assign({}, state, {
+                forms: state.forms.map((form) => {
+                    if (form.schema.id === action.payload.schema.id) {
```
chillu: The `schema.id` isn't a globally unique identifier: It only stores the form name, which is the same across multiple CMS controllers (e.g. `CMSMain` and `SecurityAdmin` have different schemas for `EditForm`). It also stores state specific to a single record. At the moment, there's no single key in the schema spec which would be unique (since I've implemented `admin/assets/schema/EditForm` without an `$ID` parameter). Storing every schema and state separately for each opened record could also lead to memory issues, or even performance issues with freezing large object structures in the reducer?

javascript/src/components/form-action/index.js
flashbackzoo: Yup - I'll fix that

javascript/src/components/form-builder/__tests__/form-builder-test.js
flashbackzoo: `__tests__` is the default test directory name in Jest see https://facebook.github.io/jest/docs/api.html#config-testdirectoryname-string

I've found having tests alongside the component code makes components more portable. For example the TextFieldComponent. This will eventually move to Framework, we know exactly where all the code for that component is, it's not scattered across the codebase.

It's also nicer in the tests to be able to do `var component = require('../')` rather than `var component = require('../../javascript/src/components/component')`

javascript/src/components/form-builder/index.js
flashbackzoo: :+1:

javascript/src/components/form-builder/index.js
flashbackzoo: I think we could do something like this...

```js
...

case 'TabSetComponent':
    const children = field.children.map((child) => {
        return this.mapFieldsToComponents(child);
    });

    return <Component>{children}</Component>

...
```

javascript/src/components/form-builder/index.js
flashbackzoo: Yeah I'll update that comment

javascript/src/components/form-builder/index.js
flashbackzoo: Yep - It doesn't need to be there. I'll remove it

javascript/src/components/form/index.js
flashbackzoo: Ahh I forgot to change this back. Nesting `<form>` elements throws an error. So when I was testing this in the AssetAdmin section it broke because the scaffolded form was inside a server rendered form.

javascript/src/components/form/README.md
flashbackzoo: It seemed like there would be less throwaway code doing it like this. If we move it to `framework` now we'd have to expose `FormComponent` as a browserify external so we could `import` it in AssetAdmin. If we were doing that we'd want to move `TextFieldComponent` as well to be consistent (and maybe `BulkActionComponent` and `FileComponent`). Then we'd remove that export code from `framework` and import those components differently in AssetAdmin when we implement DI. Seemed like lots of mucking around for little benefit when we can do it once properly in a few weeks.

javascript/src/components/text-field/index.js
flashbackzoo: Yeah that makes sense

javascript/src/components/text-field/index.js
flashbackzoo: I'm on the fence as to what is less ugly there lol but yeah using `join` I suppose is better.

javascript/src/components/form-builder/__tests__/form-builder-test.js
chillu: I agree that tests should live alongside their components - this is just about `__test__` vs. `test`. It's inconsistent with our PHP tests (which live in a `tests/` folder), and not what we've written down in "folder layout example" on https://docs.google.com/document/d/1MXDhNPz60Lt5X4VlyB6hFUJGl3H6vIPxPAqV2lmdMnI/edit. 
Can you configure jest to use `test` as a folder name, or does that cause more boilerplate on every test definition?

javascript/src/state/schema/reducer.js
flashbackzoo: Hmmm yeah I think we'll need a unique identifier.

I would expect the `forms` key to be reset when the route changes. So you would only have forms for the current view in memory at one time. The resetting isn't implemented but that's how I imagined it would work. Originally I just had a `form` key stored a single schema but there could be views which have multiple forms.

javascript/src/components/form-builder/__tests__/form-builder-test.js
flashbackzoo: That's should be possible. I'll see if it works.

javascript/src/components/form-builder/index.js
chillu: Discussed this offline with David, he's going to remove the `switch` statement in favour of each form field component receiving the same data. The field component can then decide on how to handle the schema at hand, for example by adding children in its own `render()` method.

javascript/src/components/form/README.md
chillu: Fair enough

flashbackzoo:
@chillu see https://github.com/open-sausages/silverstripe-asset-admin/commit/f95598b11f557444dc2a7ba183733cd2d0faf32c for updates based on your feedback

javascript/src/components/form-builder/index.js
```diff
@@ -0,0 +1,229 @@
...
+                fields: field.children,
+                data: field.data,
+                extraClass: field.extraClass,
+                name: field.name
+            };
+
+            // Structural fields (like TabSets) are not posted back to the server and don't receive some props.
+            if (field.type !== 'Structural') {
+                props.description = field.description;
+                props.id = field.id;
```
chillu: `CompositeField` is a structural field, and can have a `$legend` set - which I guess should go in `props.description`?

I think `props.id` is also valid for each form field - it's a unique HTML identifier. For example `Tab` uses it to make tabs uniquely addressable.

Also, we missed `tabindex` in the spec... @tractorcow was that on purpose?

tests/react/component/gallery-component-test.js
```diff
@@ -1,92 +0,0 @@
-// jest.dontMock('../../../public/src/component/gallery-component.js');
```
chillu: Why did you remove this test file? I was under the impression that we just need to fix in order to re-enable them?

chillu:
I've started writing some more tests for this, it's a bit light on it. Copying my WIP here. Doesn't block the card, but we're creating more technical debt here by skipping this step. On the specific tests, it highlights that the form field mapping isn't mockable, unless we make the targeted injector system part of the test scaffold. For now, it's fine to rely on `fakeInjector` to test field mapping.

```diff
diff --git a/javascript/src/components/form-builder/tests/form-builder-test.js b/javascript/src/components/form-builder/tests/form-builder-test.js
index 99e7807..a10d70d 100644
--- a/javascript/src/components/form-builder/tests/form-builder-test.js
+++ b/javascript/src/components/form-builder/tests/form-builder-test.js
@@ -1,27 +1,52 @@
 jest.dontMock('react');
 jest.dontMock('../index.js');
+jest.dontMock('../../text-field/index.js');
 
 var FormBuilderComponent = require('../index.js');
+var TextFieldComponent = require('../../text-field/index.js');
 
 describe('FormBuilderComponent', () => {
 
-    describe('getFormSchema()', () => {
-        var formBuilder;
-
-        beforeEach(() => {
-            const props = {
-                actions: {},
-                formSchemaUrl: 'admin/assets/schema/1',
-                schema: { forms: [{ schema: { id: '1', schema_url: 'admin/assets/schema/1' } }] }
-            };
-
-            formBuilder = new FormBuilderComponent({...props});
-        });
+    var formBuilder;
+    beforeEach(() => {
+        const props = {
+            actions: {},
+            formSchemaUrl: 'admin/assets/schema/1',
+            schema: { forms: [{ schema: { id: '1', schema_url: 'admin/assets/schema/1' } }] }
+        };
+        formBuilder = new FormBuilderComponent({...props});
+    });
 
+    describe('getFormSchema()', () => {
         it('should return the form schema for the FormBuilder', () => {
             const form = formBuilder.getFormSchema();
             expect(form.schema.id).toBe('1');
         });
     });
 
+    describe('mapFieldsToComponents()', () => {
+        it('should return null when a field component is not found', () => {
+            const fields = formBuilder.mapFieldsToComponents([
+                {component: 'Unknown'}
+            ]);
+            expect(fields.length).toBe(1);
+            expect(fields[0]).toBeNull();
+        });
+
+        it('should return component by data type', () => {
+            const fields = formBuilder.mapFieldsToComponents([
+                {component: 'String'}
+            ]);
+            console.log(fields);
+            expect(fields[0]).toEqual(jasmine.any(TextFieldComponent));
+        });
+
+        it('should return component by name', () => {
+            const fields = formBuilder.mapFieldsToComponents([
+                {component: 'TextField'}
+            ]);
+            expect(fields[0]).toEqual(jasmine.any(TextFieldComponent));
+        })
+    });
+
 });
```

flashbackzoo:
@ingo I've updated the tests, made the changes to the props, and squashed.

chillu:
Thanks for your patience David :) Merging now, acknowledging that builds aren't passing which should be our next priority to fix. 

I've pinged @tractorcow again on the `tabindex` issue. The non-unique identifier issue will come up again automatically as we implement forms, so happy to defer this as a bit of technical debt.

PR #90, opened 2016-03-06
Fix being able to edit files

scott1702:
Previously the editing of files didn't work, this fixes it. Relies on https://github.com/open-sausages/silverstripe-asset-admin/pull/86 being merged first


scott1702:
Will be fixed with implementation of form field schema instead.

PR #95, opened 2016-03-30
Remove outdated requirement: bundle-react

scott1702:
React is now included in bundle-lib, so shouldn't be required here


scott1702:
Oh—@chillu you beat me to it :grinning:

PR #100, opened 2016-04-01
Changed file section checkbox from <button> to <input type="checkbox">

steve-silverstripe:
BEFORE:
image

AFTER:
image


tractorcow:
@flashbackzoo @chillu please approve this. :D

javascript/src/components/file/index.js
```diff
@@ -175,29 +175,27 @@ class FileComponent extends SilverStripeComponent {
...
     } else {
-      actionButton = (<button
+      actionInputCheckbox = (<input
         className="item__actions__action item__actions__action--select [ font-icon-tick ]"
-        type="button"
+        type="checkbox"
         title={i18n._t('AssetGalleryField.SELECT')}
         tabIndex="-1"
         onMouseDown={this.preventFocus}
         onClick={this.handleToggleSelect}
```
flashbackzoo: This should be an `onChange` handler now it's a checkbox. You should also be able to remove the `onMouseDown` handler which looks like it was the as some sort of hack to get around focusing on the button? @scott1702

scott1702: It was for exactly that @flashbackzoo :grinning:

scott1702:
@steve-silverstripe could you please rebase?

clarkepaul:
I think this might have broken the checkboxes? I think the before and after shots might have been done without a build to the css as when I checked out the latest the markup was still "button" and after a build it was input without the styles.

clarkepaul:
I think its that a input/checkbox cant have padding applied to it so it looses the spacing. I think I'll just swap the padding for margin, should be a quick change. I hope.

clarkepaul:
Input doesn't support ::before in some browsers so they are not visible.

PR #103, opened 2016-04-03
Fix IE babel support

chillu:
Also added `deep-freeze` and sourcemaps


flashbackzoo:
Do we need to update the shrinkwrap file?

chillu:
@flashbackzoo The shrinkwrap file was already updated - I've just done it in one roll-up commit rather than separately for the `deep-freeze` commit.

tests/php/AssetAdminTest.php
```diff
@@ -17,14 +17,14 @@ public function setUp()
...
-            if(!file_exists(BASE_PATH."/$folder->Filename")) mkdir(BASE_PATH."/$folder->Filename");
+            if(!file_exists(BASE_PATH."/$folder->Filename")) mkdir(BASE_PATH."/".$folder->generateFilename());
         }
 
         // Create a test files for each of the fixture references
         $fileIDs = $this->allFixtureIDs('File');
         foreach ($fileIDs as $fileID) {
             $file = DataObject::get_by_id('File', $fileID);
-            $fh = fopen(BASE_PATH."/$file->Filename", "w");
+            $fh = fopen(BASE_PATH."/".$file->generateFilename(), "w");
```
tractorcow: That's now how you get the local file path. :D

tractorcow: See https://github.com/silverstripe/silverstripe-asset-admin/pull/108

chillu:
Tests will be fixed by this PR: https://github.com/silverstripe/silverstripe-asset-admin/pull/109, specifically https://github.com/silverstripe/silverstripe-asset-admin/pull/109/commits/82408686b33412b26195500cff70625a9ab1109e.

package.json
```diff
@@ -60,6 +67,7 @@
     "watchify": "^3.7.0"
   },
   "dependencies": {
+    "deep-freeze": "https://registry.npmjs.org/deep-freeze/-/deep-freeze-0.0.1.tgz",
```
chillu: Note that this is using the framework dependency at build time (via `external('deep-freeze')` in gulpfile.js), but it's technically still a dependency for the project itself, and `jest` testing

PR #108, opened 2016-04-05
BUG Use asset abstraction API for mocking test fixture files

tractorcow:


tractorcow:
These tests haven't been updated since we added the new API.

tests/php/AssetAdminTest.php
```diff
@@ -5,64 +5,44 @@
...
 {
+
     protected static $fixture_file = 'AssetAdminTest.yml';
 
     public function setUp()
     {
         parent::setUp();
 
-        if(!file_exists(ASSETS_PATH)) mkdir(ASSETS_PATH);
+        AssetStoreTest_SpyStore::activate('AssetAdminTest');
```
tractorcow: I regret calling this class this name... now it's everywhere haha.

PR #109, opened 2016-04-06
Add-folder feature & refactoring

sminnee:
This refactoring incorporates a number of changes

 - FileBackend is no longer used. Connection to the backend is via a
   number of API callback props, each of which expects a function that
   returns a promise.
 - AssetAdmin has a clearer set of responsibilities: tie together
   Editor & Gallery, and manage interaction with URLs.
 - The Gallery & Editor controls are more isolated from the each other
   and the running context. A visible prop on each one controls
   whether or not it is shown, coupled with show() and hide() actions
 - As part of this, Editor now has its own actions & reducers.
 - Editor and Gallery are regular React.Components
 - Editor and Gallery are connected to each other by onEvent()
   methods, that are set by AssetAdmin.

This is still a work in progress and I know of the following 
bugs:

 - Direct loading a URL of admin/assets/show/0/edit/73 doesn’t work.
   It needs to pre-fetch the folder contents before loading the
   editor, which is probably a job for AssetAdmin to coordinate.
 - Search seems to be broken; I don’t know if it previously worked.
 - Staggered loading is broken; I don’t believe it previously worked.

There are also aspects of the refactoring that I’m not entirely happy
with:

 - The triggering of the ajax-reload of a folder is a bit messy, with
   folderID and loadedFolderID.
 - The components still pull from global state in mapStateToProps().
   It’s not clear to me if this code should be part of the component
   or part of the wiring-together that AssetAdmin does.


sminnee:
Ingo has raised the following issues that I'm addressing at the moment:

- [ ] Remove commented out code in `GALLERY.ADD_FOLDER_SUCCESS`
- [ ] Remove a `console.log()` the editor action "show"
- [ ] Lint the PHP code against PSR-2
- [ ] Add PHP redirection for the SS3-style edit-file-details URL'
- [ ] Fix the string-replacement URL builder to catch the `?`
- [ ] Avoid a 2nd PHP load when shifting URL from `admin/assets` to `admin/assets/show/0`
- [ ] Add prefix-text to all action-type constants

Source:
 - https://github.com/sminnee/silverstripe-asset-admin/commit/205f40acb5c78a0894a49e11fac04ce5d9f590b9
 - https://github.com/sminnee/silverstripe-asset-admin/commit/7adaa6b9fff63fcd183914282fd72d56e5594a77

sminnee:
OK I've addressed the points above, and `npm test` works for me now too.

sminnee:
> The requested package silverstripe/cms dev-integrate-react as 4.0.x-dev could not be found.

Does anyone know what this error relates to?

tractorcow:
`- DB=MYSQL CORE_RELEASE=integrate-react CORE_INSTALLER_RELEASE=master CORE_ALIAS=4.0.x-dev`

in .travis.yml

sminnee:
OK, AssetAdmin is now namespaced. This means that it can be included into the current cms master (although you will have 2 asset admins available, the ss3 one is at admin/old-assets)

This means that we can test against CMS master instead of a sausage branch.

gulpfile.js
```diff
@@ -1,84 +1,85 @@
...
-		.pipe(gulp.dest(PATHS.JAVASCRIPT_DIST));
+  return browserify(browserifyOptions)
+    .transform(babelify)
+    .external('components/text-field/index')
+    .external('deep-freeze')
+    .external('react')
+    .external('jQuery')
+    .external('i18n')
+    .external('silverstripe-component')
+    .external('silverstripe-backend')
```
chillu: `silverstripe-backend` isn't exposed in framework `master`, so this currently relies on the `pulls/4.0/campaign-admin` branch: https://github.com/silverstripe/silverstripe-framework/pull/5278. This pull request is still under discussion, might be worth isolating that one change into separate pull request if its the only dependency?

javascript/src/sections/asset-admin/controller.js
```diff
@@ -1,159 +1,140 @@
 import $ from 'jQuery';
 import React from 'react';
 import SilverStripeComponent from 'silverstripe-component';
+import { default as GalleryContainer } from 'sections/gallery/controller';
+import EditorContainer from 'sections/editor/controller';
 import * as galleryActions from 'state/gallery/actions';
+import * as editorActions from 'state/editor/actions';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
-import FileBackend from 'backend/file-backend';
```
chillu: You should also remove the `file-backend.js` file, its no longer needed right?

javascript/src/boot/index.js
```diff
@@ -4,38 +4,101 @@ import ReactDOM from 'react-dom';
...
   $('.asset-gallery-component-wrapper').entwine({
     onadd() {
       const store = configureStore();
-      const galleryProps = getGalleryProps();
+
+      // Build API callers from the URLs provided to us in the div
+      // In time, something like a GraphQL endpoint might be a better way to run
+
+      const filesByParentApi = apiCallerFromEndpoint({
+        method: 'post',
```
chillu: This should be `get`

chillu:
There's a bunch of linting errors. You can run `eslint javascript/src` for now, but that requires some NPM global installs. I've added `npm run lint` via 347abb1 to `master` now.

image

gulpfile.js
sminnee: Agreed, will do. It also requires that post vars have a content-type header set, but it's a small change that can be a separate PR.

javascript/src/sections/asset-admin/controller.js
sminnee: Agreed, was being conservative in my change but leaving it is just clutter.

javascript/src/boot/index.js
chillu: Actually, it's a bit all over the show (not from your PR). Can you switch that to use GET consistently?

```
class AssetGalleryField extends FormField
{

    private static $url_handlers = array(
        'GET getFilesByParentID' => 'getFilesByParentID',
        'GET getFilesBySiblingID' => 'getFilesBySiblingID',
    );

    public function getFilesByParentID(SS_HTTPRequest $request)
    {
        $params = $request->postVars();
        // ...
    }

    public function getFilesBySiblingID(SS_HTTPRequest $request)
    {
        $params = $request->postVars();
        // ...
    }
```

gulpfile.js
sminnee: Hrm, actually, I'm not sure whether or not it's the only dependency — I've never tried using this with master as the whole dev cycle was done with campaign-admin loaded in.

I'll have a go at cherry-picking relevant commits and doing some end-user testing; I'm not sure that the JS test suite is sufficient for "passing tests" to mean "works".

sminnee:
To do:
 - [ ] Run `eslint javascript/src` on the whole codebase, fix all errors
 - [ ] Try and carve out a separate framework PR to support this
 - [ ] Delete file-backend.js
 - [ ] Change `filesByBackendApi` to `get`.

javascript/src/boot/index.js
```diff
@@ -4,38 +4,101 @@ import ReactDOM from 'react-dom';
...
+      }
+    }
+    return str.join('&');
+  }
+
+  return (data) => {
+    let url = endpoint.url;
+    let backendData = data;
+
+    if (endpoint.method === 'get' && endpoint.payloadFormat === 'urlencoded') {
```
chillu: Shouldn't data always be added as HTTP query params regardless of the `payloadFormat`? Can `GET` even have a HTTP message body? I can't see it explicitly excluded in https://en.wikipedia.org/wiki/HTTP_message_body, but it's pretty unusual right?

_config/config.yml
```diff
@@ -0,0 +1,6 @@
+---
+Name: assetadmin
+---
+AssetAdmin:
+  menu_title: 'Files2'
+  url_segment: 'old-assets'
```
chillu: Should we also rename the menu entry to "old files"? Looks a bit weird at the moment. Also, it's using the wrong icon?

image

sminnee:
FYI @chillu filesByBackendApi is correctly using POST, i.e., that is what the PHP API current requires. Do you want me to update both the backend-API and the javascript-call to use GET? I agree it would be more restful.

chillu:
> FYI @chillu filesByBackendApi is correctly using POST, i.e., that is what the PHP API current requires.

Yeah, my point is that the PHP API seems wrong - its a stateless HTTP call, right? I just failed to notice it until now :) Also, for some reason the UI didn't work for me on the vanilla pull request checkout, and started working after I changed the GET/POST behaviour. No point tracking down why now. So yes, please update it to GET

sminnee:
@chillu ironically the change I'm struggling with the most is fixing the icon. I need to work out how the asset-admin module can add global CSS, and I'm not sure of the best pattern to do this. Is it a blocker for merge?

chillu:
Nah, not a blocker

sminnee:
OK pushed everything except for the menu icon/label changes.

chillu:
Sweet, merged! BTW, some places in the old `admin/assets` seem to be hardcoded to that base path - for example, navigating up to the parent folder switches from `admin/assets-old` to `admin/assets`. Regression tested other features like upload, create folder, file editing and deletion in both the old and new UIs, and in general it works.

PR #112, opened 2016-04-12
REFACTOR: Update to use createEndpointFetcher

sminnee:
asset-admin previously had its own implementation of
createEndpointFetcher, apiCallerFromEndpoint. Now that
createEndpointFetcher is part of framework master, we can use that.


chillu:
Currently blocked by https://github.com/silverstripe/silverstripe-framework/pull/5316, since using the framework functionality will introduce a regression until that's merged

chillu:
@sminnee I've removed the `payloadFormat: 'querystring'` special case in https://github.com/silverstripe/silverstripe-framework/pull/5307/files#diff-8c1c640b30d75afcd1d440292986d7e9L133, which now breaks this PR. 

Can you please remove this reference from your PR here? No replacement needed. The silverstripe-backend API will handle this automatically now: If its a `get` request, any `data` will be added as the query string. If its not a `get` request, you have to use the new `payloadSchema` setting with per-key `querystring: true` settings.

chillu:
Rebased on master, merged as 3aa16d9b37252bdee9377480c7677ed7d31a998b

PR #113, opened 2016-04-18
Editorconfig for various file formats

chillu:


tractorcow:
I guess since we agreed to follow airbnb, and that dictates 2 spaces indentation, fine with me.

PR #118, opened 2016-04-20
BUG Fix asset admin in subfolders

tractorcow:
window.ss.router.resolveURLToBase() is ment to be used for creating new routes, not for routing actual requests, so removing these fixes routing in subfolders. :)


sminnee:
Seems fine to me. The lack of tests around this stuff is a bit irksome - how hard would it be to start fleshing out the JS unit tests around this?

Not necessarily a merge blocker, since this code is no worse than what's in master, but good to think about for a 2nd PR

tractorcow:
Yeah I need to figure out how to write more than trivial js tests.

PR #119, opened 2016-04-21
Rename original AssetAdmin to "Files (old)"

chillu:
Can't use menu_title via YAML config if a _t() key is defined.
Using the fully qualified class name here, the YAML parser
doesn't seem to mind the backslashes.

Will only apply in English for now, since other languages have their own translation keys. Since this is a temporary situation, there's not much point going through the hassle to change all of these IMHO

Merge with https://github.com/silverstripe/silverstripe-framework/pull/5366

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -19,7 +19,7 @@ en:
     FROMYOURCOMPUTER: 'From your computer'
     Filetype: 'File type'
     ListView: 'List View'
-    MENUTITLE: Files
+    MENUTITLE: Files (old)
     NEWFOLDER: NewFolder
     SIZE: Size
     THUMBSDELETED: '{count} unused thumbnails have been deleted'
@@ -50,3 +50,5 @@ en:
     PreviewButton: Preview
   Permission:
     CMS_ACCESS_CATEGORY: 'CMS Access'
+  SilverStripe\AssetAdmin\Controller\AssetAdmin:
+    MENUTITLE: Files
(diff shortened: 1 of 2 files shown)


sminnee:
Seems good to me.

tractorcow:
Don't forget to update `private static $menu_title = 'Files';`

lang/en.yml
```diff
@@ -19,7 +19,7 @@ en:
     FROMYOURCOMPUTER: 'From your computer'
     Filetype: 'File type'
     ListView: 'List View'
-    MENUTITLE: Files
+    MENUTITLE: Files (old)
```
tractorcow: Hm, we share transifex strings for both 3.x and 4.x. I need to figure a way of making sure this doesn't somehow modify 3.x strings.

tractorcow:
The next time we update transifex strings it's going to backport "Files (old)" to 3.x.

lang/en.yml
tractorcow: What will happen is, either:
- i18nTextCollector will revert this back to `File', or
- The next push of new masters to transifex will reset the translation for the default locale in all versions.

tractorcow:
I think the solution for this needs to be a bit of hacked in PHP logic. =/ Temporary of course before we remove the `old` suffix.

tractorcow:
Possible fix at https://github.com/silverstripe/silverstripe-cms/pull/1467: Needs review

PR #120, opened 2016-04-21
Rename JS files to comply with AirBnB conventions

chillu:
Dependant on https://github.com/silverstripe/silverstripe-framework/pull/5372


chillu:
This will also fix an issue where IE10 doesn't load `asset-admin/javascript/AssetAdmin.js` because its been converted to ES6 and "lost" its compilation step at some point. It's legacy entwine JS that no longer applies to any section in the new asset-admin, so can be removed.

chillu:
Looked this over with @tractorcow and @flashbackzoo - they've both done a partial peer review. Merging now, since it's starting to inhibit our workflow around other pull requests.

PR #126, opened 2016-04-29
Add code climate

chillu:


chillu:
Looks like Code Climate only runs the first build on the master branch, so merging this to fix the current eslint error (https://codeclimate.com/github/silverstripe/silverstripe-asset-admin/builds/1)

PR #127, opened 2016-05-02
Remove non-asset-admin strings

tractorcow:
These belong to other modules and shouldn't be duplicate translated on asset-admin

See https://github.com/silverstripe/silverstripe-cms/pull/1467


chillu:
We'll probably have to port over some of them (e.g. "From" and "To" filter labels might still be required), but that can be done on a case-by-case basis. We just have to remember it when we implement UI features (go check if strings with existing translations already exist in the `cms` module for the old `AssetAdmin`)

PR #151, opened 2016-05-10
Transifex support

chillu:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/128


tractorcow:
Several fixes to codeclimate issues at https://github.com/silverstripe/cow/pull/10

It won't reduce issues to 0, but it will fix many.

PR #154, opened 2016-05-11
Fix file editing and saving

chillu:
Merge together with https://github.com/silverstripe/silverstripe-framework/pull/5511


tractorcow:
I've added a bugfix to the `npm` test failure and merged this separately. https://github.com/silverstripe/silverstripe-asset-admin/commit/fb3df1531f04016b33aaba87133ed08ca79b46e4

PR #164, opened 2016-05-19
Checkbox styling regression

clarkepaul:
Fixes styling regression issues in #100
pasted_image_20_05_16__11_16_am

After:
pasted_image_20_05_16__11_26_am


clarkepaul:
There is a bug in IE10 which I'm looking into

clarkepaul:
Looks like you can't have ::before on an input in IE10 so the checkbox is missing.

clarkepaul:
So I've done the styles for the checkbox but it will require a dev to hook it up again with React as the markup has changed. Items should get the class .item--selected when the checkbox is triggered.

clarkepaul:
Partial fix for https://github.com/silverstripe/silverstripe-asset-admin/issues/165
Continued from changes made here https://github.com/silverstripe/silverstripe-asset-admin/pull/100

clarkepaul:
I think everything has been resolved for this to go back into master, button checkboxes have been converted into input type=checkbox and markup has changed so there is a label wrapping the checkbox for styling the fake checkbox and cancel button. Tested in IE10

PR #178, opened 2016-05-31
File editor sidebar (fixes #167)

### LINKED ISSUE #167: File details moved to side-bar
As a user I would like to see and edit the details of a file in a sidebar.

## Acceptance Criteria

 * File details shown in a side-bar, rather than a separate view.

## Tasks

 * [x] Adapt sidebar styling from existing pull request (https://github.com/silverstripe/silverstripe-asset-admin/pull/107)
 * [x] Sidebar layout and responsive view
 * [x] Use Bootstrap styled text fields (incl. compressed label style)
 * [x] Convert classes from pull request to BEM, add into existing component structure
 * [x] Style close icon
 * [x] Restyle "properties" section (moved to hidden tab)
 * [x] Batch actions functional on smaller screens
 * [x] Dropzone component class naming (convert to BEM)
 * [x] Browser testing

## Excludes

 * Magnification
 * Sticky footer
 * Additional actions (dot-dot-dot)
 * "Usage" tab

## Resources

 * [Designs](https://projects.invisionapp.com/boards/9GX68PTXUQFH#/1547737)

## Related

 * [Edit/rename folders](https://github.com/silverstripe/silverstripe-asset-admin/issues/156)

chillu:
Required a fairly major refactor of the state management in this area:
The original "show/hide" state of editor and gallery no longer applies,
and selection/highlight state needs to be synced between the two components.

Still not 100% happy with the `loadedfolderId` state,
this should be replaced with a `folderId` and `folder`
combination which keeps in sync via the reducer.
The components can then decide based on the presence/absence
of `folder` data if the folder needs to be loaded.

I've had a separate `file` state in `EditorReducer` at first,
but that was hard to keep synced with the one in `GalleryReducer`.
It's easier to make the `Editor` component dependant on `Gallery` state
via `mapStateToProps()` (but not vice versa) - so there's only one source
of truth about the currently active file.

Paul's done some good work on getting the components more aligned with BEM.

There's one outstanding bug around IE10-13 not changing the URL,
posting this PR so we can get some peer review.

See https://github.com/silverstripe/silverstripe-asset-admin/issues/167


chillu:
@unclecheese @assertchris This is probably the most involved React state management I've touched so far, would be good to get your feedback on some of the approaches taken :) 

Can you check this abbreviated piece of code below and see if you can think of a better way to achieve this? As a guiding principle, I want to keep `<Gallery>` decoupled from `<Editor>`, with `<AssetAdmin>` tying them together. Yes, there's more work required to extract presentational components :)

Issue #1: `<AssetAdmin>` sets a `folderId` and `fileId` state based on URL paths, which triggers an async folder load. On success, the `file` state is set (if required) through the components `mapStateToProps()`, based on filtering all loaded `files` for the `fileId`. A similar path is taken when a user selects a file (rather than through the URL). So we don't have `files` objects available in every context, meaning we need to track `fileId` "lazily" in both `<Gallery>` and `<Editor>` components, with the `file` object being available at a later point. To avoid state sync issues, I'm computing `file` in `mapStateToProps()`, rather than setting it through a reducer. Is that a good approach?

Issue #2: I want to set the initial form state in `<Editor>` from a `file` object, but decouple it afterwards for editing. Form changes only populate back to the `file` object on save. The same mounted `<Editor>` component can change its `file` props context without calling the constructor or `componentDidMount` again (based on URL changes or user clicks in `<Gallery>`). To ensure there's only one version of the truth for a file (`files[fileId]` in `gallery` state tree), I'm using `mapStateToProps()` to get it into `<Editor>` as a computed property. At this point I can't easily tell if the `file` has changed, or the `formState` has deviated from the `file` defaults - so I can't set `formState` without potentially overwriting existing edits. I've chosen `componentWillReceiveProps()` since I can compare the new props but existing ones (and detect `file` changes), but that means I'm setting *new* props which the component ingests, triggering another `componentWillReceiveProps()` - hardly ideal. `setFile()` in `GalleryActions` could emit editor actions as well, but that's too much coupling. Can't put the logic in the `GalleryReducer` where the `fileId` is set, since it doesn't have access to the `EditorReducer` state tree "namespace". One idea Chris had is to combine the two reducers and remove the "namespace" barrier. This might be a non-issue once we use `redux-form` with [initialState](http://redux-form.com/5.2.5/#/examples/initializing-from-state). Any ideas?

```
class AssetAdmin extends SilverStripeComponent {
  
  // called from routing (e.g. admin/assets/show/9/edit/10 for folderId=9 and fileId=10)
  handleURL(router, context) {
    // ...

    const folderId = parseInt(context.params.folderId, 10);
    const fileId = parseInt(context.params.fileId, 10);

    // These actions will will trigger refreshUrl()
    // since they case changes to the 'folderId' and 'fileId' props on this component
    // Will also trigger an AJAX load for the folder contents if not available
    this.props.actions.gallery.setFolder(folderId);
    this.props.actions.gallery.setFile(fileId);
  }

  // called when an item is clicked
  handleOpenFile(fileId) {
    this.props.actions.gallery.setFile(fileId);
  }

function mapStateToProps(state) {
  const { files, fileId, folderId } = state.assetAdmin.gallery;
  let file = null;
  if (fileId) {
    // Calculated on the fly to avoid getting out of sync with lazy loaded fileId
    // TODO Use ID-indexed maps for performance and ease of use
    file = files.find((next) => next.id === parseInt(fileId, 10));
  }
  return {
    folderId,
    files,
    fileId,
    file,
  };
}

class Editor extends Component {
  constructor(props) {
    super(props);

    // ...

    // Set initial form state
    this.props.actions.updateFormState(Object.assign({}, props.file));
  }

  componentWillReceiveProps(nextProps) {
    // If the file has changed, set the new form state accordingly.
    // Causes another prop update on this component which isn't ideal.
    if (nextProps.file !== this.props.file) {
      this.props.actions.updateFormState(Object.assign({}, nextProps.file));
    }
  }

  // ...
}

function mapStateToProps(state) {
  const { files, fileId } = state.assetAdmin.gallery;
  let file = null;
  if (fileId) {
    // Calculated on the fly to avoid getting out of sync with lazy loaded fileId
    file = files.find((next) => next.id === parseInt(fileId, 10));
  }
  return {
    file,
    formState: state.assetAdmin.editor.formState,
  };
}
```

chillu:
@hafriedlander I've squashed a bugfix into the "Fixed editor/gallery state management for sidebar display" commit with https://github.com/silverstripe/silverstripe-asset-admin/pull/178/commits/3ea49a378687a03e64906d9c112f544d915623fa. In hindsight, probably not the best idea since you're actively reviewing this. Anyway, this unsets the `fileId` state when a folder is changed to avoid it staying highlighted when you navigate back to the original folder. I've also simplified the highlighting by removing `unhighlightFiles()`, and allowing `highlightFiles()` with a `null` argument instead (used the same approach with the `setFile()` and `setFolder()` actions, rather than separating out `unsetFile()` etc).

client/src/styles/bundle.scss
```diff
@@ -1,15 +1,17 @@
 // Variables
-@import "../../../../framework/admin/client/src/styles/variables";
+// @import "../../../../framework/node_modules/bootstrap/scss/bootstrap";
```
hafriedlander: In general it's better to remove code than comment out.

unclecheese:
Hi, Ingo -- these are great (and quite common) questions in my experience. There's a lot of unsettled ideology on how to solve these problems in a React/Redux set up.

For the first issue, I've done it both ways, but my preference is to move the computation to the reducer. In fact, there's little harm in creating a `selectedFile` branch of your store. The true concern of your `<Editor/>` component is really `selectedFile` -- not so much the `fileID` or `files`. There's no reason why your editor needs to be aware of all files, unless you have some next/prev navigation embedded in it. That should be concern of your reducer. 

Sure, a `selectedFile` tree is duplicated data, and that feels a bit unclean, but because the data flow is so tightly controlled in redux, I find that there's little opportunity for things to get out of sync. It's particularly useful if your detail view contains data that isn't in the list view. You can lazily load those extra fields when the detail view is requested, and backfill them to the list for subsequent viewings.  All that said, I don't see anything wrong with your approach, and it's not the first time I've seen it done that way.

Second issue: I always find it's best to put my [router state in my store](https://github.com/reactjs/react-router-redux) to maintain a clear data flow. After all, you want the router state to be your single source of truth. My general reaction, though, is that you're a bit over-concerned about coupling. Once you break these things out into presentational components and containers, there's nothing wrong with creating a container that is coupled. I always think of my containers as having a carte blanche to be very specific to my application. The reason is that, because all they do is render presentational components, it's trivial to spin up a new one with different context and/or state awareness. So if your concern is about how you will reuse the editor in another context, I would give yourself permission to have an `AssetAdminEditor` container and a `OtherAdminEditor` container. Reusability is for the presentational components, but containers can be one-offs.

chillu:
Thanks so much @unclecheese! A lot of lightbulbs turning on in my head right now :)

PR #186, opened 2016-06-03
Fixed "gulp css" watch (renamed from "gulp sass")

chillu:
gulp.watch() was using the wrong folder (too deep) which meant it didn't watch the right files.
It also was inconsistently named between framework (gulp css).

This still has duplication between gulp.watch() and watchify, which is necessary
because watchify doesn't work at the moment.


chillu:
@unclecheese Looks good to you? Wanna merge?

PR #196, opened 2016-06-15
API update deep-freeze to deep-freeze-strict

tractorcow:
Fixes regressions in https://github.com/silverstripe/silverstripe-framework/pull/5436


tractorcow:
Thank you @hafriedlander

Issue #171, opened 2016-05-24
Use FormBuilder in file edit form


chillu:
`Editor.js` currently instanciates `<TextFieldComponent>` directly (through `props.editorFields`), which is a relic of an earlier implementation when we didn't have the `FormBuilder` component.

- [x] Create a form in the `AssetAdmin` PHP class, with "Name" and "Title" fields
- [x] Reference the form schema location in AssetAdmin->getClientConfig(), allow passing through to the `<Editor>` component
- [x] Use `FormBuilder` in `<Editor>` component to create form
- [x] Save file through standard PHP form API (so we can trigger validation errors, tracked [separately](https://github.com/silverstripe/silverstripe-framework/issues/5523)). Form submission could be hidden in current `AssetGalleryField->update()` API endpoint

tractorcow:
Question about this; Do you see us doing this correctly in CampaignAdmin? If that's the case, I can it as a base when porting across to asset-admin.

chillu:
Not sure if I understand you correctly - we're using `FormBuilder` in CampaignAdmin already. It's not 100% since we should use a form library for dealing with form state (rather than the `onChange` handlers writing into our own state tree at the moment). But CampaignAdmin has the currently correct way of creating a form through `FormBuilder` :)

flamerohr:
Framework changes:
https://github.com/open-sausages/silverstripe-framework/tree/pulls/4.0/formbuilder-file-edit
Assets-admin changes:
https://github.com/open-sausages/silverstripe-asset-admin/tree/pulls/4.0/formbuilder-file-edit

tractorcow:
PRs at:
- https://github.com/silverstripe/silverstripe-framework/pull/5880
- https://github.com/silverstripe/silverstripe-asset-admin/pull/215

PR #202, opened 2016-06-29
WIP AssetGalleryField -> AssetAdmin API refactor

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/5725 for context. Still requires a few more tests before it's ready to merge.


client/src/boot/index.js
```diff
@@ -1,66 +1,50 @@
...
+
     onadd() {
-      const store = configureStore();
+      const sectionConfig = Config.getSection('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin');
+      const store = configureStore()
+
+      const createEndpoint = function(endpointConfig) {
+        return backend.createEndpointFetcher(Object.assign(
+            endpointConfig,
+            { defaultData: { SecurityID: Config.get('SecurityID') } }
```
tractorcow: image

tractorcow: I think a {} at the start should fix that. :D

client/src/boot/index.js
```diff
@@ -1,66 +1,50 @@
...
 import backend from 'lib/Backend';
 
 $.entwine('ss', () => {
-  $('.asset-gallery-component-wrapper').entwine({
+  $('#assetadmin-cms-content .cms-content-view').entwine({
+
     onadd() {
-      const store = configureStore();
+      const sectionConfig = Config.getSection('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin');
+      const store = configureStore()
```
tractorcow: Semicolon

client/src/boot/index.js
```diff
@@ -1,66 +1,50 @@
...
 import ReactDOM from 'react-dom';
+import Config from 'lib/Config';
 import { Provider } from 'react-redux';
 import configureStore from 'state/configureStore';
 import AssetAdminContainer from 'containers/AssetAdmin/AssetAdmin';
 import backend from 'lib/Backend';
 
 $.entwine('ss', () => {
-  $('.asset-gallery-component-wrapper').entwine({
+  $('#assetadmin-cms-content .cms-content-view').entwine({
```
tractorcow: This needs to go all the way up to `#assetadmin-cms-content` otherwise you won't be able to react-control breadcrumbs. I've checked campaign admin and this is the same level that area is rooted.

code/Controller/AssetAdmin.php
```diff
@@ -78,335 +75,405 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
     );
 
+    /**
+     * Set up the controller
+     */
+    public function init()
+    {
+        parent::init();
+
+        Requirements::javascript(FRAMEWORK_DIR . "/admin/client/dist/js/bundle-lib.js");
```
tractorcow: This file is already included in the `parent::init`, but also has the `provides`, so you don't need the extra bundle-lib here.

code/Controller/AssetAdmin.php
```diff
@@ -78,335 +75,405 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
+        Requirements::add_i18n_javascript(ASSET_ADMIN_DIR . '/client/lang', false, true);
+        Requirements::javascript(ASSET_ADMIN_DIR . "/client/dist/js/bundle.js");
+        Requirements::css(ASSET_ADMIN_DIR . "/client/dist/styles/bundle.css");
 
+        CMSBatchActionHandler::register('delete', 'SilverStripe\AssetAdmin\BatchAction\DeleteAssets', 'Folder');
+    }
 
 	public function getClientConfig() {
+        $baseLink = $this->Link();
```
tractorcow: Mixed indentation here. :) Set your IDE to respect .editorconfig?

tractorcow: Ok, it looks like the new code = correct, and there is some remaining incorrect indentation. Still, worth a reformat if you can help it.

code/Controller/AssetAdmin.php
```diff
@@ -78,335 +75,405 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
+                $items[] = $this->getObjectFromData($file);
+            }
+        }
+
+        $response = new SS_HTTPResponse();
+        $response->addHeader('Content-Type', 'application/json');
+        $response->setBody(json_encode([
+            'files' => $items,
+            'count' => count($items),
+            'parent' => $folder->ParentID, // grandparent
```
flamerohr: This might need a check on $folder as well?
`$folder ? $folder->ParentID : 0`

code/Controller/AssetAdmin.php
```diff
@@ -78,335 +75,405 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
+                'method' => 'get',
+                'responseFormat' => 'json',
+            ],
+            'searchEndpoint' => [
+                'url' => Controller::join_links($baseLink, 'api/search'),
+                'method' => 'get',
+                'responseFormat' => 'json',
+            ],
+            'updateFileEndpoint' => [
+                'url' => Controller::join_links($baseLink, 'api/updateFile'),
```
tractorcow: I'm not sure I'm happy 100% with this; having an API is fine, but we've developed the form schema for this.

Is it going to be ok for me to migrate file editing to the form schema, once I've been able to complete that work? I would leave the other endpoints in place.

I'd probably not try to apply form schema to actions that are simply uploads or typing out new folder names, however.

code/Controller/AssetAdmin.php
```diff
@@ -78,335 +75,405 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
 
     /**
-     * Gets the ID of the folder being requested.
+     * @param SS_HTTPRequest $request
      *
-     * @return int
+     * @return SS_HTTPResponse
      */
-    public function getCurrentFolderID()
+    public function apiUpdateFile(SS_HTTPRequest $request)
```
tractorcow: Continuing from the comment I left above, I would prefer this action was wrapped in a Form instance, and invoked via the form schema.

code/Controller/AssetAdmin.php
```diff
@@ -78,335 +75,405 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
+            "ParentID" => $record->ParentID,
+            "ID" => $record->ID,
+            "Filename" => $record->Filename,
+        ];
 
-        return $this->redirect(Controller::join_links($this->Link('show'), $parentID ? $parentID : 0));
+        return (new SS_HTTPResponse(json_encode($result)))->addHeader('Content-Type', 'application/json');
+    }
+
+    public function legacyRedirectForEditView($request)
```
tractorcow: PHPDoc please. :D

code/Controller/AssetAdmin.php
```diff
@@ -78,335 +75,405 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
-        $this->extend('updateEditForm', $form);
+        $result = [$this->getObjectFromData($file)];
 
-        return $form;
+        return (new SS_HTTPResponse(json_encode($result)))
+            ->addHeader('Content-Type', 'application/json');
     }
 
-    public function addfolder($request)
+    public function apiCreateFolder(SS_HTTPRequest $request)
```
tractorcow: Each of these api* methods will need phpdoc.

code/Controller/AssetAdmin.php
```diff
@@ -673,4 +580,182 @@ public function providePermissions()
...
+
+    /**
+     * The form is used to generate a form schema,
+     * as well as an intermediary object to process data through API endpoints.
+     * Since it's used directly on API endpoints, it does not have any form actions.
+     *
+     * @param Folder
+     * @return Form
+     */
+    protected function getFolderEditForm(Folder $folder)
```
tractorcow: Note to self; Candidate for form schema-izeation, maybe?

code/Controller/AssetAdmin.php
```diff
@@ -673,4 +580,182 @@ public function providePermissions()
...
+            'created' => $file->Created,
+            'lastUpdated' => $file->LastEdited,
+            'owner' => null,
+            'parent' => null,
+            'attributes' => array(
+                'dimensions' => array(),
+            ),
+            'title' => $file->Title,
+            'type' => $file->is_a('Folder') ? 'folder' : $file->FileType,
+            'category' => $file->is_a('Folder') ? 'folder' : $file->appCategory(),
```
tractorcow: Probably better to override these properties in Folder.

code/Controller/AssetAdmin.php
```diff
@@ -673,4 +580,182 @@ public function providePermissions()
...
+    protected function getObjectFromData(File $file)
+    {
+        $object = array(
+            'id' => $file->ID,
+            'created' => $file->Created,
+            'lastUpdated' => $file->LastEdited,
+            'owner' => null,
+            'parent' => null,
+            'attributes' => array(
+                'dimensions' => array(),
```
tractorcow: `$file->getDimensions('array')` ?

code/Controller/AssetAdmin.php
```diff
@@ -673,4 +580,182 @@ public function providePermissions()
...
+        if ($parent) {
+            $object['parent'] = array(
+                'id' => $parent->ID,
+                'title' => $parent->Title,
+                'filename' => $parent->Filename,
+            );
+        }
+
+        /** @var File $file */
+        if ($file->hasMethod('getWidth') && $file->hasMethod('getHeight')) {
```
tractorcow: File always has this method. Use `if($files->getIsImage())` instead.

client/src/state/gallery/GalleryActions.js
```diff
@@ -9,7 +9,7 @@ export function setFolder(folderId) {
   return (dispatch) => {
     dispatch({
       type: GALLERY.SET_FOLDER,
-      payload: { folderId },
+      payload: { ID: folderId },
```
tractorcow: Should we use `id` for lower camel case convention?

tractorcow:
Not a bug, but a usability concern; I notice that `HTTP GET` methods are passing in the securityID. SecurityID is ignored in non-write actions, so it's misleading to have this passed in for such operations.

Having it included in all requests, including non-destruction actions, introduces an unnecessary appearance of reliance on this token.

As a general rule, if your action is destructive, it shouldn't be HTTP GET in the first place, and pass the CSRF via the request body / headers. Thus, I wouldn't except to see CSRF in any querystring, legacy code temporarily excused. :)

code/Controller/AssetAdmin.php
```diff
@@ -78,335 +75,405 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
-        return Controller::join_links(
-            $this->Link('show'),
-            $folderID,
-            'edit',
-            $fileID
-        );
+
+        $folderID = (int)$params['id'];
+        if ($folderID === 0) {
+            $folder = new Folder();
```
tractorcow: Maybe better to leave this null and take up @flamerohr's suggestion. We don't need an empty object.

client/src/boot/index.js
chillu: Good catch - I'm doing this as part of the react-router refactoring, since react-router needs control over the whole right hand side for all React powered sections. I don't think this blocks the merge of this PR?

client/src/state/gallery/GalleryActions.js
chillu: It's a bit of a mess in there at the moment - I'm trying to use DataObject property naming as a first pass (`ParentID` rather than `folderId`), and where appropriate I'm using the uppercase convention. We'll need to go through the entire codebase for the lowercase work anyway, so there's not much point half doing it at the moment. This particular instance is just a side effect of making the API responses consistent.

code/Controller/AssetAdmin.php
chillu: Done anyway, had Atom set up with editorconfig, but not PHPStorm :)

code/Controller/AssetAdmin.php
chillu: Just moved this existing function further down in the code, but yep, I'll add some.

code/Controller/AssetAdmin.php
chillu: I'm not using this method yet anywhere BTW

code/Controller/AssetAdmin.php
chillu: I've migrated the code as-is. Do you want to raise this as a separate PR? It's a bit tricky to resolve, since the React code does checks like `item.type === 'folder'`. That's fine for this particular case since `folder` is hardcoded above, but `File->getFileType()` returns an abitrary, localised string which shouldn't be used for comparisons. So I could override `Folder->getFileType()` with `return 'folder'`, which seems inconsistent. Or I could use `return _t('Folder.SINGULARNAME')`, which we can't use for comparisons.

We shouldn't really use `Folder.ClassName` comparisons on the client, since you could create `Folder` subclasses. Does replacing `item.type === 'folder'` with `item.category === 'folder'` make sense? Or do we need a new property?

code/Controller/AssetAdmin.php
chillu: Same thing, that's existing code. But I'll fix it up.

code/Controller/AssetAdmin.php
chillu: It's the existing approach. Will be fixed through https://github.com/silverstripe/silverstripe-asset-admin/issues/171

chillu:
Agreed that readonly requests shouldn't have CSRF, that was an oversight, applied it too broadly. Fixed now.

I think I've addressed all your feedback now, thanks for reviewing! Wanna check?

PR #208, opened 2016-07-29
Fix upload saving dynamic class

flamerohr:
Saving all uploads as `File` class break images in particular.


code/Controller/AssetAdmin.php
```diff
@@ -371,8 +371,8 @@ public function apiCreateFile(SS_HTTPRequest $request)
         }
 
         // TODO Allow batch uploads
-
-        $file = File::create();
+        $fileClass = File::get_class_for_file_extension(File::get_file_extension($tmpFile['name']));
+        $file = new $fileClass();
```
tractorcow: Can you please use Injector for this? `Injector::inst()->create($fileClass);`

PR #210, opened 2016-08-02
Create hr.yml

hdpero:
Croatian translation for label.


tractorcow:
In future can you please update at https://www.transifex.com/silverstripe/silverstripe-asset-admin/translate/#hr/master/84008810 instead of making a pull request? I've manually updated this translation.

PR #214, opened 2016-08-09
BUG Remove copy of duplicate files from CMS module

tractorcow:
Dependency of https://github.com/silverstripe/silverstripe-cms/issues/1572

When installed alongside cms module, composer would load the incorrect behat fixtures for running CMS tests.

If we activate behat tests on this module, then we should add back any necessary fixture classes in the correct (and distinct) namespace.


hafriedlander:
Do we know why we aren't running the behat tests as part of CI now?

tractorcow:
I can't see if we ever were sorry.

PR #215, opened 2016-08-09
Switch Editor form to use FormBuilder

flamerohr:
Relies on https://github.com/silverstripe/silverstripe-framework/pull/5880

Switching from the previous practice of building the form manually to using the FormBuilderComponent in framework.


tractorcow:
When I save a file I get `Uncaught TypeError: e is not a function`.

The "Save" text on the save button is missing.

code/Controller/AssetAdmin.php
```diff
@@ -604,19 +615,83 @@ protected function getFolderEditForm(Folder $folder)
...
     }
 
     /**
+     * @param SS_HTTPRequest $request
+     *
+     * @return SS_HTTPResponse
+     */
+    public function FileEditForm(SS_HTTPRequest $request)
+    {
+        parse_str($request->getBody(), $data);
```
tractorcow: You shouldn't need to do this; Just get the Id from the request and populate the form from that.

See what we do in CampaignAdmin.php


```php
public function DetailEditForm($request) {
		// Get ID either from posted back value, or url parameter
		$id = $request->param('ID') ?: $request->postVar('ID');
		return $this->getDetailEditForm($id);
	}
```

code/Controller/AssetAdmin.php
```diff
@@ -46,6 +50,10 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
 
     private static $tree_class = 'Folder';
 
+    private static $edit_thumbnail_width = 380;
```
chillu: This is inconsistent with the already existing `File.cms_thumbnail_width`. There's also `File.asset_preview_width`. Both of these are defined on the `ImageManipulation` trait, but apply to the `File` class. `File.asset_preview_width` is `400` instead of `380`, but really it should just be "a value that's bigger than the required preview, but smaller than a potentially huge source image". Browsers are pretty good at smooth down sampling these days. I'd suggest we use `File.asset_preview_width`, and either make sure it works with `400` as well as any other values which might have been configured. Or serve the image with `400`, but code the `` tag to 380 via the React component (as a prop)

code/Controller/AssetAdmin.php
```diff
@@ -601,22 +616,108 @@ protected function getFolderEditForm(Folder $folder)
...
+        $file = $this->getList()->byID($id);
+
+        // Get thumbnail, CMSThumbnail() is smaller than what we'd like
+        $thumbnailWidth = $this->config()->edit_thumbnail_width;
+        $thumbnailHeight = $this->config()->edit_thumbnail_height;
+
+        // todo: use getCMSFields
+        $fields = FieldList::create([
+            HeaderField::create('TitleHeader', $file->Title, 1),
+            LiteralField::create("ImageFull", $file->ThumbnailIcon($thumbnailWidth, $thumbnailHeight)),
```
chillu: Should use `PreviewThumbnail()` without args, which uses the existing `File.asset_preview_width` (see comment above)

code/Controller/AssetAdmin.php
```diff
@@ -601,22 +616,108 @@ protected function getFolderEditForm(Folder $folder)
...
+    public function getFileEditForm($id)
     {
+        /** @var File $file */
+        $file = $this->getList()->byID($id);
+
+        // Get thumbnail, CMSThumbnail() is smaller than what we'd like
+        $thumbnailWidth = $this->config()->edit_thumbnail_width;
+        $thumbnailHeight = $this->config()->edit_thumbnail_height;
+
+        // todo: use getCMSFields
```
chillu: Minor note: Please use `TODO` (all uppercase, no double colon) to denote TODOs - it's pretty standard, and gets picked up by IDEs

code/Controller/AssetAdmin.php
flamerohr: Thanks, I wasn't aware of which of the Thumbnail functionality to use, there were a few.

I'll switch to `PreviewThumbnail()` and let `File.asset_preview_width` determine what the width is and it'll only be one place to change it all.

code/Controller/AssetAdmin.php
flamerohr: done, I've changed casing to uppercase

code/Controller/AssetAdmin.php
```diff
@@ -601,22 +612,104 @@ protected function getFolderEditForm(Folder $folder)
...
+        $fields = FieldList::create([
+            HeaderField::create('TitleHeader', $file->Title, 1),
+            LiteralField::create("ImageFull", $file->PreviewThumbnail()),
+            TextField::create("Title", _t('AssetTableField.TITLE','Title')),
+            TextField::create("Name", _t('AssetTableField.FILENAME','Filename')),
+            HiddenField::create('ID', $id),
+        ]);
+        if ($file->getIsImage()) {
+            $size = sprintf('%spx, %s', $file->getDimensions(), $file->getSize());
+            $fields->push(ReadonlyField::create("DisplaySize", "Size", $size));
```
chillu: Needs `_t()`

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -33,7 +33,6 @@ class AssetAdmin extends SilverStripeComponent {
       createFolderApi: this.createEndpoint(sectionConfig.createFolderEndpoint),
       readFolderApi: this.createEndpoint(sectionConfig.readFolderEndpoint, false),
       updateFolderApi: this.createEndpoint(sectionConfig.updateFolderEndpoint),
-      updateFileApi: this.createEndpoint(sectionConfig.updateFileEndpoint),
```
chillu: We should remove the server endpoint as well then (`AssetAdmin->apiUpdateFile()`).

client/src/containers/Editor/Editor.js
```diff
@@ -1,280 +1,63 @@
...
-
-            <div className="btn-toolbar">
-
-              <div className="btn-group hidden-xs-up" role="group" aria-label="">
-                <FormAction
-                  type="submit"
-                  bootstrapButtonStyle="primary"
-                  icon="rocket"
-                  handleClick={this.onFilePublish}
-                  loading={this.state.isSaving}
```
chillu: The save button no longer has a progress indication - having a look now.

client/src/containers/Editor/Editor.js
```diff
@@ -1,280 +1,63 @@
...
-        <ul className="nav nav-tabs hidden-xs-up" role="tablist">
-          <li className="nav-item">
-            <a className="nav-link active" data-toggle="tab" href="#file-details" role="tab">Details</a>
-          </li>
-          <li className="nav-item">
-            <a className="nav-link" data-toggle="tab" href="#file-usage" role="tab">Usage</a>
-          </li>
-        </ul>
-
-        <div className="tab-content">
```
chillu: I've noted all that removed markup on the issue which plans to implement bootstrap tabs: https://github.com/silverstripe/silverstripe-framework/issues/5612

client/src/containers/Editor/Editor.js
chillu: Chris is tackling that in as part of https://github.com/silverstripe/silverstripe-framework/issues/5613

PR #220, opened 2016-08-16
Namespace Admin module

tractorcow:
https://github.com/silverstripe/silverstripe-framework/issues/5878

Adds in missing psr-4 autoload configuration.


dhensby:
I merged this by hand but I couldn't get the bundle.js to build :/ hopefully it doesn't break anything (I just accepted the current master version)

PR #222, opened 2016-08-17
Show breadcrumbs that went missing

flamerohr:


client/src/containers/Gallery/Gallery.scss
```diff
@@ -1,8 +1,8 @@
 .gallery {
   width: 100%;
-  height: 100%;
+  height: calc(100% - 53px);
   position: absolute;
-  top: 0; // Temp: for legacy markup
+  top: 53px;
```
chillu: No magic pixel values please. At the very least, we should create a variable for `$toolbarNorthHeight`.

@clarkepaul Have you investigated a flexbox layout for this? We need to support IE10 (partially supports flexbox, see http://caniuse.com/#feat=flexbox), but are using https://github.com/postcss/autoprefixer which should take care of the syntax differences there.

scott1702: Since asset-admin requires framework's variables we should be able to use $toolbar-total-height for now to get breadcrumbs back up. (https://github.com/silverstripe/silverstripe-framework/blob/2ea65e90645b599f2d43da4e16d99f397302c1de/admin/client/src/styles/_variables.scss#L409)

scott1702:
This can be closed now https://github.com/silverstripe/silverstripe-asset-admin/pull/222 is merged.

Issue #227, opened 2016-08-18
"delete" label on batch actions is missing


chillu:
![image](https://cloud.githubusercontent.com/assets/111025/17789337/f547bef8-65e5-11e6-99db-69ffabebf8da.png)

Same with the dialog confirmation.

Pointer: There's language files in `client/lang`

chillu:
Fixed by https://github.com/silverstripe/silverstripe-asset-admin/pull/229/

Issue #224, opened 2016-08-18
File edit sidebar is not keyboard accessible


chillu:
Somehow the keyboard focus state gets cancelled out when going through the gallery images before it can reach the file edit form (in Chrome 52)

clarkepaul:
I think from memory it is tab accessible but it is before the gallery when tabbing through.

chillu:
Fixed by https://github.com/silverstripe/silverstripe-asset-admin/issues/229

PR #229, opened 2016-08-19
ISSUE-224: Fix bug causing editor to close when using keyboard navigation

scott1702:
Fix for https://github.com/silverstripe/silverstripe-asset-admin/issues/224

I also pulled in the fix at https://github.com/silverstripe/silverstripe-asset-admin/pull/222 because it was annoying me :)


client/src/constants/index.js
```diff
@@ -7,9 +7,11 @@ export default {
   BULK_ACTIONS: [
     {
       value: 'delete',
-      label: i18n._t('AssetAdmin.BULK_ACTIONS_DELETE'),
+      label: i18n._t('AssetAdmin.BULK_ACTIONS_DELETE', 'Delete'),
```
scott1702: Also—I'm not quite sure why, but i18n is just pulling in empty strings in this file so I added a placeholder here as a temporary fix.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -308,18 +308,11 @@ GalleryItem.propTypes = {
...
-  spaceKey: React.PropTypes.number,
-  returnKey: React.PropTypes.number,
   handleActivate: React.PropTypes.func.isRequired,
   handleToggleSelect: React.PropTypes.func.isRequired,
   handleDelete: React.PropTypes.func.isRequired,
   messages: React.PropTypes.array,
   uploading: React.PropTypes.bool,
 };
 
-GalleryItem.defaultProps = {
```
chillu: Thanks for getting rid of that, it's much better kept in constants - can't think of why you'd want to customise that on a component instance :)

client/src/constants/index.js
chillu: Sweet, this should fix https://github.com/silverstripe/silverstripe-asset-admin/issues/227

Issue #226, opened 2016-08-18
"Selected" style no longer coming through on file when in "edit file" mode


chillu:
See designs with blue border and background on https://projects.invisionapp.com/boards/9GX68PTXUQFH/#/1547737/48578214. If you click on a file that's no longer the case.

Pointer: Check `admin/client/containers/Gallery.js/scss`

clarkepaul:
Fixed

clarkepaul:
Issue still exists

flamerohr:
Is there business rules around how multiple files are highlighted? that's different to how files are selected?
e.g. Holding down shift and click on item?

Currently, I can see this highlighting work for one File (that's currently open), but maybe I'm missing context for how highlighting more than one.

clarkepaul:
Nope there is no "hold shift and select" or any other fancy business yet, not unless it was built into dropzone and I haven't noticed it. 
Checked checkbox = border (multiple items can be checked)
Selected = background and border (only one item and be selected at a time)

There will be some more fancy stuff coming probably in Alpha3 or Beta with drag and drop and perhaps the ability to select multiple items to edit but not yet.

flamerohr:
I've removed the "highlighted_files" feature for now
https://github.com/open-sausages/silverstripe-asset-admin/commit/efafabc0d850b4c9f6cefc9193a7394e6fd37060

This can be re-implemented later when edit multiple items is implemented.

flamerohr:
Pull requests made:
https://github.com/silverstripe/silverstripe-asset-admin/pull/255
https://github.com/silverstripe/silverstripe-framework/pull/6034

tractorcow:
https://github.com/silverstripe/silverstripe-asset-admin/pull/255 is merged, and I've left some feedback on https://github.com/silverstripe/silverstripe-framework/pull/6034

PR #230, opened 2016-08-19
ISSUE-226: Fix files not being highlighted when selected

scott1702:
https://github.com/silverstripe/silverstripe-asset-admin/issues/226


chillu:
This has been superseded by the react-router work, with does away with `setFile()` in favour of routing params (and `router.push()` calls). Thanks for your work anyway Scott, sorry we didn't look at it earlier!

PR #231, opened 2016-08-22
BUG Creating folder doesn't refresh (fixes #209)

### LINKED ISSUE #209: BUG Creating folder doesn't refresh UI
After creating a folder the json response for the new folder is returned, but the UI won't display this until a refresh.

`{"ParentID":0,"ID":3,"Filename":"Uploads\/"}` for instance.

I would expect "Uploads" to appear beside the folder here.

image

chillu:
Regression from react-router work. Needed to factor out Gallery.js promptFolderName() in order to make it mockable. Also fixed _t() call use, and removed the leftover `SET_FOLDER` and `SET_FILE` actions.

Uses a promise returned from a thunk in order to hook into the createFolder logic. This is a technique documented on https://github.com/gaearon/redux-thunk. Ideally this view refresh would happen through action creator chaining, but that's not possible because the `SET_FOLDER` action can no longer be used alongside the route param (see https://github.com/silverstripe/silverstripe-framework/issues/5711#issuecomment-240367840).


client/lang/en.js
```diff
@@ -40,6 +40,7 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "AssetAdmin.LASTEDIT": "Last changed",
     "AssetAdmin.LOADMORE": "Load more",
     "AssetAdmin.NOITEMSFOUND": "No items found",
+    "AssetAdmin.PROMPTFOLDERNAME": "Folder name (or blank to cancel)",
```
flamerohr: how about have this a bit more instructive like:
> Please enter a folder name (or blank to cancel)

client/src/containers/Gallery/Gallery.js
```diff
@@ -246,15 +246,18 @@ export class Gallery extends Component {
    *
    * @param object event - Click event.
    */
-  handleCreateFolder() {
+  handleCreateFolder(event) {
```
flamerohr: although it's not an issue due to the button being `type="button"`, add missing `event.preventDefault();` call?

client/src/containers/Gallery/Gallery.js
```diff
@@ -289,6 +292,13 @@ export class Gallery extends Component {
     }
   }
 
+	/**
+   * @return String
+   */
+  promptFolderName() {
+    return prompt(i18n._t('AssetAdmin.PROMPTFOLDERNAME'));
```
flamerohr: what is the benefit of moving this line to a new function? It seemed more intuitive to have this in the `handleCreateFolder` function because prompt is quite unique for javascript.

chillu: Yeah "test-driven apis" - not great, but I couldn't see any other way. `handleCreateFolderName()` is already fairly atomic as far as passed-in props go for the component.

client/src/containers/Gallery/Gallery.js
chillu: Good call, done!

client/lang/en.js
chillu: Just taken over the existing label, but changed it - thanks for your attention to detail there!

client/src/containers/Gallery/Gallery.js
tractorcow: Not a blocker for this PR, but can we make sure to include default text for JS translation strings here as well? It provides a reference for the user, as well as the to-be-developed js text collector.

client/src/containers/Gallery/tests/Gallery-test.js
```diff
@@ -55,292 +53,28 @@ describe('Gallery', () => {
     };
   });
 
-  describe('handleSuccessfulUpload', () => {
```
tractorcow: Did we need to delete all these tests?

PR #233, opened 2016-08-26
Edit folder form

flamerohr:
Can edit the current folder by the icon in the breadcrumbs or edit one selected folder with the batch action options (will be disabled for zero or more than one folders selected)

Merge with
 * https://github.com/silverstripe/silverstripe-framework/pull/5933
 * https://github.com/silverstripe/silverstripe-framework/pull/5929
 * https://github.com/silverstripe/silverstripe-framework/pull/5934

Once all of this is merged, we can remove AssetAdmin from the `cms` module: https://github.com/silverstripe/silverstripe-cms/pull/1590


chillu:
Breadcrumbs don't change when folder title change is saved

flamerohr:
Thanks, I've fixed the breadcrumbs not updating and removed the "More Options" PopoverField for folders

client/src/components/BulkActions/BulkActions.js
```diff
@@ -67,6 +74,15 @@ export class BulkActions extends SilverStripeComponent {
     return this.props.gallery.selectedFiles;
   }
 
+  getSelectedFolder() {
```
chillu: In general, please try to add more comments about the intent of your code. It's not immediate apparent why there's `getSelectedFiles()` (plural), but `getSelectedFolder()` (singular). In this case, the constraint is that we don't support multi-editing yet, and that's the only use case for `getSelectedFolder()`. I've added a comment myself for now.

client/src/constants/index.js
```diff
@@ -4,13 +4,20 @@ export default {
   CSS_TRANSITION_TIME: 300,
   THUMBNAIL_HEIGHT: 150,
   THUMBNAIL_WIDTH: 200,
-  BULK_ACTIONS: [
```
chillu: By changing this from an array to a fixed list of constants, you're removing the ability to define further bulk actions without rewriting half the `BulkActions` component. While we don't have fully configurable bulk actions yet, this moves us further away from this goal. I'm just having a look if there's any quick wins in refactoring this.

client/src/constants/index.js
```diff
@@ -4,20 +4,40 @@ export default {
...
-    disabled: false,
-  },
+  BULK_ACTIONS: [
+    {
+      value: 'delete',
+      label: i18n._t('AssetAdmin.BULK_ACTIONS_DELETE', 'Delete'),
+      className: 'font-icon-trash',
+      destructive: true,
+      callback: null, // defined in <Gallery> for now
+      confirm: () => {
```
flamerohr: Would we be able to convert this to boolean and BulkActions could provide a confirm handler?

It seems close to an anti-pattern to have a function contained inside a constants folder/file.

client/src/components/BulkActions/BulkActions.js
```diff
@@ -30,110 +27,95 @@ export class BulkActions extends SilverStripeComponent {
...
     }
 
-    if (option.destructive === true) {
-      // eslint-disable-next-line no-alert
-      if (confirm(i18n.sprintf(i18n._t('AssetAdmin.BULK_ACTIONS_CONFIRM'), option.label))) {
-        this.applyAction(option.value);
-      }
+    // Optionally execute confirmation logic (can be async)
+    if (typeof option.confirm === 'function') {
+      promise = option.confirm(this.props.items)
```
flamerohr: Do we need a `option.confirm` as well as `option.callback`?

This could be simplified to just calling `option.callback` and the option callback could handle the confirm in itself? Since they're both directly chained to each other.

client/src/constants/index.js
```diff
@@ -4,20 +4,40 @@ export default {
...
+          promise.reject();
+        }
+
+        return promise;
+      },
+    },
+    {
+      value: 'edit',
+      label: i18n._t('AssetAdmin.BULK_ACTIONS_EDIT', 'Edit'),
+      className: 'font-icon-edit-list',
```
flamerohr: This could probably siwtch to `font-icon-edit` since it could apply to both folders and files now

client/src/components/BulkActions/BulkActions.js
chillu: Hmm, you're right in a way, great that you picked that up - peer review in action! :) Buuuut: I think I've got a case for separating them - progress indicators. If we want the component to show its own bulk action progress at some point, we need to be able to distinguish between "waiting for user feedback" and "waiting for server to process your request". Does that make sense, or am I overthinking it?

client/src/constants/index.js
chillu: Good point! We've already had procedural logic in here (`_t()`) - I don't see much value in keeping the config declarative only (e.g. in JSON format). Different actions should be able to provide different confirm handlers, so a `<BulkActions confirmHandler=...>` won't cut it if we want to keep bulk actions flexible and have third party modules define their own ones. We could wrap this in `<BulkAction>` (singular) and have confirm handlers there. But that still leaves third party modules with the burden of hooking into the rendering process of the container element, in this case `<Gallery>`. Nested higher-order components from multiple third party modules around `<Gallery>` sound like a bit of a nightmare to debug, so I'm tending towards a more configuration-drive, declarative approach.

client/src/constants/index.js
chillu: Oh right, good point - changed locally now.

client/src/components/BulkActions/BulkActions.js
flamerohr: That makes perfect sense ;) I think it'll be good to keep this separated with that in mind, would it be alright to add a comment to provide context in future?

client/src/constants/index.js
flamerohr: Fair call, I hadn't considered the extensibility that a configuration approach would provide for later on.

code/Controller/AssetAdmin.php
```diff
@@ -603,6 +603,17 @@ public function getFileEditForm($id)
             FormAction::create('save', _t('CMSMain.SAVE', 'Save'))
                 ->setIcon('save')
         ]);
+
+        // Delete action
+        $actions->push(
+            FormAction::create(
```
flamerohr: Could include this in the $actions definition, and save a push

Issue #221, opened 2016-08-17
Upload indicator has red background


chillu:
![image](https://cloud.githubusercontent.com/assets/111025/17722841/671b0e62-6489-11e6-9d6e-08771e1fc023.png)

Related: Missing files have red checkmark (https://github.com/silverstripe/silverstripe-asset-admin/issues/218)

clarkepaul:
This isn't just a CSS issue from what i can tell. Dropzone should be creating a thumbnail which should be almost immediately usable, so I'm assuming we are applying a error class because the thumbnail generation isn't working. I don't know enough React to tamper with it unfortunately.

flamerohr:
Had a quick look at the dropzone and galleryitem components, and it looks like it checked if it exists in the database without checking if it is uploading
I'll push a fix up soon :)

flamerohr:
Added unit tests for file uploading, I ran into issues mocking images, so will need to research on how to do that...

Pull request is: https://github.com/silverstripe/silverstripe-asset-admin/pull/241

PR #241, opened 2016-09-02
Uploading preview adjustments

flamerohr:
fixes silverstripe/silverstripe-asset-admin#221

Also fixes:
* preview image size (was always blown up)
* overlay "view" showing for upload preview, which doesn't do anything.
* Added missing unit tests (incomplete, missing image loading unit test)
* `handleAddedFile` returns a promise, so it can be chained for unit testing (and if the caller wants to chain something else)

Found new issue `Gallery.handleCancelUpload()` is broken


client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -154,13 +154,17 @@ class GalleryItem extends SilverStripeComponent {
     return this.props.item.exists;
   }
 
+  uploading() {
```
chillu: Masking a prop in a trivial getter seems unnecessary here - it just makes the code harder to read. What's the rationale for this?

flamerohr: The intent was that the `exists()` function was already in place and I was originally tweaking that function instead.
Turned out it needed splitting out and I had found it easier to read `this.uploading()` alongside `this.exists()` as opposed to mixing `this.props.uploading` with `this.exists()`

chillu: Yep, makes sense

chillu:
> Uploading preview adjustments (fixes silverstripe/silverstripe-asset-admin#221)

Feedback on commit message: You can leave out the URL if its on the same repo, so this should be:

> Uploading preview adjustments (fixes #221)

Also, careful with line length on commit message first lines - it should be less than 100 chars or so, otherwise most git repo browsers (like github) will cut it

chillu:
Rebased on master and force pushed.

PR #242, opened 2016-09-02
Update to React v15.0

chillu:
Merge with https://github.com/silverstripe/silverstripe-framework/pull/5942
Required for https://github.com/silverstripe/silverstripe-framework/issues/5524


chillu:
Figured out this mystery Travis failure on https://travis-ci.org/silverstripe/silverstripe-asset-admin/jobs/156984381#L1425: It's using an outdated React library in `framework` (not `asset-admin`), since https://github.com/silverstripe/silverstripe-framework/pull/5942 isn't merged yet.

The tests are passing locally with the exact same NodeJS version, so I think this should be merged (hard to get tests green without creating temporary commits to change Travis environment constants to point to the branch).

Aside: We shouldn't rely `framework/node_modules` in `asset-admin`, but that's how the system was structured, and there's no easy way around this. Even if we publish every custom JS in `framework` as an NPM module, you'll still duplicate all this code between multiple modules, which would need to be resolved via `expose/external`, or create bloated bundle files with duplicates - so you're just shifting the problem from build time to run time.

PR #245, opened 2016-09-08
BUG Ensure that uploads of duplicate files detect existing filenames properly

tractorcow:


tractorcow:
There's a bug in core that needs fixing; Folder::validate() doesn't bypass extension checking, so it will fail validation in some cases.

sminnee:
> There's a bug in core that needs fixing; Folder::validate() doesn't bypass extension checking, so it will fail validation in some cases.

Fix merged in https://github.com/silverstripe/silverstripe-framework/pull/6056

sminnee:
Reviewing this code, it seems as though we've missed a feature in `framework:SilverStripe/Assets` which I've logged here: https://github.com/silverstripe/silverstripe-framework/issues/6063

However, this PR ensure that asset-admin does what it is supposed to: perform a context-aware `canEdit()` check.

PR #258, opened 2016-09-21
Bulk action slide bug

clarkepaul:
Fix for the slide of the bulk actions, it shouldn't slide sidewards

image


flamerohr:
Looks good, I think this could be merged in :)
The sliding sideways looks cool but yeah it shouldn't do that

PR #259, opened 2016-09-21
Slightly smoother transition on bulk action, tries to fix a slight judder

clarkepaul:
Slightly smoother transition on bulk action, tries to fix a slight judder. Also reducing box-shadow before entry and after leave to enhance effect. @scott1702 maybe you have time to review this, isn't needed for alpha2 so no rush.


clarkepaul:
@tractorcow smoothed out the transition of the bulk action panel in assetAdmin as we discussed if your interested to review, I marked it as alpha3 so not priority.

flamerohr:
looks much better, it does seem to jump snap 1 pixel smaller at the end of animation to add the shadow.

clarkepaul:
yeah I don't think the shadow in fading properly, 300% improvement though

PR #264, opened 2016-09-21
A mix of minor tidy up with AssetAdmin

clarkepaul:
Position of Gallery sort dropdown (slightly moved down to align with other toolbar actions)
pasted_image_22_09_16__9_54_am
pasted_image_22_09_16__9_19_am

Add white background to smaller thumbnails (image which do not fill normal thumbnail space)
pasted_image_22_09_16__9_55_am
pasted_image_22_09_16__9_19_am

Editor dimension styles
pasted_image_22_09_16__10_00_am
pasted_image_22_09_16__9_19_am

Improve variables for spacing
#
Requires https://github.com/silverstripe/silverstripe-framework/pull/6054


clarkepaul:
Note: There is some style changes regarding the hover state of the Edit panel preview as it used to have an icon and a link which opened the image at full size in a new tab. The link was removed prior to the React conversion but the styles still exist. For now it is hidden—I commented out display: block but have left in there.

PR #267, opened 2016-09-22
FIX: Break dist javascript and onto newlines.

chillu:


sminnee:
Fine by me (as a squash) once tests pass.

PR #270, opened 2016-09-22
Updated loadImage to scale down the preview image

flamerohr:
So it doesn't take a crazy amount of time before uploading starts :D (due to how large dataURLs are for images


tractorcow:
I've tested and approved this PR. Once build is green it can be merged.

PR #272, opened 2016-09-22
FIX show correct error message on top of gallery item

flamerohr:
Rather than a generic "An error has occurred"


flamerohr:
Reference:
Pull request https://github.com/silverstripe/silverstripe-asset-admin/pull/270 was merged without the commit which contained this fix and was related to that.

chillu:
Unassigning myself to focus on redux-form

PR #273, opened 2016-09-22
BUG Added logic for current folder that could be deleted

flamerohr:
Editing current folder is available by an icon in the breadcrumbs

### DIFF
--- a/code/Controller/AssetAdmin.php
+++ b/code/Controller/AssetAdmin.php
@@ -204,6 +204,7 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
             array_unshift($parents, [
                 'id' => $next->ID,
                 'title' => $next->getTitle(),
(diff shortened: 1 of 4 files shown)


code/Controller/AssetAdmin.php
```diff
@@ -220,6 +222,11 @@ public function apiReadFolder(HTTPRequest $request)
             'title' => $folder->getTitle(),
             'count' => count($items),
             'parents' => $parents,
+            'parent' => [
```
tractorcow: This assumes $parent isn't null. What if $folder was a root folder?

tractorcow:
@flamerohr can you please review my changes?

flamerohr:
@tractorcow Your changes look good, I've noticed that the AssetAdmin container doesn't have any unit tests, which is of concern

PR #276, opened 2016-09-26
Switched to SASSLint

chillu:
See https://github.com/silverstripe/silverstripe-framework/pull/6016


chillu:
I've removed the CodeClimate webhook now, so the next build shouldn't trigger it

tractorcow:
Testing :)

Issue #237, opened 2016-08-30
Image thumbnails are loading in full size (not resampled)


chillu:
They've got the `_Resampled` suffix, but are in fact the original size - that's a showstopper for performance, since it'll load megabytes of image data - particularly until we implement infinite scrolling or pagination.

chillu:
Confirmed this is still an issue, even with https://github.com/silverstripe/silverstripe-asset-admin/issues/248 merged.

flamerohr:
Pull request https://github.com/silverstripe/silverstripe-asset-admin/pull/283

PR #283, opened 2016-09-30
Adjust thumbnail size for gallery items

flamerohr:
This is to improve loading speed and user experience, so for images it no longer downloads the full image.

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/237


clarkepaul:
Peer review:
The output file size is good for thumbnails and works fine in IE10. As far as code goes this should have a quick glance over by someone who can merge it, otherwise looks good to merge.

tractorcow:
One bugfix pushed up to ensure that missing files don't break the CMS. :)

PR #287, opened 2016-10-12
Use redux-form

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/5524
Merge with https://github.com/silverstripe/silverstripe-framework/pull/6172


sminnee:
Cool, so that removed schemaReducer is one concrete benefit of this: we have a single formschema reducer that handles all formbuilder-based forms?

chillu:
> we have a single formschema reducer that handles all formbuilder-based forms?

No, the `schemaReducer.js` was just moved from asset-admin to framework, and left duplicated in here. Schemas still need to be stored in the state tree. But we've gotten rid of `formsReducer.js` in framework, since all that heavy lifting is now handled by redux-forms (in the `reduxForm` reducer namespace)

Issue #253, opened 2016-09-15
Tidy up structure of File's formschema


clarkepaul:
PR:
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/6178
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/288

## Acceptance criteria:

 * File's form schema doesn't embed a tabset at the top level.
 * There's a way of extending File's detail fields. It's not `File::updateCMSFields()` – it's an object provided by asset-admin (probably a Form)
 * You are still able to create custom tabs in the 2nd half of the form.

To consider:

 * Proof of Concept - scaffolder for Controller-Model relationship which generates the form
 * Should the field list provided for editing files be a separate class within asset-admin, rather than a method on the File object?

----

Was: Additional unwanted tab panel hidden within Edit panel

Not sure why there would need to be tab panel surrounding everything in the edit panel, this doesn't effect and visuals to my knowledge but could potentially break stuff. The inner nested tab set is good though with "Details" and "Usage".

![image](https://cloud.githubusercontent.com/assets/555033/18539964/5a337390-7b6f-11e6-9a7d-a03e71144b54.png)

## Tasks

- [x] Remove top-level tabset from form file edit schema
- [x] Extension hook for editing form file edit schema on AssetAdmin
- [x] Create class which will scaffold form fields
- [x] map out ideas for Injector

sminnee:
Yeah, this is a hangover of SS3 applied to the new react form. The `getCMSFields()` content for File details shouldn't have forms in there just because we had that in SS3.

sminnee:
Assigning this to the react-forms epic as it kind of impacts how we think about the FormSchemas provided by the files.

tractorcow:
PR created: See top comment. :)

tractorcow:
Updated based on changes to framework PR.
See https://github.com/silverstripe/silverstripe-framework/pull/6178#issuecomment-254092537

sminnee:
This is done now.

tractorcow:
Thanks very much @sminnee !

PR #288, opened 2016-10-13
API Update to use FormBuilder API

tractorcow:
Requires https://github.com/silverstripe/silverstripe-framework/pull/6178

Implements https://github.com/silverstripe/silverstripe-asset-admin/issues/253


code/Controller/AssetAdmin.php
```diff
@@ -564,6 +570,25 @@ public function providePermissions()
...
 
     /**
+     * Build a form scaffolder for this model
+     *
+     * NOTE: Volatile api. May be moved to {@see LeftAndMain}
+     *
+     * @param File $file
+     * @return FormBuilder
+     */
+    public function getFormBuilder(File $file)
```
sminnee: We could do this the other way around. Rather than calling back to Injector with 3 sightly arbitrary service names, we could inject 3 properties into AssetAdmin via its service definition: FolderFormBuilder, ImageFormBuilder, and FileFormBuilder.

That way, AssetAdmin isn't globally coupled to Injector. It would require that AssetAdmin is instantiated by Injector, but that's okay...

code/Forms/AssetFormBuilder.php
```diff
@@ -0,0 +1,121 @@
...
+use SilverStripe\Forms\FormAction;
+use SilverStripe\Forms\HeaderField;
+use SilverStripe\Forms\HiddenField;
+use SilverStripe\Forms\LiteralField;
+use SilverStripe\Forms\Tab;
+use SilverStripe\Forms\TabSet;
+use SilverStripe\Forms\DefaultFormBuilder;
+use SilverStripe\Forms\TextField;
+
+abstract class AssetFormBuilder extends DefaultFormBuilder
```
sminnee: I kind of think that this should be created from scratch and not use `getCMSFields()` or `DefaultFormBuilder`. It should just build a form, in 1 place. It will be clearer than having the logic scattered across 3 places.

Perhaps it can `use Extensible` so that people can add `updateFields` extensions to it?

code/Forms/AssetFormBuilder.php
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * @return FormAction
+     */
+    protected function getSaveAction()
+    {
+        $record = $this->getRecord();
+        $editing = $record->isInDB();
+        $canSave = $editing ? $record->canEdit() : $record->canCreate();
+        if ($canSave) {
+            return FormAction::create('save', _t('CMSMain.SAVE', 'Save'))->setIcon('save');
```
sminnee: There's no indication here of what `save()` should actually do. Needs more work, but eventually I'd expect a closure in here that:

 * Is passed the file object to operate on.
 * Throws ValidationException in the case of an issue
 * Does something yet-to-be-defined on success.

What kind of API is provided for successful actions to say what they should do is probably the most challenging bit.

tractorcow:
AssetFormFactory no longer extends DefaultFormFactory; It's an independent interface now.

code/Forms/AssetFormBuilder.php
tractorcow: Done.

code/Controller/AssetAdmin.php
```diff
@@ -491,7 +496,8 @@ public function getSearchContext()
             'audio' => _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.AppCategoryAudio', 'Audio'),
             'document' => _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.AppCategoryDocument', 'Document'),
             'flash' => _t(
-                'SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.AppCategoryFlash', 'Flash',
+                'SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.AppCategoryFlash',
```
sminnee: At some stage, it would be good to provide optimisation here such that if the form factory didn't vary by record, the same form schema could be used once for all records.

I've put suggestions in the formfactory interface to allow for this, but I think actually implementing it is a separate story as it would require a reasonably large asset-admin change beyond the scope of this one.

sminnee:
@tractorcow the tests are still using the old `getForm()` arguments.

tractorcow:
Sorry for neglecting to update these tests. It's now fixed.

Thanks for the review @sminnee :)

PR #289, opened 2016-10-13
API Enforce PSR-2 and js / scss linting

tractorcow:
I've renamed a few form names to match PSR-2 method name requirements, and I've updated travis config.

I've also updated the test fixtures with better namespaces; Fixtures for a test are now stored in a namespace of the same name as the test, rather than using `_` separated names.

Also included some fixes to js / scss linting to ensure that these checks are correctly validated by CI.


code/Controller/AssetAdmin.php
```diff
@@ -617,7 +619,7 @@ public function getFileEditForm($id)
      *
      * @return Form
      */
-    public function FileEditForm()
+    public function fileEditForm()
```
chillu: @tractorcow Hmm, it feels inconsistent to start using lowercase form names as a convention, without also changing FormField instance names (and subsequently DataObject database column names).

Function names in PHP are case insensitive (http://www.php.net/manual/en/functions.user-defined.php), so this won't cause breakages in subclasses. We can't rename it to `getFileEditForm()` since those semantics are different based on conventions set in `LeftAndMain`: `EditForm($request)` vs. `getEditForm($id, $fields)` (https://github.com/silverstripe/silverstripe-framework/blob/master/admin/code/LeftAndMain.php#L1487). That's always been confusing as hell - maybe we rename those to `getEditForm($request)` and `getEditFormForRecord($id, $fields)` as a new convention?

package.json
```diff
@@ -21,7 +21,9 @@
     "test": "NODE_PATH=\"./client/src:../framework/client/src:../framework/admin/client/src\" jest",
     "coverage": "jest --coverage",
     "lock": "npm-shrinkwrap --dev",
-    "lint": "eslint client/src & sass-lint -v"
+    "lint": "eslint client/src & sass-lint -v",
```
chillu: Shouldn't sneak this in a PSR-2 change ;)

tractorcow: I'll rename the PR. :)

chillu:
@tractorcow I've rebased your PR now, sorry for letting it get stale!

chillu:
I've also adapted the phpcs ruleset.xml from framework. Can't do the same with `composer run-script lint` shortcuts since the module isn't able to run standalone yet (with framework installed in `vendor/silverstripe/framework`)

chillu:
@tractorcow The camel caps PHPCS rule has been disabled in framework - I take this as a sign that we're not planning to fix UpperCamelCased methods anytime soon for PSR-2 compat? Or is it just going to take a bit longer given the complexities involved with existing (framework) rather than new (asset-admin) code? https://github.com/silverstripe/silverstripe-framework/blob/master/tests/phpcs/ruleset.xml#L8

chillu:
Once this is merged, we should rebase and merge https://github.com/silverstripe/silverstripe-asset-admin/pull/326

sminnee:
@chillu all rules that aren't able to be fixed automatically were disabled in the default rules.

The plan is to fix the other ones, rule by rule, modifying ruleset.xml as we go.

chillu:
We decided to keep the form name in UpperCamelCase, in order to avoid breaking HTML identifiers which are frequently used in CSS/JS. Only the form method name needs to be lowerCamelCase for PSR-2 compliance here.

tractorcow:
> We decided to keep the form name in UpperCamelCase, in order to avoid breaking HTML identifiers which are frequently used in CSS/JS. Only the form method name needs to be lowerCamelCase for PSR-2 compliance here.

In order to keep form names consistent with react I've reversed this in a new commit; We do not need to follow a particular case, but in the case of new sections based on react it makes sense to case these consistently between PHP and JS.

PR #291, opened 2016-10-18
API Deletes trigger archive instead of delete

tractorcow:
Merge with https://github.com/silverstripe/silverstripe-framework/pull/6195

### DIFF
--- a/code/Controller/AssetAdmin.php
+++ b/code/Controller/AssetAdmin.php
@@ -234,7 +234,7 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
             'parentID' => $folder->exists() ? $folder->ParentID : null, // grandparent
             'folderID' => $folderID,
             'canEdit' => $folder->canEdit(),
-            'canDelete' => $folder->canDelete(),
+            'canDelete' => $folder->canArchive(),
         ]));
 
         return $response;
@@ -292,7 +292,7 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
         }
 
         if (!min(array_map(function (File $file) {
-            return $file->canDelete();
+            return $file->canArchive();
         }, $files))) {
             return (new HTTPResponse(json_encode(['status' => 'error']), 401))
                 ->addHeader('Content-Type', 'application/json');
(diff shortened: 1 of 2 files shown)


code/Controller/AssetAdmin.php
```diff
@@ -288,15 +288,15 @@ public function apiDelete(HTTPRequest $request)
         }
 
         if (!min(array_map(function (File $file) {
-            return $file->canDelete();
+            return $file->canArchive();
```
flamerohr: Is this necessary? Seems like it's only for testing purposes and many other places still use `canDelete()`

sminnee: Only the test implementation is for testing purposes. `Versioned::canArchive()` provides the real version.

tests/php/AssetAdminTest.php
```diff
@@ -502,6 +502,14 @@ public function canDelete($member = null)
         }
     }
 
+    public function canArchive($member = null)
+    {
+        if ($this->owner->Name === 'disallowCanDelete.txt') {
+            return false;
+        }
+        return $this->owner->canDelete($member);
```
sminnee: Shouldn't this be calling the Versioned::canArchive() code, or returning `null` to result in a no-op?

code/Controller/AssetAdmin.php
flamerohr: Sorry, I jumped that assumption about testing :D
What I mean is shouldn't this switch to `canArchive()` carry through to the `canDelete()` calls in functions such as `getObjectFromData()`?
It seemed weird to have `canArchive()` in the `apiDelete()` function but the schemas still reflect `canDelete()`

tests/php/AssetAdminTest.php
tractorcow: This is a work around to an annoying issue with multiple extensions both applying a default method implementation.

a canArchive here would be an infinity loop.

code/Controller/AssetAdmin.php
tractorcow: Yes I should apply it there too.

Issue #174, opened 2016-05-31
Additional space if there are no folders


clarkepaul:
If there are no folder present there is an extra gap above files which should be removed.
![pasted_image_31_05_16__3_38_pm](https://cloud.githubusercontent.com/assets/555033/15662576/139f7636-2746-11e6-984b-da437641f367.png)

chillu:
Confirmed this is still the case

clarkepaul:
PR https://github.com/silverstripe/silverstripe-asset-admin/pull/301
and https://github.com/silverstripe/silverstripe-framework/pull/6289

PR #301, opened 2016-11-04
Remove additional spacing if no folders present

clarkepaul:
Removes additional spacing above gallery items if there are no folders. 
I've swapped out the layout to flexbox which removes the additional spacing.
Also added updates to some missing variables for file states to match page states.

shows on hover of the little circle
image

Requires: https://github.com/silverstripe/silverstripe-framework/pull/6289
Fixes: https://github.com/silverstripe/silverstripe-asset-admin/issues/174

Was:
pasted_image_4_11_16__12_33_pm

New:
pasted_image_4_11_16__12_32_pm

### DIFF
--- a/client/src/components/GalleryItem/GalleryItem.scss
+++ b/client/src/components/GalleryItem/GalleryItem.scss
@@ -2,7 +2,6 @@ $gallery-item-label-height: 40px;
 
 .gallery-item {
   position: relative;
-  float: left;
   margin: 0 $spacer-x $spacer-y 0;
   background-color: $white;
   border: 1px solid;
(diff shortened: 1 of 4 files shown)


client/src/components/GalleryItem/GalleryItem.scss
```diff
@@ -248,21 +247,37 @@ $gallery-item-label-height: 40px;
...
-  border: 1px solid #fff;
+  width: 8px;
+  height: 8px;
+  border: 1px solid $state-draft;
+  box-shadow: 0 0 1px .5px $white;
   border-radius: 100%;
+
+  &:hover::after {
+    top: -4px;
+    font-size: 10px;
```
flamerohr: Would it be better to store this as a variable or reference another font-size variable?
I can see this being reused somewhere else.

clarkepaul:
Updated font size to variable.

flamerohr:
Sweet, all good to merge, I think :)

Issue #144, opened 2016-05-06
Sort by title is not working in IE10


jeffreyguo:
Tested in IE10 on Win7, the sort by title for files are not working, while it works in Chrome.
![screen shot 2016-05-06 at 3 48 30 pm](https://cloud.githubusercontent.com/assets/6565447/15063903/51a89474-13a2-11e6-8a3e-f66313164d06.png)

chillu:
`SCRIPT5007: Unable to get property 'field' of undefined or null reference `

tractorcow:
Fix at https://github.com/silverstripe/silverstripe-asset-admin/pull/305

flamerohr:
As mentioned in the pull request, a fair amount of refactor happening and will have fixed this issue

PR #305, opened 2016-11-15
#144 Fix sort by in IE10

wilr:
IE10 does not support dataset. An alternative approach is required for IE.


tractorcow:
No changes to dist files?

wilr:
Yeah sorry I still getting used to including dist in commits, normally leave out to avoid conflicts. Checked out branch and ran `npm build`.

sminnee:
Yeah, we need dist files in the build if we want to avoid forcing *every* SilverStripe user to also install npm.

tractorcow:
@wilr I've only just given you access to the repo, but please use the open-sausages fork; This lets other teammembers rebase / rebuild your PR as necessary. :)

In the mean time, rebase please!

wilr:
Looks like some major refactoring since 42f91b46 so closing as I think the issue will be resolved.

Issue #294, opened 2016-10-26
History view on a file


sminnee:
As an author, I want to see the history of changes to a file, so that I can audit what was published on a given day, and who has made changes

## Acceptance criteria
- I can see the history of changes to a file in the UI
- I can see who made changes
- I can see old meta-data
- I can view old versions of a file only if this feature is enabled in my SilverStripe installation
- I can view the old versions of the file if I am logged in as an appropriate user
- I can see published vs unpublished versions
- Versions are sorted newest first

Exclusions

 - No compare mode
 - No re-sorting
 - No revert functionality
 - No pagination
 - Using graphql ( as we don't want to go down that rabbit hole just yet - Ingo)

## Tasks

- [x] Write behat test to open a file and view history list shows correct data, modifying a file updates the UI and on revert the user receives a message (assume revert logic / permissions is covered by unit level tests).
- [x] Write basic unit level for content API endpoint which loads the history data.
- [x] Build content API endpoint in asset-admin server side for providing a JSON object of data to React. 
- [x] Ensure history has correct permission filtering.
- [x] Ensure files with large histories do not cause excessive server load 
- [x] Build React Component(s) for history item, readonly form and DOM for history tab viewer
- [x] If previous thumbnail not stored by SilverStripe, display a notice message.
- [x] Ensure tests pass.
- [x] Smoke test on various cases (no history, recently uploaded file, recently changed file..)
- [x] Design review with Paul
- [x] Browser Test IE
- [x] Browser Test Chrome
- [x] Browser Test Firefox
- [x] Browser Test Safari

Pull request: https://github.com/silverstripe/silverstripe-asset-admin/pull/306

clarkepaul:
Design concept, feedback welcome https://projects.invisionapp.com/boards/5CX68PWMSYVK/

clarkepaul:
More advanced version based on feedback with functionality to restore and view the changes like you would with pages. https://projects.invisionapp.com/boards/BXX68PGVNWFR/

I'm not sure this version is a must have for the majority of people but would be a nice addition.

clarkepaul:
Just had a chat to @sminnee and he is good with the second design with revert. @wilr with the historic preview lets cap the max-height at 100px, the doc thumbs are 92px so they would look okay there as well.

I don't think the cancel button in my design is even necessary.

wilr:
[...]

PR #306, opened 2016-11-15
#294 Add history tab to assets

wilr:
See https://github.com/silverstripe/silverstripe-asset-admin/issues/294

For review, tests need to be finished. To run will need https://github.com/silverstripe/silverstripe-framework/pull/6316


client/src/components/HistoryItem/HistoryItem.js
```diff
@@ -0,0 +1,32 @@
...
+
+class HistoryItem extends SilverStripeComponent {
+
+  constructor (props) {
+    super(props)
+
+    this.handleClick = this.props.handleClick.bind(this)
+  }
+
+  getListClassName () {
```
flamerohr: We could remove this function, it seems to be used only here?

client/src/containers/HistoryList/HistoryList.js
```diff
@@ -0,0 +1,153 @@
+import $ from 'jQuery';
```
flamerohr: A few unused imported modules, does your IDE highlight ones that aren't used?

client/src/containers/HistoryList/HistoryList.js
```diff
@@ -0,0 +1,153 @@
...
+
+  componentDidUpdate(prevProps, nextState) {
+    this.refreshHistoryIfNeeded(prevProps);
+  }
+
+  /**
+   *
+   */
+  refreshHistoryIfNeeded (prevProps) {
+    const sectionConfig = Config.getSection('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin');
```
flamerohr: I wonder if there's a way for this to be accepted as a props, so it could be more general purpose (e.g. used for the Pages section in the future)

client/src/containers/HistoryList/HistoryList.js
```diff
@@ -0,0 +1,153 @@
...
+    this.props.historySchemaUrl = sectionConfig.form.FileHistoryForm.schemaUrl;
+
+    if (!prevProps || (prevProps.data.fileId !== this.props.data.fileId)
+      || !this.state.history || (prevProps.data.latestVersionId !== this.props.data.latestVersionId)) {
+
+      var self = this
+
+      let api = this.createEndpoint(sectionConfig.historyEndpoint);
+
+      api({fileId: this.props.data.fileId}).then(function(json) {
```
flamerohr: if you used es2015's fat arrow function `(json) => {` you won't need to define `self`

code/Controller/AssetAdmin.php
```diff
@@ -387,6 +401,96 @@ public function apiCreateFile(HTTPRequest $request)
...
+    public function apiHistory(HTTPRequest $request)
+    {
+        // CSRF check not required as the GET request has no side effects.
+        $fileId = $request->getVar('fileId');
+
+        if(!$fileId || !is_numeric($fileId)) {
+            return new HTTPResponse(null, 400);
+        }
+
+        $class = 'SilverStripe\\Assets\\File';
```
flamerohr: can use php5's `File::class` :D (this applies to other class name references as well)

code/Controller/AssetAdmin.php
```diff
@@ -387,6 +401,96 @@ public function apiCreateFile(HTTPRequest $request)
     }
 
     /**
+     * Returns a JSON array for history of a given file ID. Returns a list of all the history.
+     *
+     * @param HTTPRequest
+     *
+     * @return HTTPResponse
+     */
+    public function apiHistory(HTTPRequest $request)
```
flamerohr: We could pass the history data as part of the FormSchema. @tractorcow what do you think about a `HistoryViewerField` kind of class and corresponding React component?

code/Controller/AssetAdmin.php
```diff
@@ -664,6 +768,69 @@ public function FileEditForm()
     }
 
     /**
+     *
+     */
+    public function getFileHistoryForm($id)
```
flamerohr: I think this could stay though!

code/Controller/AssetAdminFile.php
```diff
@@ -16,4 +18,54 @@ public function updateCMSEditLink(&$link)
...
         $link = Director::absoluteURL($controller->getFileEditLink($this->owner));
     }
+
+    /**
+     * @param int $from
+     * @param int $to
+     *
+     * @return string
+     */
+    public function humanizedChanges($from, $to) {
```
flamerohr: maybe it's closer to be called `getVersionedDifferences()`?
This can return an array and React could implode it on the client-side

code/Forms/FileHistoryFormFactory.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace SilverStripe\AssetAdmin\Forms;
```
flamerohr: I wonder if we could Reuse the FileFormFactory and utilise the `$context` param or even call `setReadonly()` after getting the form.

code/Controller/AssetAdmin.php
tractorcow: Yes please; It needs to be a part of the state as the source of truth (along with schema). It also saves another request. :)

code/Controller/AssetAdmin.php
tractorcow: Or just `$this->config()->get('tree_class')` if you want to soft code it.

I'd rather opt as above to merge this into the form state though and not have this extra endpoint.

code/Controller/AssetAdminFile.php
```diff
@@ -16,4 +18,54 @@ public function updateCMSEditLink(&$link)
...
+        $output = array();
+
+        foreach($changes as $change) {
+            $human = $change;
+
+            if($change == "ParentID") {
+                // updated folder ID
+                $human = _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdminFile.MOVEDFOLDER', "Moved file");
+            } else if($change == 'Title') {
+                $human = _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdminFile.RENAMEDTITLE', "Updated title to ") . $fromRecord->Title;
```
tractorcow: Shouldn't these refer to the `$to` record not the `$from` record.

code/Controller/AssetAdminFile.php
```diff
@@ -16,4 +18,54 @@ public function updateCMSEditLink(&$link)
...
     }
+
+    /**
+     * @param int $from
+     * @param int $to
+     *
+     * @return string
+     */
+    public function humanizedChanges($from, $to) {
+        $fromRecord = Versioned::get_version($this->owner->class, $this->owner->ID, $from);
```
tractorcow: If there is no $fromRecord maybe a shortcode to `Initial creation` or something to that effect?

code/Controller/AssetAdminFile.php
tractorcow: That could work but I don't think it's necessary.

code/Forms/FileHistoryFormFactory.php
tractorcow: Isn't that's what @wilr is doing?

code/Forms/FileHistoryFormFactory.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $record = $context['Record'];
+
+        return new FieldList(
+            LiteralField::create('Thumbnail', $this->getIconMarkup($record)),
+            LiteralField::create('FileSpecs', $this->getSpecsMarkup($record)),
+            ReadonlyField::create("Title", File::singleton()->fieldLabel('Title')),
+            ReadonlyField::create('Name', File::singleton()->fieldLabel('Filename')),
+            ReadonlyField::create("Path", _t('AssetTableField.PATH', 'Path'), $this->getPath($record))
+        );
```
tractorcow: Please include an extension point here. :)

```php
$this->invokeWithExtensions('updateFormFields', $fields, $controller, $name, $context);
```

code/Forms/FileHistoryFormFactory.php
flamerohr: this is creating a new FormFactory rather than reusing the FileFormFactory, different to what I'm describing

tractorcow: Ah I see. Hm, I think it's ok to have a separate class for a separate use-case.

code/Controller/AssetAdmin.php
wilr: @tractorcow @flamerohr Ok, getting around to this feedback, I'm not sure how you would inject the history data into the FormSchema's `data` field? Looks impossible to set?

tractorcow: Hi @wilr I'll write a PoC for you to follow if you like.

tractorcow:
@wilr I've added you as a member to https://github.com/orgs/open-sausages/teams/sausage-masters so you can push your future PR up to our fork. :)

client/src/components/HistoryItem/HistoryItem.js
```diff
@@ -0,0 +1,32 @@
+import i18n from 'i18n'
```
chillu: Every component needs a `README`

client/src/components/HistoryItem/HistoryItem.js
```diff
@@ -0,0 +1,32 @@
...
+        data-id={this.props.item.id}
+        onClick={this.handleClick}>
+        <p className="history-item-version">v.{this.props.item.versionid}</p>
+        <p><small className="history-item--date">{this.props.item.date} {this.props.item.author}</small>{this.props.item.summary}<small><span className="editor__status-flag">{this.props.item.status}</span></small></p>
+      </li>
+    )
+  }
+}
+
+export default HistoryItem
```
chillu: Needs propTypes definitions

client/src/components/HistoryItem/HistoryItem.scss
```diff
@@ -0,0 +1,83 @@
...
+  position: relative;
+  min-height: 30px; /* back button */
+
+  &.history-container--loading {
+    width: 100%;
+    height: 80px;
+    background: url("../../images/loading.gif") no-repeat 50% 50%;
+  }
+
+  .editor__file-preview-link,
```
chillu: This isn't really BEM - can it be solved differently?

client/src/components/HistoryItem/HistoryItem.scss
```diff
@@ -0,0 +1,83 @@
...
+    background: $background-darker;
+    margin: 10px auto;
+  }
+
+  .editor__specs {
+    text-align: center;
+  }
+}
+
+.history-list {
```
chillu: Same here, not BEM

client/src/containers/HistoryList/HistoryList.js
```diff
@@ -0,0 +1,153 @@
...
+import { withRouter } from 'react-router';
+import { bindActionCreators } from 'redux';
+import ReactTestUtils from 'react-addons-test-utils';
+import Config from 'lib/Config';
+import CONSTANTS from 'constants/index';
+import * as galleryActions from 'state/gallery/GalleryActions';
+import HistoryItem from 'components/HistoryItem/HistoryItem';
+import FormBuilderLoader from 'containers/FormBuilderLoader/FormBuilderLoader';
+
+export class HistoryList extends Component {
```
chillu: Needs a `README`

client/src/containers/HistoryList/HistoryList.js
```diff
@@ -0,0 +1,153 @@
...
+  refreshHistoryIfNeeded (prevProps) {
+    const sectionConfig = Config.getSection('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin');
+    this.props.historySchemaUrl = sectionConfig.form.FileHistoryForm.schemaUrl;
+
+    if (!prevProps || (prevProps.data.fileId !== this.props.data.fileId)
+      || !this.state.history || (prevProps.data.latestVersionId !== this.props.data.latestVersionId)) {
+
+      var self = this
+
+      let api = this.createEndpoint(sectionConfig.historyEndpoint);
```
chillu: Shouldn't be created on every refresh

client/src/containers/HistoryList/HistoryList.js
```diff
@@ -0,0 +1,153 @@
...
+                   handleClick={this.handleClick.bind(this, history.versionid)} />
+          )
+       })}
+      </ul>
+    </div>
+    )
+  }
+}
+
+export default HistoryList;
```
chillu: propTypes please

client/src/containers/HistoryList/HistoryList.js
```diff
@@ -0,0 +1,153 @@
...
+  getContainerClassName () {
+    return (this.state.viewDetails && !this.state.loadedDetails) ? 'history-container history-container--loading' : 'history-container';
+  }
+
+  render () {
+    if (!this.state.history) {
+      return (
+      <div className={this.getContainerClassName()}>
+        <p>
+          {this.props.errorMessage}
```
chillu: Shouldn't this leave out the `<p>` tags if no error message is present?

client/src/containers/HistoryList/HistoryList.js
```diff
@@ -0,0 +1,153 @@
...
+        </p>
+      </div>
+      )
+    }
+
+    // TODO Use this.props.config once the store is consolidated with framework
+    const securityID = Config.get('SecurityID')
```
chillu: You can just get it from `state.config` through `mapStateToProps`, I don't think this needs any consolidation?

chillu:
@clarkepaul Was it your idea to add a `loading.gif` here? Seems like we have enough of those in core already :D Also, shouldn't we make this an SVG/CSS based animation, and keep it consistent with other components?

code/Controller/AssetAdmin.php
chillu: I disagree, we shouldn't try to cram more structured data into form schema - it's going to make it harder to move this over to GraphQL. Separate requests after returning a form schema will be unavoidable, e.g. we can't inline all data required for a React-powered GridField into form schema.

code/Controller/AssetAdmin.php
```diff
@@ -387,6 +401,96 @@ public function apiCreateFile(HTTPRequest $request)
...
+        if(!$file) {
+            return new HTTPResponse(null, 404);
+        }
+
+        if(!$file->canView()) {
+            return new HTTPResponse(null, 403);
+        }
+
+        $versions = Versioned::get_all_versions($class, $fileId)
+            ->limit(100)
```
chillu: Can we make this configurable on the `AssetAdmin` key? It's going to depend on user needs and stuff like available PHP memory.

code/Controller/AssetAdmin.php
```diff
@@ -387,6 +401,96 @@ public function apiCreateFile(HTTPRequest $request)
...
+            }
+
+            $prev = $v;
+        }
+
+        foreach($versions as $version) {
+            $author = null;
+
+            if($version->AuthorID) {
+                $author = DataObject::get('SilverStripe\\Security\\Member')
```
chillu: There's no caching on the ORM level here, so this will be expensive for large history lists - can we add some inline caching here?

tractorcow:
> I disagree, we shouldn't try to cram more structured data into form schema - it's going to make it harder to move this over to GraphQL. Separate requests after returning a form schema will be unavoidable, e.g. we can't inline all data required for a React-powered GridField into form schema.

Unnecessary requests are avoidable. In any case, I've already written a proof of concept for this. https://github.com/wilr/silverstripe-asset-admin/pull/1

In any case, graphql is a theoretical solution, and formschema is the existing one, which is my basis for preferring it in this case.

chillu:
I really don't think that all of our UI has to fit inside the concept of a "form field" just in order to expose it through a "form schema" response. It goes against an API-driven UI, and misuses the schema for more than it should be responsible for. What happens when you want to show the third most recent edits from the history in another view that's not a form? Even if we don't do GraphQL, I'd see this in a REST endpoint similar to what Will has built already.

wilr:
Thanks for feedback (npm lint is great 👍 ) I've implemented the minor feedback but think we're at 2 votes for/2 against for having a separate API history endpoint vs embedding the data within a form schema...

chillu:
From my understanding, its 3:1. Vote for separate API endpoint: Will, Ingo, Sam. Vote for embedded data in form schema: Damian.

Quoting Sam from our team chat:

> Sam: A 2nd API call is not a blocker.
> Sam: I don’t think it’s a design goal of react form-schema parsing to allow things like the history UI to be added/removed in responses to server-side code changes.
> Ingo: Will’s approach still creates a `LiteralField` with `component = "HistoryList"`, but that’s a necessary evil given how we construct react forms. It doesn’t introduce a separate PHP class for something that’s really a React view concern though
> Sam: That seems like an acceptable middle ground without further out-of-scope refactorings.
> Sam: Given that with GraphQL looming any API we build is a temporary API I’m not going to die in a ditch over which temporary API we choose.
> Sam: But `AssetAdmin::apiHistory()` seems like a smaller surface area for a temporary API compared to building a new field type

tractorcow:
Yeah I don't feel that strongly about it. Go ahead with the extra endpoint.

clarkepaul:
@wilr I've made some changes to the CSS but more importantly I've added some mock data in place of where I think it would good to have some more details. Hope you don't mind I've pushed them straight in to your branch.

wilr:
@clarkepaul Can you mark up where you need data from the API to populate?

clarkepaul:
Yup can do, tomorrow

code/Forms/FileFormFactory.php
```diff
@@ -192,8 +208,17 @@ protected function getSpecsMarkup($record)
         if (!$record || !$record->exists()) {
             return null;
         }
+        /**
+         * Can remove .label and .label-info when Bootstrap has been updated to BS4 Beta
+         * .label is being replaced with .tag
+         */
         return sprintf(
```
tractorcow: A small issue with the below, is that these aren't implemented in the subclassed ImageFormfactory::getSpecsMarkup. Is there a design for this?

code/Controller/AssetAdmin.php
```diff
@@ -829,14 +831,24 @@ public function FileInsertForm()
...
+        $versionId = $context['RecordVersion'];
+        if(!$id || !$versionId) {
+            return $this->httpError(404);
+        }
 
-        $request = $this->getRequest();
+        /** @var File $file */
+        $file = Versioned::get_version(File::class, $id, $versionId);
 
         if (!$file->canView()) {
```
flamerohr: should this have a `!$file` check as well?
`get_version()` could return null right?

code/Forms/FileFormFactory.php
clarkepaul: Latest versions don't need a number, as the are just current in my opinion, as far as I'm aware there is nowhere you need to select the current version as its always at the top.

tractorcow:
History view form now looks like the below:

image

tractorcow:
Approved by internal peeps, so merging. :)

tractorcow:
I'm not sure how but this PR zeroed all the images.

Issue #173, opened 2016-05-30
New-style Insert Media dialog


sminnee:
## Acceptance criteria

 - The insert media dialog matches the look and feel of the new asset-admin section
 - I can upload one or more files prior to inserting it
 - The last of mulitple uploaded files is auto-selected and shows in the sidebar ready to insert
 - The multi-select abilities are removed
 - We have at least 1 behat test

Exclusion

 - No multi-file insert feature. It never made any sense
 - Insert from URL (separate story)
 - Filter files by images (https://github.com/silverstripe/silverstripe-asset-admin/issues/121)

---

Insert media link triggers the files area to appear in a pop up modal. Refer to https://projects.invisionapp.com/boards/6GX68P9YKERH/ 

Ideally, this would be part of the asset-admin module, which would add the relevant toolbar item to the CMS. Otherwise, we introduce a dependency form HTMLEditorField to asset-admin.

# Tasks
- [x] Refactor `<AssetAdmin>`, `<Gallery>` and  `<Editor>` to work standalone (or load entire admin/assets into an iframe)
- [x] Get AssetAdmin to appear on the pages section using entwine
- [x] Create "Insert" form (using a different React component based on `<Editor>`, but with `HtmlEditorField` form field schema)
- [x] Edit details on Insert form (to be confirmed)
- [x] ~Determine Generate shortcode logic~ (not required for this story, as tinymce plugin handles shortcode)
- [x] Insert images
- [x] Checkboxes are not available for modal
- [x] Uploading files, the latest file is selected/opened in Editor. (NOTE: this only opens the last file that finished uploading, so it can be a little unpredictable...)
- [x] Docblocks
- [x] Unit tests
- [x] Behat tests

# Excludes
- Recently edited files

# Blockers
- Behat async behaviours
- Components using FormBuilderLoader cannot have unit tests due to jest mocking errors

marcello-silverstripe:
Chris and Paul to write stories for "Insert from URL" (OEmbed) and "Insert link", potentially design refresh with retaining current Entwine UI.

marcello-silverstripe:
Paul to determine if "Details" tab should show readonly, and how we deal with different semantics for "Title" fields (title on link of this particular instance vs. title on the file record itself)

clarkepaul:
Edit details design for placing files: There are a few ways we could have gone with this but I think if we limit what they can edit when placing the image the UI will be less confusing. 

[...]

PR #307, opened 2016-11-15
Insert media modal in react

flamerohr:
This adds the Insert media modal react component, so you can open the asset admin section in a Modal box and pick an image to "insert".
The modal is intended for use with the WYSIWYG (TinyMCE) in silverstripe framework.

For more details, see https://github.com/silverstripe/silverstripe-asset-admin/issues/173

To be merged before this: https://github.com/silverstripe/silverstripe-framework/pull/6317


tractorcow:
I've added enough basic scaffolding so that we can write behat tests.

clarkepaul:
Removed vh fix for height as IE10 doesn't read it and replaced with flexbox styles. Increased width and position of modal to match designs.

chillu:
Currently broken for me - no JS errors or failed HTTP requests. Tried an `npm run build` in framework and asset-admin without any luck

image

chillu:
Awesome job Chris, that looks like a lot of work! :) One niggle: When selecting an image in TinyMCE, the modal dialog button should read "Update" rather than "Insert"

chillu:
Also, the dimensions were keeping the aspect ratio before - it's a lot less useful if they don't.

flamerohr:
Hmm, I'm not sure how to handle Changing the Insert button to relabel as Update, would it be adding some POST data to the schema retrieval?

I believe @tractorcow is working on the algorithm for aspect ratio currently :)

tractorcow:
I've only updated the initial values to respect aspect ratio. I haven't implemented any on-update support for retaining aspect ratio. These should though.

client/src/containers/InsertMediaModal/InsertMediaModal.js
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * Compares the current properties with received properties and determines if overrides need to be
+   * cleared or added.
+   *
+   * @param {object} props
+   */
+  setOverrides(props) {
+    if (!props || this.props.schemaUrl !== props.schemaUrl) {
+      // clear any overrides that may be in place
+      this.props.actions.formData.destroySchemaStateOverrides(this.props.schemaUrl);
```
chillu: Can we make this `setSchemaStateOverrides(null)` instead? Not too keen to have different variations for each array modification action

flamerohr: I would prefer not...
Makes it clear what it's supposed to do, rather than implied with `null`
Many other libraries and modules also follow a destroy call for destroying

flamerohr:
Tests green now :D

tractorcow:
https://github.com/silverstripe/silverstripe-framework/pull/6317 is merged, so restarting tests.

clarkepaul:
Passing :)

Issue #297, opened 2016-11-01
API Enable react UploadField.js support within non-react sections


tractorcow:
In order to fully deprecate framework UploadField it will be necessary to build an uploadfield which works outside of react.

I propose creating a react wrapper around https://github.com/silverstripe/silverstripe-asset-admin/issues/296 which will bootstrap react / redux around UploadField.js, allowing it to be used in place of any uploadfield.

Acceptance Criterea:
 - The asset-admin UploadField can be used in php forms
 - This UploadField allows files to be added or removed from a has_one or many_many
 - Look and feel matches designs
 
Tasks:
 - [x] Fix js errors https://github.com/silverstripe/silverstripe-asset-admin/pull/329
 - [x] Merge framework PR https://github.com/silverstripe/silverstripe-framework/pull/6392
 - [ ] Merge asset-admin PR https://github.com/silverstripe/silverstripe-asset-admin/pull/331
 - [x] Get UploadField.js to render in a normal form
 - [x] Form submission works
 - [x] Upgrading notes / guide

chillu:
Once that's implemented we'll want to remove the original `UploadField`, since we'll have a mostly equivalent replacement. Sam, Damian and myself have decided that the new 4.x `AssetField` should be removed altogether without a replacement, since its use case isn't well defined yet (saving `DBField` straight onto a record). `AssetField` relies on `UploadField` JavaScript and CSS, and we're aiming to get rid of that complexity from core.

tractorcow:
By removed, we mean deprecated for now right?

chillu:
I would've thought removed (or maybe moved to a module), assuming we can provide 95% equivalent features in the new `FileField`. And there's no reason to deprecate `AssetField` because it's not even in a stable release yet.

chillu:
Turns out the old field hasn't been removed as part of this work, here's a follow up ticket: https://github.com/silverstripe/silverstripe-framework/issues/6481

PR #312, opened 2016-11-22
API Implement react FileField

### LINKED ISSUE #296: API Create proof of concept react UploadField.js
Asset-admin module currently uses https://github.com/silverstripe/silverstripe-asset-admin/tree/master/client/src/components/AssetDropzone, which I propose should be used as the base for a react powered upload field.

Designs for this field have already been developed at https://github.com/silverstripe/silverstripe-asset-admin/issues/191 (see https://projects.invisionapp.com/boards/VGX68PKC7PYB/)

Acceptance critera for this step:

 - A react form with an UploadField can be created
 - This UploadField allows files to be added or removed from a has_one or many_many
 - Look and feel matches designs
 
 PR:
  - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/312
  - [x] https://github.com/silverstripe/silverstripe-framework/pull/6325

tractorcow:
Requires https://github.com/silverstripe/silverstripe-framework/pull/6325

Closes #296

See https://github.com/silverstripe/silverstripe-asset-admin/issues/297 for follow-up issue about making this field work in existing CMS sections


client/src/components/FileFieldItem/FileFieldItem.scss
```diff
@@ -0,0 +1,158 @@
...
+  padding: 11px 12px;
+  height: 100%;
+  line-height: $file-item-thumb-width - 22;
+  margin: 0;
+  overflow: hidden;
+  box-sizing: border-box;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  position: relative;
+  display: block;
```
flamerohr: would it be worth adding a `min-width` for this? probably not, but mobiles will find it very cramped and almost no information shown.

client/src/lib/fileFactory.js
```diff
@@ -0,0 +1,39 @@
+import deepFreeze from 'deep-freeze-strict';
+
+function fileFactory() {
```
flamerohr: This could be `const fileDefault = deepFreeze({` and do not require function calls. The deepFreeze (should) prevent mutations, so the benefit of generating new each time is lost.

Also, a fileFactory definition still lingers in `FileFieldReducer` ;) change to import

client/src/state/fileField/FileFieldReducer.js
```diff
@@ -0,0 +1,125 @@
+import deepFreeze from 'deep-freeze-strict';
+import ACTION_TYPES from './FileFieldActionTypes';
+
+export function fileFactory() {
```
flamerohr: with the `import fileFactory from 'lib/fileFactory';` statement at the top, you can put `export { fileFactory };` at the bottom, but since it's already importing, maybe just import from `'lib/fileFactory'` anyway? (pending getting renamed to `fileDefault`)

client/src/state/fileField/tests/FileFieldReducer-test.js
```diff
@@ -0,0 +1,150 @@
...
+  describe('FILEFIELD_ADD_FILE', () => {
+    it('should add a new file to a list', () => {
+      const nextState = fileFieldReducer(initialState, {
+        type: ACTION_TYPES.FILEFIELD_ADD_FILE,
+        payload: {
+          fieldId: 'myfield',
+          file: Object.assign({}, fileFactory(), { id: 7, name: 'NewFile.jpg' }),
+        },
+      });
+      expect(nextState.fields.myfield.length).toEqual(3);
```
flamerohr: Everywhere else uses `toBe` for value compares, could we keep to that as a standard?
`toEqual` does have the advantage of checking object properties, which I think is acceptable to use.

client/src/components/AssetDropzone/AssetDropzone.js
```diff
@@ -258,6 +267,20 @@ class AssetDropzone extends SilverStripeComponent {
   }
 
   /**
+   * Generate approximate guid.
+   * Credit to http://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript#2117523
+   *
+   * @returns {String}
+   */
+  generateQueuedId() {
```
tractorcow: How can I make sure it's globally unique? Is there a static property I can use on javascript classes?

client/src/state/fileField/tests/FileFieldReducer-test.js
tractorcow: Yes I think so.

client/src/components/AssetDropzone/AssetDropzone.js
flamerohr: It can be unique for that session, so having a counter "global" for that class (or possibly even Reducer instead of class file) which is shared across all instances of that class:
* a `let counter = 0;` at the top of the `AssetDropzone.js` file outside the class declaration.
* increment the counter for each new upload by that user: `file._queuedId = ++counter;`

Or even use Redux store to keep a counter would be another way.

I think UUID would be a reasonable solution for if the uploading status was shared across all sessions somehow.

tractorcow:
I've rebased and put all changes based on feedback in a separate commit, @flamerohr

client/src/components/FileFieldItem/README.md
```diff
@@ -0,0 +1,24 @@
...
+* `item` (object): File object this record contains. The shape of this object is equivalent to that
+  returned by `AssetAdmin::getObjectFromData` with the following extra considerations:
+   - `item.uploaded` (boolean): A flag which is set to true if this file was recently added by the user
+   - `item.queuedId` (string): A GUID assigned to any in-progress uploads which are yet to be saved to the database.
+* `handleRemove` (function): A callback which will be invoked in case the 'remove' button is clicked. This
+  callback has these arguments:
+    - `event` (object): The event
+    - `item` (object): The file record for the FileFieldItem instance being removed
+
+ _NOTE:_ For other properties, please refer to the [react-bootstrap FormControl](https://react-bootstrap.github.io/components.html#forms-props-form-control) documentation.
```
flamerohr: Sorry @tractorcow , I should've noted this one for removal too in my previous peer review

code/Forms/FileField.php
```diff
@@ -0,0 +1,140 @@
...
+    protected $multiUpload = null;
+
+	/**
+	 * Create a new file field.
+	 *
+	 * @param string $name The internal field name, passed to forms.
+	 * @param string $title The field label.
+	 * @param SS_List $items Items assigned to this field
+	 */
+	public function __construct($name, $title = null, SS_List $items = null) {
```
chillu: Indentation

code/Forms/FileField.php
```diff
@@ -0,0 +1,140 @@
...
+        $defaults['data']['parentid'] = $this->getFolderID();
+        return $defaults;
+    }
+
+    /**
+     * Get ID of target parent folder
+     *
+     * @return int
+     */
+    protected function getFolderID() {
```
chillu: Coding standards

code/Forms/FileField.php
```diff
@@ -0,0 +1,140 @@
...
+ */
+class FileField extends FormField
+{
+    use FileUploadable;
+
+    /**
+     * @config
+     * @var int
+     */
+    private static $thumbnail_width = 60;
```
chillu: Shouldn't we start writing new statics as camelCase according to PSR-2?

client/src/state/fileField/FileFieldActions.js
```diff
@@ -0,0 +1,97 @@
...
+ * Updates a queued file when it successfully uploads.
+ *
+ * @param {String} fieldId - Identifier of FileField
+ * @param {String} queuedId - Temporary id assigned when this file was queued
+ * @param {Object} json - json encoded data for object sent from server
+ */
+export function succeedUpload(fieldId, queuedId, json) {
+  return (dispatch) =>
+    dispatch({
+      type: ACTION_TYPES.FILEFIELD_SUCCEED_UPLOAD,
```
chillu: We've used `<action>_SUCCESS` vs `<action>_FAILURE` elsewhere (e.g. `PUBLISH_CAMPAIGN_REQUEST`), so think we should stick with this naming convention

client/src/components/FileField/FileField.js
```diff
@@ -0,0 +1,254 @@
...
+        handleSending={this.handleSending}
+        handleUploadProgress={this.handleUploadProgress}
+        preview={dimensions}
+        options={dropzoneOptions}
+        securityID={securityID}
+        className="file-field__dropzone"
+      >
+        <div className="file-field__dropzone-area">
+          <span className="file-field__droptext">
+            <a className="file-field__upload-button">Browse</a>
```
chillu: Needs `_t()`

client/src/components/FileField/FileField.scss
```diff
@@ -0,0 +1,50 @@
...
+  top: 0;
+  right: auto;
+  bottom: auto;
+  margin: 0 0 10px 0;
+  border-width: 2px;
+  border-style: dashed;
+  border-color: $border-color-dark;
+  border-radius: 5px;
+  padding: 0;
+  background: #fafbfc;
```
chillu: Shouldn't we use colour placeholders for this?

client/src/components/FileFieldItem/FileFieldItem.js
```diff
@@ -0,0 +1,286 @@
+import i18n from 'i18n';
+import React from 'react';
+import SilverStripeComponent from 'lib/SilverStripeComponent';
+import CONSTANTS from 'constants/index';
+import fileShape from 'lib/fileShape';
+
+class FileFieldItem extends SilverStripeComponent {
```
chillu: If this component isn't designed to be standalone, it should be in `components/FileField/FileFieldItem.js` (alongside `FileField.js`).

client/src/components/FileFieldItem/FileFieldItem.js
```diff
@@ -0,0 +1,286 @@
...
+
+    return null;
+  }
+
+  /**
+   * Gets the remove item button
+   *
+   * @returns {XML}
+   */
+  getRemoveButton() {
```
chillu: Methods returning virtual DOM should always use `render*()` (rather than `get*()`)

chillu:
Awesome work, Damian - you're officially a React expert now! :) I'm still trying to wrap my head around the general direction here: Are you proposing that we make an AssetAdmin specific `FileField` (which uses `admin/assets/api/createFile`), which co-exists with the current `UploadField`? I thought the overarching goal here was to visually overhaul `UploadField`, and while we're at it rewrite it in React? If we're keeping `UploadField` as-is, that's just increasing the long-term maintenance burden. Also, where is `FileField` actually used by default? I can't see an instantiation of it.

code/Forms/FileField.php
tractorcow: We aren't doing psr-2 yet in asset-admin.

tractorcow:
> I thought the overarching goal here was to visually overhaul UploadField, and while we're at it rewrite it in React?

The issue here is that I couldn't do it in a way that was independent of asset-admin, not without copying a lot of code from asset-admin to framework. The other issue is that UploadField.php simply does not work in react due to nested routed handlers not being supported. (e.g. allowed_actions on formfields). It's a problem we won't solve, as we are moving away from forms-as-controllers and to a graphql based solution, so it would be wasteful to try to fix.

tractorcow:
I'll fix your feedback and update soon. :)

tractorcow:
Updated @chillu

code/Forms/FileField.php
chillu: Heh true, I keep forgetting that it only applies to method names (http://www.php-fig.org/psr/psr-1/#4-3-methods)

chillu:
Two issues when testing this in `AssetAdmin`:

- A `has_one` causes "upload cancelled"
- A field bound to a `has_one` still allows you to select multiple files, but will only save the last one uploaded

tractorcow:
FIxed has_one issues and renamed from FileField to UploadField.

tractorcow:
Just waiting for psr2 linting fixes to pass CI :)

Issue #313, opened 2016-11-22
Draft/modified symbol is below missing file notice


clarkepaul:
A z-index: 1 needs to be added to the symbol so it sits above the red area.
![pasted_image_23_11_16__12_31_pm](https://cloud.githubusercontent.com/assets/555033/20546304/1afa68e2-b179-11e6-9c85-3b346fa0510e.png)

PR #314, opened 2016-11-22
State is hidden under missing file notice

clarkepaul:
Move state indicator above red overlay.
Also added fix for missing hover state on missing files.

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/313


clarkepaul:
Rebased, will wait for tests to merge

Issue #278, opened 2016-09-28
List view for files


sminnee:
## Acceptance criteria

- Designs are finalised
- I can switch between gallery view and list view on any folder of the asset admin
- I can see [tbc] columns in the list view
- I can click a column header to sort by that column
- Pagination works (either infinite scroll or actual page list)
- list vs gallery view indicated in the URL, as does sorting and current page. querystring is fine
- Shifting between gallery view and list view doesn't unnecessarily cause a refresh from server (i.e. it's using the same piece of redux state). Partial refresh due to the showing different pages of data is allowed.

Excluded

 - Column selector control for narrower width devices (separate story)
 - Multi-select for batch actions (separate story)

https://projects.invisionapp.com/boards/9GX68PTXUQFH#/1547737/129314959

# Dependency
https://github.com/silverstripe/silverstripe-asset-admin/issues/173 - because it has a lot of the url param and routing rework required for functionality such as sorting.

# Tasks
- [x] Add view switch function/component
- [x] Separate tile view to a component
- [x] Create list view component
- [x] Implement column sorting
- [x] Pagination (with page links/numbers)
- [x] Docblocks
- [x] Unit tests

# Front-end tasks
- [x] Thumbnail size and list size have same height
- [x] Toggle button size increase (not blurry)
- [x] Checkbox and thumb column at 1%
- [x] Pagination to match rest of CMS pagination
- [x] Remove additional ascending/descending character ▲
- [x] Browser testing

Opening as separate issues:
- Turn sort by into icon 
- Adjust small folder and doc icons so not blurry
- While loading table the editor panel floats in centre of screen (not fixed right)

# Pull requests
- https://github.com/silverstripe/silverstripe-asset-admin/pull/327
- https://github.com/silverstripe/silverstripe-framework/pull/6371

clarkepaul:
Visual direction for list view of items:
https://projects.invisionapp.com/boards/2JX68PNPDF6K/

The toggle for List vs Thumbnail view is to be simplified to a show an icon for the opposite view. e.g. if the user is viewing the thumbnail view then show the list view icon and vice-versa.
https://projects.invisionapp.com/boards/JKX68PN94BSP/

flamerohr:
I had a quick look around for modules and compared [Griddle](http://griddlegriddle.github.io/Griddle/index.html) vs [Facebook's fixed-data-table](https://facebook.github.io/fixed-data-table/getting-started.html)

[...]

PR #327, opened 2016-12-08
List view for gallery

flamerohr:
This adds a list view (GridField type table view) for gallery items, to contrast with the default thumbnail view that is currently in place.

See https://github.com/silverstripe/silverstripe-asset-admin/issues/278 for more details

requires https://github.com/silverstripe/silverstripe-framework/pull/6371 to be merged first


tractorcow:
Better than a listing gallley

tractorcow:
There's a missing progress indicator when uploading files in gallery view.

Files also temporarily display `NaN GB` as files size.

tractorcow:
When toggling between two states, shouldn't we show these as side-by-side tabs, rather than just the current icon?

image

Something similar to 

image

tractorcow:
I've added a fix for griddle re-sorting values. I'll let you fix the upload progress indicator. :)

flamerohr:
I've added the progress bar and fixed the NaN issue

tractorcow:
Merged framework, restarted tests.

flamerohr:
thanks for fixing the behat tests @tractorcow, looking good :)

Issue #121, opened 2016-04-26
Assets search


chillu:
## Acceptance criteria:

- You can search for files by a keyword
- You can search by file type
- You can search by last updated date, as a date range (datepicker control needed)
- You can search in either the current folder and its subfolders, or just current folder (depending on pattern created and tested is most intuitive) and across the whole asset library

---
UX CSS:
Search wireframes: https://projects.invisionapp.com/boards/FAX68PE52Y8W/
* [x] Mock templates (using HTML from form templates)
  * [x] Expanding search field
  * [x] Search field and button
  * [x] Expanding filter panel 
  * [x] File type dropdown
  * [x] Limit to current folder checkbox
  * [x] Date range, fields (not datepicker)
  * [x] Search button / cancel button
  * [x] BEM styles
  * [x] Design pattern for clearing results and know you are viewing filtered results
* [x] Browser testing

DEV:
* [x] Replace temp markup with form schema
  * [x] Filter by type (image, audio, video (includes flash), archive, document, folder [nice to have])
  * [x] Date fields
  * [x] Limit to current folder checkbox
* [x] On search load, search field (not filter panel) must remain open and both search field and filters remain populated
* [x] When selecting search the search input is in focus
* [x] Search results with message to clear
* [x] Breadcrumbs represent search
* [x] Ensure hitting "enter" on any field in the form triggers a search.
* [x] Ensure all aria-expanded="true" are set to false when closed and vice versa 
* [x] Remove temporary JS
* [x] Clear form when clearing results

# Pull requests
* [ ] https://github.com/silverstripe/silverstripe-framework/pull/6389
* [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/330

tractorcow:
We should remove the `admin/assets` => `admin/assets/show/0` redirect, and have `admin/assets` as the canonical URL for showing assets in the root directory. It also breaks the "back" button when going back through this url.

sminnee:
As long as we're aware that this means we'll need to redirect `admin/assets/show/0` => `admin/assets`.

sminnee:
We can probably live without this for alpha1.

chillu:
Paul has removed search icon, fixing can wait until alpha2

clarkepaul:
I done the styling for the search in https://github.com/silverstripe/silverstripe-asset-admin/issues/200 , from discussions we will probably go with just the search input and hold off on the advanced filters for the time being.

[...]

PR #330, opened 2016-12-13
Assets search feature

clarkepaul:
Requires framework branch
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/121

Close https://github.com/silverstripe/silverstripe-asset-admin/issues/345 alongside this PR


flamerohr:
Work has been completed and this is ready for peer review, travis won't pass until the framework portion is merged first

tractorcow:
I'm in the process of rebasing this, but the rebased version requires a bit more work.


https://github.com/open-sausages/silverstripe-asset-admin/tree/pulls/4.0/asset-admin-search-rebased

It needs to be re-written to work with graphql.

tractorcow:
I think someone with more graphql knowledge needs to do this for me. It's gone above my head.

tractorcow:
@chillu I'm in early today to get my head around graphql properly. :) That way next week I'll be fresh with all the correct knowledge needed to re-try this rebase.

tractorcow:
👍  on your refactors and suggestions. I'll look at them during the rebase.

chillu:
Also fixes [Pagination stays in place between folders](https://github.com/silverstripe/silverstripe-asset-admin/issues/345), please close once this is merged.

tractorcow:
Pagination seems to have issues:

 - First page is `&page=0`, when it should match the same value in the dropdown. `1`. it makes for an unintuitive UX.
 - pagination querystring arguments are present when not needed. You don't need `page=1` / `page=0` when viewing the first page; maybe we just add it on the second page and beyond?
 - The "next page" button (`>`) doesn't work.

client/src/containers/AssetAdmin/AssetAdminRouter.js
```diff
@@ -18,36 +45,23 @@ class AssetAdminRouter extends Component {
   /**
    * Generates the Url for a given folder and file ID.
    *
-   * @param {number} [folderId]
```
chillu: There's no `Integer` type in JavaScript, it's all `Number` - see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects. Since it's a type, it should start with an uppercase letter though

`typeof parseInt(1) // "number"` and `typeof parseFloat(1.1) // "number"`

chillu:
Rebased and squashed a bit. Merging with one broken Behat test, since that test is already broken in "master" at the moment (https://travis-ci.org/silverstripe/silverstripe-asset-admin/jobs/190123962).

PR #331, opened 2016-12-13
ENHANCEMENT Enable UploadField.js within non-react sections	

### LINKED ISSUE #297: API Enable react UploadField.js support within non-react sections

tractorcow:
Relies on fixes in https://github.com/silverstripe/silverstripe-asset-admin/pull/329, so rebased on top of this commit.

Requires https://github.com/silverstripe/silverstripe-framework/pull/6392 to be merged first.

Fixes #297

This is what the field looks like in cms main:

image


client/src/components/UploadField/UploadField.js
```diff
@@ -140,6 +140,7 @@ class UploadField extends SilverStripeComponent {
    */
   handleSelect(event) {
     event.preventDefault();
+    return false;
```
dhensby: My understanding is that `return false` is bad practice.

If you want to stop propagation of the event, then use `event.stopPropagation()` and not `return false`.

client/src/entwine/UploadField/UploadFieldShiv.js
```diff
@@ -0,0 +1,36 @@
...
+   * Get props to use for this object
+   *
+   * @returns {*}
+   */
+  getUploadProps() {
+    return Object.assign(
+      {},
+      this.props.schema,
+      this.props.state,
+      { data: Object.assign({}, this.props.schema.data || {}, this.props.state.data || {}) }
```
flamerohr: This should use `merge.recursive(true, {}, data, data);` so you don't have different functionality to `FormBuilder.mergeFieldData()`

Actually, I prefer if this prop could be done in the entwine `getAttributes` and take advantage of `$.extends(true, {}, data, data);` and you won't need the shiv class layer.

tractorcow:
Feedback implemented

dhensby:
leaving to @flamerohr to make sure he's happy with resolutions

flamerohr:
Looks good, I had amended a: `export default fieldHolder(connect(mapStateToProps, mapDispatchToProps)(UploadField));` to 
`export default fieldHolder(ConnectedUploadField);` to reduce code duplication which I hadn't noticed in my previous review :)

client/src/entwine/UploadField/UploadFieldEntwine.js
```diff
@@ -0,0 +1,48 @@
...
+ * Shiv for inserting react UploadField into entwine forms
+ */
+jQuery.entwine('ss', ($) => {
+
+  /**
+   * See boot/index.js for `.react-boot` bootstrap
+   */
+  $('.react-boot input.entwine-uploadfield').entwine({
+
+    onunmatch() {
```
chillu: Did you omit `this._super();` in `onmatch()` and `onunmatch()` on purpose here?

client/src/entwine/UploadField/UploadFieldEntwine.js
```diff
@@ -0,0 +1,48 @@
...
+
+    onmatch() {
+      this.refresh();
+    },
+
+    refresh() {
+      const store = window.ss.store;
+      const props = this.getAttributes();
+      ReactDOM.render(
+        <Provider store={store}>
```
chillu: At first I thought that multiple `<Provider>` calls in with the same store would be bad, but it [looks like](https://github.com/reactjs/react-redux/blob/master/src/components/Provider.js) it's OK - just a shortcut for react context, the meat of the implementation is in the individual `connect()` calls

client/src/entwine/UploadField/UploadFieldEntwine.js
tractorcow: No, I legit forgot them. I'll restore them.

PR #336, opened 2016-12-16
Honor the canDelete flag in bulk delete

mikenz:
Can only bulk delete the selected files if canDelete is true for all selected files


chillu:
Thanks Mike! I've run an `npm run build` in the folder to ensure the dist files get generated - we're committing them to the repo to avoid a NodeJS requirement for SilverStripe devs at the moment (yes, it's painful)

Issue #295, opened 2016-10-26
File replacement action


sminnee:
As a user, I want to be able to replace a file, so that I can have a new file downloaded from an existing URL.
## Acceptance criteria
- On the file detail view, there is a replace action, that allows the uploading of a single file
- If the replace action is used, the resulting URL is the same as the previous file, and the old file is tracked in the version history
- If you upload a file with a different extension, it warns that the URL will change
- If you upload a file with an invalid extension, it will throw an error

Exclusion

- When bulk uploading files, if there is a file of the same name already in the folder, an option to replace exist (separate story)

## Tasks
 - [x] Create a preview image component, which handles rendering the preview and icon-bar to the right.
 - [x] Update FormFactory to use the preview image component. This allows the submission of new file tuple (Filename, Hash, Variant) to replace the contained DBFile.
 - [x] Show progress of upload for replacement
 - [x] Add endpoint for uploading the replacement image.
 - [x] Ensure that component is updatable via state (from form submission returning updated formschema state).
 - [x] Check for different file extension and display a warning
 - [x] Unit tests
 - [x] Behat tests

# Pull requests
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/339
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/6428

flamerohr:
User flow for replacing file, I think, goes something like this:

1. User opens the Editor for a file.
2. User clicks "Replace" call-to-action button.
3. Modal opens with Upload Field for user to provide replacement file for.
4. User will get a confirmation with warnings.
5. User clicks "Ok" for the confirmation and upload happens replacing file, retaining filename where necessary.

clarkepaul:
This was wireframe produced by @scott1702 way back but it doesn't really suit.
https://invis.io/C596UOLDZ#/117448956_SilverStripe_UI_Copy_PS_Edit_File

As you described @flamerohr but we just need a browse window and once you chose a new file it would upload in place of the old image with progress bar. 
The image would only be updated once "Save" is selected I'm thinking but not sure on that one?

clarkepaul:
Something I'm thinking about... Once you select the replace action would it be weird to bring up the files area in the modal (same as the insert into pages modal) so the user can replace with an existing file or upload a new one.

flamerohr:
[...]

PR #339, opened 2016-12-20
File replacement action

flamerohr:
This enables the user to upload a new file to replace the currently opened file in the asset-admin area.

Only supports uploading one file as a replacement.

Please merge https://github.com/silverstripe/silverstripe-framework/pull/6428 first.
Please see https://github.com/silverstripe/silverstripe-asset-admin/issues/295 for more details


code/Controller/AssetAdmin.php
```diff
@@ -447,7 +455,74 @@ public function apiCreateFile(HTTPRequest $request)
...
     }
-
+    
+    /**
+     * Creates a single file based on a form-urlencoded upload.
+     *
+     * @param HTTPRequest $request
+     * @return HTTPRequest|HTTPResponse
+     */
+    public function apiUploadFile(HTTPRequest $request)
```
tractorcow: Too much duplication between this and apiCreateFile, with lack of docs distinguishing the different use cases.

Perhaps it should be apiReplaceFile to make the purpose clear?

tractorcow: Pushed a change to just document it better.

client/src/components/AssetDropzone/AssetDropzone.js
```diff
@@ -326,9 +355,9 @@ class AssetDropzone extends SilverStripeComponent {
         category: this.getFileCategory(file.type),
         filename: file.name,
         queuedId: file._queuedId,
-        size: fileSize(file.size),
+        size: file.size,
```
tractorcow: Why did you remove this? We lose all formatting. e.g. a `1.8 MB` file will appear as `1800000` now.

image

flamerohr: Removed because it was causing double formatting.
The reverse is that you get "NaN" when uploading in the table view.

tractorcow: I've fixed this by adding fileSize to UploadFieldItem.js

flamerohr:
Thanks for the fixes @tractorcow, they all look good!

Issue #361, opened 2017-01-11
UploadField "add from files" is broken


chillu:
![image](https://cloud.githubusercontent.com/assets/111025/21845426/805dc5c6-d857-11e6-8294-93d91339ee46.png)

Partial fix by adding `<ApolloProvider>`, but it still fails with other errors:

```diff
diff --git a/client/src/components/UploadField/UploadField.js b/client/src/components/UploadField/UploadField.js
index cba65b2..c8735ce 100644
--- a/client/src/components/UploadField/UploadField.js
+++ b/client/src/components/UploadField/UploadField.js
@@ -2,6 +2,7 @@ import i18n from 'i18n';
 import React from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
+import { ApolloProvider } from 'react-apollo';
 import CONSTANTS from 'constants/index';
 import SilverStripeComponent from 'lib/SilverStripeComponent';
 import fieldHolder from 'components/FieldHolder/FieldHolder';
@@ -248,16 +249,20 @@ class UploadField extends SilverStripeComponent {
   }
 
   renderDialog() {
+    const store = window.ss.store;
+    const client = window.ss.apolloClient;
     return (
-      <InsertMediaModal
-        title={false}
-        show={this.state.selecting}
-        onInsert={this.handleAddInsert}
-        onHide={this.handleAddHide}
-        bodyClassName="modal__dialog"
-        className="insert-media-react__dialog-wrapper"
-        type="select"
-      />
+      <ApolloProvider store={store} client={client}>
+        <InsertMediaModal
+          title={false}
+          show={this.state.selecting}
+          onInsert={this.handleAddInsert}
+          onHide={this.handleAddHide}
+          bodyClassName="modal__dialog"
+          className="insert-media-react__dialog-wrapper"
+          type="select"
+        />
+      </ApolloProvider>
     );
   }
 
```

Assigning this to alpha4 since we surely want people to try out the shiny new uploadfield implementation, which won't happen if it comes broken out of the box :)

flamerohr:
Pull request created https://github.com/silverstripe/silverstripe-asset-admin/pull/364

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-asset-admin/pull/364

PR #364, opened 2017-01-15
Fix Add from files in Upload fields in an entwine context

flamerohr:
Fixes upload fields that have an entwine context

Please see https://github.com/silverstripe/silverstripe-asset-admin/issues/361 for more details


tractorcow:
Test failures are pre-existing and will be fixed with https://github.com/silverstripe/silverstripe-graphql/pull/56

PR #370, opened 2017-01-17
BUG Ensure that file replacement checks for clashes with existing files

### LINKED ISSUE #369: UploadFile api does not check for conflicts
Paraphrasing @tractorcow 
This is an issue where saving a file in asset-admin wasn't safely ensuring the name was unique
The asset store does a file-exists uniqueness check, but there's a case where a File dataobject with a missing physical file would not trigger a conflict.

So we need to enforce uniqueness against:
- Filename against all File dataobjects
- Filename and Hash against the actual file store

tractorcow:
Fixes #369

Requires https://github.com/silverstripe/silverstripe-framework/pull/6527

Note: Updated so that existing files are detected to avoid duplication, as per discussion with @sminnee

### DIFF
--- a/client/src/components/PreviewImageField/PreviewImageField.js
+++ b/client/src/components/PreviewImageField/PreviewImageField.js
@@ -4,6 +4,7 @@ import AssetDropzone from 'components/AssetDropzone/AssetDropzone';
 import CONSTANTS from 'constants';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
+import { formValueSelector } from 'redux-form';
 import * as previewFieldActions from 'state/previewField/PreviewFieldActions';
 import { getFileExtension } from 'lib/DataFormat';
 
@@ -19,6 +20,7 @@ class PreviewImageField extends Component {
     this.handleCancelUpload = this.handleCancelUpload.bind(this);
     this.handleRemoveErroredUpload = this.handleRemoveErroredUpload.bind(this);
     this.canFileUpload = this.canFileUpload.bind(this);
+    this.updateFormData = this.updateFormData.bind(this);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -73,6 +75,7 @@ class PreviewImageField extends Component {
       handleSending: this.handleSending,
       handleUploadProgress: this.handleUploadProgress,
       canFileUpload: this.canFileUpload,
+      updateFormData: this.updateFormData,
     };
   }
 
@@ -166,10 +169,9 @@ class PreviewImageField extends Component {
       this.props.onAutofill('FileHash', json.Hash);
       this.props.onAutofill('FileVariant', json.Variant);
 
-      const oldExt = getFileExtension(this.props.data.url);
-      const newExt = getFileExtension(json.Filename);
-      if (oldExt !== newExt) {
-        // name field to autofill if extension has changed
+      // Note: This Name was posted back from the current form field value,
+      // and may have been modified on the server. If so, update the form value
+      if (json.Name) {
         this.props.onAutofill(this.props.data.nameField, json.Name);
       }
     }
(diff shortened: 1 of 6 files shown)


client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -166,10 +168,8 @@ class PreviewImageField extends Component {
       this.props.onAutofill('FileHash', json.Hash);
       this.props.onAutofill('FileVariant', json.Variant);
 
-      const oldExt = getFileExtension(this.props.data.url);
-      const newExt = getFileExtension(json.Filename);
-      if (oldExt !== newExt) {
-        // name field to autofill if extension has changed
+      // If server has requested a filename change, modify it here
+      if (json.Name) {
```
flamerohr: This can have the undesired effect of replacing the Name value an edited Name field with the original value? It's not the best user experience.

code/Controller/AssetAdmin.php
```diff
@@ -250,28 +253,57 @@ public function apiCreateFile(HTTPRequest $request)
...
+        /** @var File $file */
+        $file = File::get()->byID($data['ID']);
+        if (!$file) {
+            return new HTTPResponse('File not found', 404);
+        }
+        $folder = $file->ParentID ? $file->Parent()->getFilename() : '/';
+
+        // If extension is the same, attempt to re-use existing name
+        if (File::get_file_extension($tmpFile['name']) === $file->getExtension()) {
+            $tmpFile['name'] = $file->Name;
```
flamerohr: rather than using the Name saved in the database, I wonder if the component could send the API the current value of the field on the form?

tractorcow:
Updated based on feedback.

Note that framework PR https://github.com/silverstripe/silverstripe-framework/pull/6527 is now required for this fix.

client/src/components/PreviewImageField/PreviewImageField.js
tractorcow: I've changed it so that the current Name is always posted back, so that json.Name will reflect the current value, unless explicitly changed.

code/Controller/AssetAdmin.php
tractorcow: Done.

tests/behat/features/replace-file.feature
```diff
@@ -14,24 +15,38 @@ Feature: Replace a file with a new file
...
+  Scenario: I upload a new file into the file I have open
     When I attach the file "testfile.jpg" to dropzone "PreviewImage"
       And I wait for 1 second
     Then I should see a ".preview__toolbar-button--remove" element
       And I should see a ".preview__message--success" element
-      And I press the "Save" button
-      And I wait for 1 second
+      And the "Name" field should contain "file1.jpg"
+    When I press the "Save" button
+      And I wait for 2 seconds
```
chillu: Not too happy about manual waits here, but it's just following a bad precedent. We need to work out if a `fetch()` is in progress through Behat feature context

tractorcow: It's annoying I know, but we need a better wait-until-ready solution for behat.

tractorcow: It's annoying I know, but we need a better wait-until-ready solution for behat.

tractorcow:
I can't reproduce the composer issue locally; Downgrading to 1.3.1 to see if it's due to the recent composer release.

tractorcow:
I've raised https://github.com/composer/composer/issues/6115 to investigate the possible regression in composer 1.3.2.

I suspect that the issue might be due to asset-admin dependency in the PR branch and the one cached by packagist differing.

I've raised https://github.com/silverstripe/silverstripe-asset-admin/pull/378, and suspect that force-merging this (with a broken build) and then updating packagist may resolve the issue. It can't help to experiment. :)

tractorcow:
I've force-merged https://github.com/silverstripe/silverstripe-asset-admin/pull/378 and updated packagist to determine if the issue lies between travis and packagist (which could explain why I couldn't resolve the issue locally).

tractorcow:
For some reason that issue resolved itself, yet without leaving behind any explanation as to why it was broken. :)

tractorcow:
Only failure now is the behat failure, i.e. a real one. However this already exists prior to this PR. (siiigh).

Issue #371, opened 2017-01-18
HistoryList is exported twice


zacps:
In `client/src/containers/HistoryList/HistoryList.js`

chillu:
Thanks, I think this has been fixed since :)

PR #374, opened 2017-01-18
Add missing dependencies dropzone & griddle-react (fixes #372)

### LINKED ISSUE #372: Add missing npm modules for 'run build' and 'shrinkwrap'
`npm run build` requires `dropzone` and `griddle-react` neither of which are specified in `package.json`.
`npm shrinkwrap` requires `react-addons-css-transition-group`

zacps:
Fixes #372.
There were two missing dependencies for `npm run build`, `dropzone` and `griddle-react`.


zacps:
The build will fail without #371.

zacps:
After some more playing around with it I'm not sure this is the problem. There seem to be some modules that don't get pulled from `npm install` almost every time I run it, but I haven't been able to figure out the cause of it.

flamerohr:
dropzone and griddle-react are in the `dependencies` already, what you've done is add them to the `devDependencies` :)

I'm not sure what happened with your dependency installation, but it looks like a lot of packages were removed from npm-shrinkwrap...
Closing, as the dependencies are not missing

Issue #279, opened 2016-09-28
Drag & drop movement of files


sminnee:
## Acceptance criteria:

 - I can drag a file from the gallery into a folder which is in the same view 
 - I can drag a file into a folder from my desktop directly into a folder
 - If I have multi-select enabled, then dragging any of the selected files will move all selected files into the destination folder
 - A message is present with an indication of how many items have been moved if successful
 - If multiple items are moved and some cannot be moved (ie no permissions), then we move the items that can be moved and alert the user to which items could not be moved.

## Exclusions:

 - "Rectangle drag" based multi-select (nice-to-have / optional)
 - No drag & drop in mobile unless built into functionality as part of library

clarkepaul:
See https://projects.invisionapp.com/boards/C3X68PKQ6M7Z/ for initial UX concept for dragging a single file into a folder and bulk selection of files (this would allow items to be dragged as a group and enable the bulk actions).

Multi-select would also be enabled for adding and removing items if multiple items are selected using "command + click" (mac) and respective shortcut for windows.

flamerohr:
Griddle would either need a lot of hacking/tweaking to work, or a full replacement...
its behaviour is so weird... default wrapper is `<table>` with headers, but if you want to customise the row component, it goes all out and replaces the wrapper with `<div>` and no headers... Meaning, I won't be able to apply the necessary HOCs to the row/item without replicating all the Griddle-internal render logic.

sminnee:
What if we just started with the gallery view and left the list view until later?

flamerohr:
Sure, I've almost got all the kinks out for the Gallery view with drag and drop, it's working really nicely.
I suggest pulling the list view to a separate story for a more detailed investigation?

clarkepaul:
![pasted_image_16_02_17__4_43_pm](https://cloud.githubusercontent.com/assets/555033/23006505/1f797f4e-f467-11e6-9587-18dbd2928831.png)

@flamerohr hopefully these aren't too late otherwise I can do some more styling during beta. From memory I have some styles for depth level (variations of shadow depths) and thats what .z-depth-x is.

clarkepaul:
Also I think two layers communicates enough to say there is more than one item being dragged, the direction of the stacking is different from what you had though.

flamerohr:
thanks @clarkepaul, not late at all :)
[...]

PR #385, opened 2017-02-20
Added drag and drop capability for files to folders

flamerohr:
This applies for the gallery view only.

Enables files and folders to be dragged into other folders, or the back button to move it out of the current folder.

Please see https://github.com/silverstripe/silverstripe-asset-admin/issues/279 for more details
Requires https://github.com/silverstripe/silverstripe-framework/pull/6634


tractorcow:
If I upload two files, and drag-drop them into a subfolder, I see the notification that +2 were moved, but the files remain in view.

image


Also can't create folders with a `this.props.mutateCreateFolder is not a function`

flamerohr:
I had shifted the mutations to actions, so they're more manageable for later on.

I'll be looking at moving those graphql queries to HOCs or imported actions, they're quite bloating when placed with the Components/Containers.

tractorcow:
Fixed the PHPCS issue; waiting for tests to finish.

PR #386, opened 2017-02-20
Fix editor panel is too large for smaller screens

flamerohr:
And fixed "Gallery" area when it is loading or errors. (Editor no longer floats in the centre)

I've changed from `100vh` to `100%` because the modal may not necessarily have 100% view height.

Fixes https://github.com/silverstripe/silverstripe-cms/issues/1706


tractorcow:
Rebased; Let's see if tests pass.

PR #397, opened 2017-03-03
Add Code Climate config

larkinscott:
Code Climate config that uses your existing ESLint config and your custom PHP ruleset.


tractorcow:
Thanks @larkinscott for setting this up. :) I'm still trying to figure out how to setup github integration with codeclimate.

PR #398, opened 2017-03-06
Upgrade to Node6 and yarn

flamerohr:
Switch to node6 with yarn, as per framework changes https://github.com/silverstripe/silverstripe-framework/pull/6679

Jest needed to be updated to a version which supports node6.
Some refactor work included:
* Split out graphql queries to separate state files for easier management.
* Tweak some of the `jest.mock()` and `jest.unmock()` calls made, as auto-mocking is turned off by default.
* Fix name-case of imported modules.


tractorcow:
History Form seems broken...

tractorcow:
Hm, look at https://github.com/silverstripe/silverstripe-framework/pull/6527/files

You moved the .form to .meta.form, but the assignment in framework wasn't moved.

tractorcow:
ok, fixed it, but requires this patch https://github.com/silverstripe/silverstripe-framework/pull/6694

PR #406, opened 2017-03-21
Re-instate the insert oembed function

flamerohr:
I've split the functionality of oembed to its own form and plugin, so insert media and oembed can be independently removed (if desired)

See https://github.com/silverstripe/silverstripe-framework/issues/6480 for details

Depends on these PRs:
https://github.com/silverstripe/silverstripe-admin/pull/6
https://github.com/silverstripe/silverstripe-framework/pull/6727


tractorcow:
Shortcodes given a rather major cleanup; I ran into a bug I couldn't resolve @flamerohr , please see the `@todo` in the code :P

tractorcow:
Ah, fixed it.

Issue is Url isn't posted back since it's a readonly field. Added it as hidden field and it's good to go.

client/src/entwine/TinyMCE_ssembed.js
```diff
@@ -0,0 +1,375 @@
...
+        .addClass(data.Placement)
+        .addClass('ss-htmleditorfield-file embed');
+
+      // Add placeholder image
+      const placeholder = jQuery('<img />')
+        .attr('src', data.PreviewUrl)
+        .addClass('placeholder');
+
+      // Set dimensions
+      if (data.Width && data.Height) {
```
flamerohr: Same as above with width and height being coupled

client/src/entwine/TinyMCE_ssembed.js
```diff
@@ -0,0 +1,375 @@
...
+          .text(data.CaptionText);
+        base.append(caption);
+      }
+
+      // Find best place to put this embed
+      const node = $(editor.getSelectedNode());
+      let replacee = $(null);
+      if (node.length) {
+        // Find find closest existing embed
+        replacee = node.closest('[data-shortcode=\'embed\']');
```
flamerohr: As above, the selected node could be the embed element

client/src/entwine/TinyMCE_ssembed.js
```diff
@@ -0,0 +1,375 @@
...
+
+        // Fail over to closest image
+        if (replacee.length === 0) {
+          replacee = node.closest('img');
```
flamerohr: I'm not sure how this would work, since `` cannot contain children?

client/src/entwine/TinyMCE_ssembed.js
```diff
@@ -0,0 +1,375 @@
...
+        }
+
+        // Replace existing node
+        if (replacee.length === 0) {
+          replacee = node;
```
flamerohr: This could end up replacing a paragraph, which we wouldn't want

client/src/entwine/TinyMCE_ssembed.js
```diff
@@ -0,0 +1,375 @@
...
+            .addClass(data.class)
+            .addClass('ss-htmleditorfield-file embed');
+
+          // Add placeholder
+          const placeholder = jQuery('<img />')
+            .attr('src', data.thumbnail)
+            .addClass('placeholder');
+
+          // Set dimensions
+          if (data.width && data.height) {
```
tractorcow: Probably could set them independently, although it's probably going to be both or neither in practice. :)

client/src/entwine/TinyMCE_ssembed.js
tractorcow: Yep good. :)

client/src/entwine/TinyMCE_ssembed.js
tractorcow: See above; The container has the data-shortcode, so it'll be matched every time.

client/src/entwine/TinyMCE_ssembed.js
tractorcow: It works if node === an image. It's not really clear though is it?

client/src/entwine/TinyMCE_ssembed.js
tractorcow: Yeah maybe make it inserted into the paragraph rather than replace it?

flamerohr:
thanks @tractorcow, the changes you've pushed up looks good

tractorcow:
I've fixed up the CS linting issues.

PR #410, opened 2017-03-22
FIX thumbnail displays in table view

phalkunz:


client/src/containers/TableView/TableView.js
```diff
@@ -332,21 +332,25 @@ class TableView extends Component {
    */
   renderThumbnail(props) {
     const url = props.data || props.rowData.url;
+    const category = props.rowData.category;
```
flamerohr: category could be empty value, add a ` || 'false'` to the end like in `GalleryItem`

PR #411, opened 2017-03-24
Update class namespace for campaign admin

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/6735

### DIFF
--- a/code/Controller/AssetAdmin.php
+++ b/code/Controller/AssetAdmin.php
@@ -3,7 +3,7 @@
 namespace SilverStripe\AssetAdmin\Controller;
 
 use InvalidArgumentException;
-use SilverStripe\Admin\AddToCampaignHandler;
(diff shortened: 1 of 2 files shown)


sminnee:
Needs a rebase — looks like it'll become a 1 line change.

tractorcow:
done.

chillu:
Given the `asset-admin` module depends on campaign admin functionality, should they require it in `composer.json`? e.g. for `AddToCampaignHandler`

chillu:
Damian agrees that we need that dependency, so I've added it here.

sminnee:
OK that's tolerable for now – i.e. doesn't block merge – but a weird piece of coupling.

tractorcow:
In the future we can de-couple it with a few class_exists(), or maybe just an extension hook.

PR #415, opened 2017-04-04
Move create folder to right editor panel

flamerohr:
So we're no longer using a prompt popup!

See https://github.com/silverstripe/silverstripe-framework/issues/5627 for details

### DIFF
--- a/client/src/containers/AssetAdmin/AssetAdmin.js
+++ b/client/src/containers/AssetAdmin/AssetAdmin.js
@@ -15,7 +15,7 @@ import { withApollo } from 'react-apollo';
 import Search, { hasFilters } from 'components/Search/Search';
 import readFilesQuery from 'state/files/readFilesQuery';
 import deleteFileMutation from 'state/files/deleteFileMutation';
-import CONSTANTS from 'constants';
+import CONSTANTS from 'constants/index';
 
 class AssetAdmin extends SilverStripeComponent {
 
@@ -301,7 +301,7 @@ class AssetAdmin extends SilverStripeComponent {
         }
         // TODO Update GraphQL store with new model,
         // see https://github.com/silverstripe/silverstripe-graphql/issues/14
-        this.props.actions.files.readFiles()
+        const readFiles = this.props.actions.files.readFiles()
           .then(() => {
             // open the containing folder, since folder edit mode isn't desired
             if (action !== 'action_createfolder' || ['admin'].indexOf(this.props.type) === -1) {
@@ -309,7 +309,7 @@ class AssetAdmin extends SilverStripeComponent {
             }
           });
 
-        return response;
+        return Object.assign({}, response, { readFiles });
       });
   }
 
--- a/client/src/containers/AssetAdmin/AssetAdmin.js
+++ b/client/src/containers/AssetAdmin/AssetAdmin.js
@@ -15,6 +15,7 @@ import { withApollo } from 'react-apollo';
 import Search, { hasFilters } from 'components/Search/Search';
 import readFilesQuery from 'state/files/readFilesQuery';
 import deleteFileMutation from 'state/files/deleteFileMutation';
+import CONSTANTS from 'constants';
 
 class AssetAdmin extends SilverStripeComponent {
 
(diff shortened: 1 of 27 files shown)


code/Forms/AssetFormFactory.php
```diff
@@ -56,17 +57,33 @@ public function getForm(Controller $controller, $name = FormFactory::DEFAULT_NAM
...
+    
+    /**
+     * Get the validator for the form to be built
+     *
+     * @param $controller
+     * @param $name
+     * @param $context
+     * @return RequiredFields
+     */
+    protected function getValidator(Controller $controller, $name, $context = []) {
```
phalkunz: Why declaring arguments when they're not being used and the method doesn't implement that of the interface?

code/Forms/AssetFormFactory.php
```diff
@@ -56,17 +57,33 @@ public function getForm(Controller $controller, $name = FormFactory::DEFAULT_NAM
...
+        $validator = new RequiredFields('Name');
```
phalkunz: Should this use the name supplied in the `name` argument?

client/src/containers/AssetAdmin/AssetAdminRouter.js
```diff
@@ -4,28 +4,40 @@ import { withRouter } from 'react-router';
 import AssetAdmin from 'containers/AssetAdmin/AssetAdmin';
 import { decodeQuery } from 'lib/DataFormat';
 import qs from 'qs';
+import CONSTANTS from 'constants/index';
 
 const sectionConfigKey = 'SilverStripe\\AssetAdmin\\Controller\\AssetAdmin';
 
+const actions = Object.keys(CONSTANTS.ACTIONS).map((key) => CONSTANTS.ACTIONS[key]);
```
flamerohr: unfortunately `Object.values()` isn't going to pass jest tests (it's still an experimental es2017 feature)
If even es2016 polyfills do not support this, I'm not sure about adding that feature in yet.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -457,29 +471,40 @@ class AssetAdmin extends SilverStripeComponent {
...
-      default:
-        schemaUrl = config.form.fileEditForm.schemaUrl;
-        break;
+    let targetId = null;
+
+    if (this.props.viewAction === CONSTANTS.ACTIONS.CREATE_FOLDER) {
+      schemaUrl = config.form.folderCreateForm.schemaUrl;
+      targetId = this.props.folderId;
     }
+    else if (this.props.viewAction === 'edit') {
```
flamerohr: good catch, I missed this!

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -294,11 +295,21 @@ class AssetAdmin extends SilverStripeComponent {
     }
     return promise
       .then((response) => {
+        if (action === 'action_createfolder' && ['admin'].indexOf(this.props.type) > -1) {
```
flamerohr: yup sure, but it'll need to be strict typing equals :)

code/Controller/AssetAdmin.php
```diff
@@ -765,7 +769,44 @@ public function fileHistoryForm()
...
+                throw new \InvalidArgumentException(sprintf(
+                    '%s#%s not found',
+                    Folder::class,
+                    $parentID
+                ));
+            }
+        }
+    
+        // Check permission
+        if (!Folder::singleton()->canCreate(Member::currentUser(), $data)) {
```
flamerohr: `canCreate()` shouldn't have an instance to call on yet, it hasn't been created (check `apiCreateFile` and it also creates a singleton class)
The `canCreate()` call gets a "ParentID" passed through, which in turn can check permissions for the parent folder if desired, but I don't think it should be the default behaviour.

This can be arguably a static method, but that's an API change.

On that note: I'll add the Parent record to the context data passed to `canCreate()`

code/Forms/AssetFormFactory.php
flamerohr: The `getValidator()` of a child could use those parameters, but for our case it's not used.

I used the same parameters as `getFormFields()` and `getFormActions()`, as I felt they can be closely related. Since those two functions can change the fields returned depending on the parameters, you would expect the validator to also change to match.

code/Forms/AssetFormFactory.php
flamerohr: Nope, the `$name` is the Form's name given from `getForm()`

I'll rename the variable to be a bit more concise

client/src/containers/AssetAdmin/AssetAdminRouter.js
phalkunz: Ah true!

tractorcow:
I've tracked some JS errors in firefox to this pull request.

Even though the build is green, there are JS errors appearing.

https://travis-ci.org/silverstripe/silverstripe-asset-admin/jobs/220364644

PR #425, opened 2017-04-10
WIP Thumbnail spacing improvements

clarkepaul:
Improvements for the issue https://github.com/silverstripe/silverstripe-asset-admin/issues/470
Also includes minor visual improvements:
Resizing of file icons like docs, video, audio
Increase visibility of drop area of the "Up folder" icon
Change preview image icon to enlarge icon rather than search icon


flamerohr:
Moved to beta2 as the pull requests haven't aged very well (could be potentially not relevant anymore)
Will be better to investigate properly, than rush and push this over the line

client/src/containers/TableView/TableView.scss
```diff
@@ -137,6 +137,14 @@ $gallery-list-thumb-height: 68px;
   width: $gallery-list-thumb-width;
 }
 
+.gallery__table-image--folder,
+.gallery__table-image--archive,
+.gallery__table-image--audio,
+.gallery__table-image--video,
+.gallery__table-image--document {
+  background-size: calc(#{$gallery-list-thumb-width} - #{$panel-padding-x}) auto; // Allow for space either side of folder
+}
```
flamerohr: @clarkepaul do you remember why you targeted each file type like this instead of targeting the generic `.gallery__table-image` class?

flamerohr:
I've squashed and rebased the commits to the `1` branch

client/src/containers/TableView/TableView.scss
clarkepaul: From memory the icons were taking the full height in list view (like images should do), but other files which aren't files need some space surrounding them, I've taken a look at list view and that doesn't seem to be a problem anymore. Maybe not relavent?

flamerohr: Yeah, it potentially could be irrelevant. I'll play around with the style a little bit and make the relevant changes if needed :)

flamerohr: I've shifted this to be the default and what was in `.gallery__table-image` is now only applied to images, since this list is basically only not targeting images

flamerohr:
Rebased, I have smoked tested and works fine, will merge on green.

PR #428, opened 2017-04-13
BUG Ensure that UploadField.ss retains selected values if never rendered by react

### LINKED ISSUE #400: Page with UploadField loses file association upon saving from another tab
## Summary

A `Page` subclass has an `UploadField` on it's Main tab, and additional fields on a second tab.  Upon saving while the second tab is active, the association is stored.  However, saving a second time from the second tab, the association is lost.

## Steps to Reproduce

1. Create a `Page` or `SiteTree` subclass with an `UploadField` on the main tab, and some other fields on second tab.
2. Upload a file via the `UploadField` on the main tab.
3. Switch to the second tab.
4. Save the record.
5. Save the record once more.

## Expected Results

The file association remains in place after saving multiple times while other tabs are active.

## Actual Results

The file association has vanished!

## Environment

- SilverStripe 4 `dev-master`
- PHP 5.6.29
- Apache 2.4.23 (x64)
- Firefox Developer Edition 54.0a2 (2017-03-12) (x64)
- Windows 10 Pro x64 1607 Build 14393.693

tractorcow:
Fixes #400

### DIFF
--- a/templates/SilverStripe/AssetAdmin/Forms/UploadField.ss
+++ b/templates/SilverStripe/AssetAdmin/Forms/UploadField.ss
@@ -1,1 +1,8 @@
 <input $AttributesHTML <% include SilverStripe/Forms/AriaAttributes %> />
+<% if $Items %>
+    <div class="entwine-placeholder">
+        <% loop $Items %>
+            <input type="hidden" name="$Up.Name[Files][]" value="$ID.ATT" />
+        <% end_loop %>
+    </div>
+<% end_if %>


phalkunz:
I have the following test code:
```
<?php
use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\AssetAdmin\Forms\UploadField;

class Page extends SiteTree
{
    private static $db = array(
    );

    private static $has_one = array(
        'File' => File::class
    );

    public function getCMSFields()
    {
        $fields->push(UploadField::create('FileID', 'File'));
        return $fields;
    }
}
```
It still doesn't retain the uploaded file after saving or publishing.

phalkunz:
The issue is still reproducible with the following test code:
```
<?php
use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\AssetAdmin\Forms\UploadField;
use SilverStripe\Assets\File;

class Page extends SiteTree
{
    private static $db = array(
    );

    private static $has_one = array(
        'File' => File::class
    );

    public function getCMSFields()
    {
        $fields->push(UploadField::create('File', 'File'));
        return $fields;
    }
}
```

## Steps to reproduce:
1. Upload a file
2. Click the "Settings" tab
3. Click "Published" button
4. Switch back to the "Content" tab. The uploaded file is no longer there.

tractorcow:
My test code

```php
<?php

use SilverStripe\AssetAdmin\Forms\UploadField;
use SilverStripe\CMS\Model\SiteTree;

class Page extends SiteTree
{
	private static $has_one = [
	    'Image' => Image::class
    ];

	public function getCMSFields()
    {
        $fields->addFieldToTab('Root.Image', new UploadField('Image', 'Image'));
        return $fields;
    }
}

```

What you're running into is the fact that UploadField doesn't seem to work with unsaved field navigation. That's a real issue, but not the one we are trying to fix. :D

tractorcow:
I've logged the change detection issue at https://github.com/silverstripe/silverstripe-asset-admin/issues/432. :D It doesn't need to block this.

PR #429, opened 2017-04-13
BUG Fix insert-media modal not loading correct folder

### LINKED ISSUE #396: Wrong folder open for opened file
This happens in particularly with InsertMediaModal, where you have a file selected to modify attributes for.

It will open the file properly in the Editor, but you cannot save the modified data because the Gallery has the wrong folder open (some metadata relies on the proper data being loaded in the Gallery view)

Potential fix is remove the reliance on the Gallery view data and have all data ready in the Editor.
Alternatively, could setup a check and redirect if in a folder which doesn't contain the opened file. I find the former better because of its decoupling from Gallery.

Fix at https://github.com/silverstripe/silverstripe-asset-admin/pull/429

tractorcow:
API Allow files to be queried by a child ID if root ID is not known
Fixes #396


code/GraphQL/FileFilterInputTypeCreator.php
```diff
@@ -104,6 +108,14 @@ public function filterList(Filterable $list, $filter)
             }
         }
 
+        // Filter unknown id by known child
+        if (isset($filter['anyChildId'])) {
+            /** @var File $child */
+            $child = File::get()->byID($filter['anyChildId']);
+            $id = $child ? ($child->ParentID ?: 0) : 0;
```
flamerohr: wouldn't `$child->ParentID` return `0` when it's falsey anyway?

tractorcow: It could return `(string)"0"` depending on the DB connector. I could force a cast to int instead of the extra ternary.

flamerohr:
Added patch to load gallery when no file is selected (or search is applied) to fix behat tests

PR #431, opened 2017-04-18
Fix: readFiles promise was hanging in redux, removed that

flamerohr:
Related https://github.com/silverstripe/silverstripe-admin/pull/32

### DIFF
--- a/client/src/containers/AssetAdmin/AssetAdmin.js
+++ b/client/src/containers/AssetAdmin/AssetAdmin.js
@@ -301,15 +301,14 @@ class AssetAdmin extends SilverStripeComponent {
         }
         // TODO Update GraphQL store with new model,
         // see https://github.com/silverstripe/silverstripe-graphql/issues/14
-        const readFiles = this.props.actions.files.readFiles()
+        return this.props.actions.files.readFiles()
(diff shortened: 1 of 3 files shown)


client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -301,15 +301,14 @@ class AssetAdmin extends SilverStripeComponent {
         }
         // TODO Update GraphQL store with new model,
         // see https://github.com/silverstripe/silverstripe-graphql/issues/14
-        const readFiles = this.props.actions.files.readFiles()
+        return this.props.actions.files.readFiles()
```
tractorcow: So this `then` inside the outer promise returns another promise?

flamerohr: yup that's right, so if a promise.then returns a promise, that will add to the chain:

```
Promise.then(() => {
  /* do something here */
  return Promise.then(() => /* do this with closure function */);
});
```

is the same as this without any closured variables:
```
Promise.then(() => {
  /* do something here */
})
.then(() => /* do this with another function */);
```

PR #433, opened 2017-04-18
Pulls/4.0/insert image constraint dimensions

unclecheese:


flamerohr:
@chillu was the aspect ratio constraint always on in ss3? perhaps we could add the switch to turn it on or off as another story, what do you think?

flamerohr:
@unclecheese I've pushed a cleanup commit and noted down what I've done in the commit message

unclecheese:
Looks good to me.

Issue #422, opened 2017-04-06
Minor spacing issues of thumbnails and back arrow


clarkepaul:
Files require 20px gap left of the thumbnail in this case and the up folder action should only be 10px from the left (both files list view and insert media list view).

<img width="117" alt="pasted_image_6_04_17__10_49_pm" src="https://cloud.githubusercontent.com/assets/555033/24750929/d86bee6c-1b1b-11e7-9656-f688e48141f8.png">

chillu:
Pointers how to fix: `TableView.js` in asset-admin has a `renderThumbnail()` method. Styling is done in `TableView.scss`

phalkunz:
This is assigned to Mikaela Y

phalkunz:
I think we shouldn't fix the folder action buttons group alignment in this issue. This isn't simple as it looks because the buttons have hover state which causes the buttons have different bounding boxes when in normal and hover states. @clarkepaul I think we talked about hover states earlier. There should be another issue to address this.

mikaelayoung:
Merge Request: https://github.com/silverstripe/silverstripe-asset-admin/pull/436

PR #436, opened 2017-04-20
Add thumbnail spacing in file modal list view

mikaelayoung:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/422

### DIFF
--- a/client/src/containers/TableView/TableView.scss
+++ b/client/src/containers/TableView/TableView.scss
@@ -121,6 +121,11 @@ $gallery-list-thumb-height: 68px;
   .gallery__table .gallery__table-row > & {
     padding: 0;
     vertical-align: middle;
+
+    &:first-child {
+      max-width: none;
+      padding-left: 20px;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


client/src/containers/TableView/TableView.scss
```diff
@@ -121,6 +121,10 @@ $gallery-list-thumb-height: 68px;
   .gallery__table .gallery__table-row > & {
     padding: 0;
     vertical-align: middle;
+
+    &:first-child {
+      padding-left: 20px;
```
flamerohr: Could we use `$table-cell-padding` here instead? :)

client/src/containers/TableView/TableView.scss
```diff
@@ -168,6 +172,12 @@ $gallery-list-thumb-height: 68px;
   }
 }
 
+.gallery__table-column--title {
+  &:nth-child(2) {
+    padding-left: 36px;
```
flamerohr: same here, could we use `$table-lg-cell-padding` instead?

PR #443, opened 2017-04-22
API Update to behat3

tractorcow:
See https://github.com/silverstripe/silverstripe-behat-extension/pull/152


behat.yml
```diff
@@ -22,8 +22,8 @@ default:
             - %paths.modules.asset-admin%/tests/behat/files/
   extensions:
     SilverStripe\BehatExtension\Extension:
-      bootstrap_file: %paths.modules.cms%/tests/behat/serve-bootstrap.php
-      screenshot_path: %behat.paths.base%/artifacts/screenshots
+      bootstrap_file: cms/tests/behat/serve-bootstrap.php
```
sminnee: Why isn't this paths.module.cms?

If I check out installer and run `vendor/bin/behat @asset-admin` I think this hard-coding will break?

client/src/components/GalleryItem/GalleryItem.scss
```diff
@@ -39,7 +39,7 @@ $gallery-item-label-height: 40px;
   height: $gallery-folder-height;
 
   .gallery-item__thumbnail {
-    background: url("../images/folder.png") 12px 14px no-repeat;
+    background: url("../../images/folder.png") 12px 14px no-repeat;
```
sminnee: Why were these path changes needed? Why did it work before?

code/Controller/AssetAdmin.php
```diff
@@ -129,9 +130,10 @@ public function init()
     {
         parent::init();
 
-        Requirements::add_i18n_javascript(ASSET_ADMIN_DIR . '/client/lang', false, true);
-        Requirements::javascript(ASSET_ADMIN_DIR . "/client/dist/js/bundle.js");
-        Requirements::css(ASSET_ADMIN_DIR . "/client/dist/styles/bundle.css");
+        $module = ModuleLoader::getModule('silverstripe/asset-admin');
```
sminnee: Great for now but it would be good to have a more concise idiom for this in future.

composer.json
```diff
@@ -21,6 +21,9 @@
       "dev-master": "1.0.x-dev"
     }
   },
+  "scripts": {
```
sminnee: Where's the phpcs file? Can we put it in the default location? Is it possible to include the `code/ tests/` definition in that file somehow?

behat.yml
tractorcow: everything except `suites` config is ignored when imported, so we can rely on this path for the time being.

The reason it's removed is, apparently, the issue with `%behat.paths.base%` not evaluating for some reason. The behat extension was trying to evaluate the path literally with the placeholder.

It can be fixed but requires another behat-extension release. This is a patch for now.

client/src/components/GalleryItem/GalleryItem.scss
tractorcow: Because those were hard-coded to dist files that weren't in source folder. I moved the files to the source folder, and ensured that the build re-generated them, and re-wrote the dist/*.css correctly.

The reason I moved these files was due to certain framework images not being findable by the webpack builder due to indeterminable framework path.

This change means that deleting dist folder and running rebuild correctly rebuilds the dist, where before it wasn't.

composer.json
tractorcow: It's moved to the root and renamed to the pattern that phpcs checker automatically finds.

sminnee: :+1:

client/src/components/GalleryItem/GalleryItem.scss
sminnee: :+1:

behat.yml
sminnee: Maybe it's because that setting has change to `%paths.base%`, and other code hasn't been updated to suit?

In any case, doesn't block this PR, so :+1:

tractorcow: I tried that, but it failed too. I'm not sure why, but didn't have time to investigate too deeply.

flamerohr:
Sam's happy with the approach and I'm pretty happy with the scss/js fixes as well, let's get this merged :)

PR #444, opened 2017-04-24
WIP: Pulls/4.0/i18n namespacing

unclecheese:


client/lang/src/el.js
```diff
@@ -1,4 +1,100 @@
 {
-    "AssetAdmin.ConfirmDelete": "Do you really want to delete this folder and all contained files?",
-    "Folder.Name": "Folder name"
-}
+    "AssetAdminFileField.BROWSE": "Browse",
```
tractorcow: This file hasn't been upgraded properly. Some obsolete keys still.

lang/en.yml
```diff
@@ -1,49 +1,24 @@
 en:
```
tractorcow: Can you please sort the keys alphabetically?

tractorcow:
I've fixed some issues with the upgrader. We may need to re-do yml migration with https://github.com/silverstripe/silverstripe-upgrader/pull/31

Issue #349, opened 2017-01-05
Show message when current folder or file record can't be found


chillu:
When opening `admin/assets/show/1`, if a Folder with `ID=1` doesn't exist, it simply shows an empty screen - not great. We should display a friendly error message, with an option to navigate to the root folder.

Note that this is different from a file not being found in the "edit file" preview (see https://github.com/silverstripe/silverstripe-asset-admin/issues/265)

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/446
- [x] https://github.com/silverstripe/silverstripe-admin/pull/41

flamerohr:
The issue is GraphQL doesn't provide an indication that loading has started/stopped, and so even though it has finished loading, the components that are connected to GraphQL still thinks it's loading data.

clarkepaul:
When does admin/assets/show/1 get used except when you click on on the actual folder, I must be missing something as navigating to the files area takes you to admin/assets/.

PR #446, opened 2017-04-26
Show message when viewing file or folder that does not exist

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/349


phalkunz:
@flamerohr Thanks. Is there any reason why you don't have `lint` in `yarn watch`. I think we should.

flamerohr:
webpack doesn't support that? at least I haven't seen it :(

Plus build times are very slow already, adding linting on top and it's going to get difficult to develop in.

phalkunz:
True about the performance. Well, I'll get eslint working in vim. Cheers.

PR #456, opened 2017-05-01
Fix Set dimensions constraints to Embed height and width

flamerohr:
So that behaviour is consistent across Insert Media modal and Insert Embed modal


tractorcow:
I've noticed this issue, but it's not caused by this fix, so not a blocker.


image

PR #458, opened 2017-05-02
Remove `cms` and `campaign-admin` dependencies.

### LINKED ISSUE #450: Make asset-admin usable without the cms and campaign-admin dependencies
Asset admin should be usable without having to install `cms` and `campaign-admin`. 

composer.json lists these as requirements, but from a first glance, these modules aren't really required for asset-admin to work. Deleting the folders manually doesn't cause any errors, except that the **Add to campaign** button no longer works (just opens an empty dialog):

Add to campaign button

Is this the only part of asset-admin that relies on `campaign-admin`? Why is there a `cms` requirement?

bummzack:
Add `admin` dependency instead.
Add form-button that is specific to `campaign-admin` via an extension that is only applied when that module is present.
Explicitly set the popover actions field name for easier targeting in extensions and tests.
Updated unit-tests (test form fields with- and without extension applied).

Fixes #450

### DIFF
--- a/code/Extensions/CampaignAdminExtension.php
+++ b/code/Extensions/CampaignAdminExtension.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace SilverStripe\AssetAdmin\Extensions;
+
+use SilverStripe\Assets\File;
+use SilverStripe\Core\Extension;
+use SilverStripe\Forms\FormAction;
+
+/**
+ * Extension that updates the Popover menu of `FileFormFactory`.
+ * This extension will only be applied if the `campaign-admin` module is installed.
+ * @package SilverStripe\AssetAdmin\Extensions
+ */
+class CampaignAdminExtension extends Extension
+{
+    /**
+     * Update the Popover menu of `FileFormFactory` with the "Add to campaign" button.
+     *
+     * @param array $actions
+     * @param File $record
+     */
+    public function updatePopoverActions(&$actions, $record)
+    {
+        if ($record && $record->canPublish()) {
+            array_unshift($actions, FormAction::create(
+                'addtocampaign',
+                _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.ADDTOCAMPAIGN', 'Add to campaign')
+            ));
+        }
+    }
+}
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -34,3 +34,11 @@ After:
 SilverStripe\Core\Injector\Injector:
   SilverStripe\Forms\FileHandleField:
     class: SilverStripe\AssetAdmin\Forms\UploadField
+
+---
+Only:
+  moduleexists: 'silverstripe/campaign-admin'
+---
+SilverStripe\AssetAdmin\Forms\FileFormFactory:
+  extensions:
+    - 'SilverStripe\AssetAdmin\Extensions\CampaignAdminExtension'
--- a/code/Forms/AssetFormFactory.php
+++ b/code/Forms/AssetFormFactory.php
@@ -218,6 +218,7 @@ abstract class AssetFormFactory implements FormFactory
         if ($popoverActions) {
             return PopoverField::create($popoverActions)
                 ->setPlacement('top')
+                ->setName('PopoverActions')
                 ->setButtonTooltip(_t(
                     'SilverStripe\\AssetAdmin\\Forms\\FileFormFactory.OTHER_ACTIONS',
                     'Other actions'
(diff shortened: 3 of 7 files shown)


code/Extensions/CampaignAdminExtension.php
```diff
@@ -0,0 +1,35 @@
...
+     * Update the Popover menu of `FileFormFactory` with the "Add to campaign" button.
+     *
+     * @param FieldList $actions
+     * @param Controller $controller
+     * @param string $formName
+     * @param array $context
+     */
+    public function updateFormActions($actions, $controller, $formName, $context)
+    {
+        if ($popoverField = $actions->fieldByName('PopoverActions')) {
```
tractorcow: Please ensure you keep the permission check `canPublish`, otherwise it'll display actions the current user may not be able to perform.

There is a risk that getPopoverFields() is empty with the removal of this action (e.g. canDelete is false, but canPublish is true). you could fix this by adding an extension point to getPopoverActions() (updatePopoverActions) and instead extending that, instead of updateFormActions. Don't forget to include the extension point in overriding and base implementation. :)

bummzack: Wow, good catch! You're right… if there are no actions (currently just the "delete" action), then the Popover menu won't even be created. Thus the `updateFormActions` triggers too late. I guess the only way is to have an extension point for the popover actions themselves.
Will implement additional test to make sure the different permission scenarios are covered.

bummzack: Currently, `FileFormFactory` builds its own array of actions in `getPopoverActions`, instead of merging with `parent::getPopoverActions`. Can you give some reasoning why this is so? Is it to have better control over button ordering?

When implementing a `updatePopoverActions` extension point, I think it would make sense to have it called from the base-class and use `parent::getPopoverActions` in subclasses? Thoughts?

bummzack:
I'm sorry, I pressed the `Update branch` button and master got merged into this PR :(
I'll create a clean PR if requested.

code/Extensions/CampaignAdminExtension.php
flamerohr: I think it may have been a relic of the past, back when the `Delete` action for folders was not in the `PopoverField`. Could be refactored to be cleaner, but not necessary in this pull request :D

An extension point for the popover actions is a good idea, make sure in subclasses you'll need to call `$this->beforeExtending()` like how `getFormFields()` does.

bummzack: I think your concerns have been addressed with the new commit? See: https://github.com/silverstripe/silverstripe-asset-admin/pull/458/commits/ee2f8dc8a41550b46552c04408935f4bb935bf3d

tractorcow:
Can you rebase and squash locally? Just do a force push when you've cleaned it up on your machine. That's cleaner than making a new PR.

code/Extensions/CampaignAdminExtension.php
flamerohr: you're right! sorry, I didn't see it earlier

bummzack:
@tractorcow Cleaned up to 1 commit now. I think the failing Behat tests aren't caused by me? Seems to be a problem with the setup?

tractorcow:
Well, if you aren't installing CMS anymore, behat can't find the bootstrap file. You need to adjust `.travis.yml` to bootstrap behat without cms module.

tractorcow:
Actually, you should change it so that cms is installed for testing, since behat covers tests that require pages. You can add it to the `composer require` line.

tractorcow:
Same for campaign admin.

tractorcow:
Add these to require-dev also.

bummzack:
I just added `cms` and `campaign-admin` as dev-requirements. It would be nicer, if these could only be included for test-environments, but I guess that's easier said than done…
Now I'm curious to see if the behat tests pass.

flamerohr:
I'm not sure I agree with require-dev, perhaps a suggest would be less intrusive?

bummzack:
@flamerohr I added it, so that behat tests work (they seem to rely on `campaign-admin` and `cms`). It makes sense to want to test the File upload field within a cms page context… I think the cleanest solution would be to only include these dependencies in tests. But I'm not sure how it could be achieved.

flamerohr:
@bummzack ah yeah, I understand that
hmm... if only tests require it, possibly could be configured with the `.travis.yml` perhaps, I guess that's a bit too hidden/magical as well

flamerohr:
ok, I need to understand composer more, cms and campaign-admin are only included when you composer install in the asset-admin module itself :D

PR #460, opened 2017-05-02
MINOR: Don't notify Slack on PR failure

sminnee:


tractorcow:
Should we keep build successes? We can use `on_failure: never`

chillu:
Heh damn, already committed that - and removed notifications altogether. Let's revisit notifications from the 17 repos with `on_pull_requests: false` after we get the combined build going (which should cover most of those cases).

Issue #367, opened 2017-01-16
Bulk action delete triggers view switch


flamerohr:
Deleting a file or folder causes the view to change back to default gallery/thumbnail view, when it should stay in list/table view.

Deleting item(s) should also preserve all display queries:
- Pagination
- Sorting
- Search
- Filter
- ...

## Excludes:

- Deleting last item(s), user shouldn't see an empty page (https://github.com/silverstripe/silverstripe-asset-admin/issues/466)

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/463

chillu:
Confirmed this is still an issue.

phalkunz:
This also happens with single item deletion as well.

PR #463, opened 2017-05-03
Fix issue when deleting files, the gallery always change to thumbnail…

phalkunz:
…s view

Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/367

### DIFF
--- a/client/src/containers/AssetAdmin/AssetAdmin.js
+++ b/client/src/containers/AssetAdmin/AssetAdmin.js
@@ -382,7 +382,8 @@ class AssetAdmin extends SilverStripeComponent {
 
       // redirect to open parent folder if the file/folder is open and on screen to close it
       if (file) {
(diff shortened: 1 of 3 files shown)


client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -382,7 +382,8 @@ class AssetAdmin extends SilverStripeComponent {
 
       // redirect to open parent folder if the file/folder is open and on screen to close it
       if (file) {
-        this.handleBrowse((file.parentId) ? file.parentId : 0);
+        const view = this.props.query && this.props.query.view;
+        this.handleBrowse((file.parentId) ? file.parentId : 0, null, { view });
```
flamerohr: This could force the gallery to jump back to page 1 after each delete, would it be worth adding a check if `file.id === folder.id` and add pagination appropriately?

Issue #461, opened 2017-05-03
Load protected thumbnails inlined


chillu:
## Acceptance Criteria

  - When an image is loaded into asset admin by graphql, thumbnail urls are only returned for public assets that have a thumbnail.
  - When an image is loaded into asset admin by graphql, protected images that have thumbnails will have their content transmitted in the same request via a field.
 - Memory usage and processing times in browsers does not significantly increase

## Excludes

 * Deferred generation of thumbnails
 * Resolution of potential memory exhaustion issues (which largely already exist in SS3)
 * No pregeneration of thumbnails on upload (same behaviour as SS3)

tractorcow:
Testing results:

 - 50 thumbnail images - PNG format (uncompressed) - 9.6MB payload, 1.4 seconds load.
 - 50 thumbnail images - JPG format (compressed) - 1.5mb, 1.0 seconds load.
 
 I deem it acceptable. :)

tractorcow:
I've decreased thumbnail size slightly, so that we can still get @2 retina images for standard gallery thumbnails.

 - 50 thumbnail images - PNG format (uncompressed) - 7.6MB payload, 1.4 seconds load.
 - 50 thumbnail images - JPG format (compressed) - 1.3mb, 1.0 seconds load.

tractorcow:
![image](https://cloud.githubusercontent.com/assets/936064/25835530/948677d6-34d3-11e7-82dd-77410efbdbe6.png)

Memory usage while navigating through various pages of both JPG or PNG files seemed to fluctuate between 30 and 150 MB of usage.

Initial requests to generate thumbnails could be quite slow, but any subsequent request were fast (see prior comment for benchmarks). We will address pre-generation of thumbnails in another story.

PR #468, opened 2017-05-09
Update pagination to 50 and inline thumbnails

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/461


code/GraphQL/FileTypeCreator.php
```diff
@@ -16,6 +19,14 @@
 class FileTypeCreator extends TypeCreator
 {
     /**
+     * Safely limit max inline thumbnail size to 200kb
+     *
+     * @config
+     * @var int
+     */
+    private static $max_thumbnail_bytes = 200000;
```
chillu: I think this should be on `AssetAdmin` - `FileTypeCreator` in an implementation more than a "config key".

tractorcow: Good idea, I'll move it. Or @flamerohr can do that tomorrow if you want it faster. :)

Issue #177, opened 2016-05-31
File sidebar form fields not part of tabindex (not keyboard navigable)


chillu:
## Tasks
- [x] Auto-focus happens for the Editor when it is opened (either by Editing a file/folder, or creating a folder)
- [x] Gallery is hidden (display none) when on small screen, not just covered by the Editor.
- [x] Gallery and Editor is both shown (and can both be tabbed through) in that order when the screen is large enough.

## Pull requests:

- [x] https://github.com/silverstripe/silverstripe-framework/pull/6919
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/473
- [x] https://github.com/silverstripe/silverstripe-admin/pull/68

clarkepaul:
They are but they appear before galleryItems. I moved the Editor panel before the Gallery so that the Gallery could be resized based on whether or not the Editor markup was present. We could change the order of Gallery and Editor back if there was a bit of JS to say if Editor is present.

chillu:
@clarkepaul How much effort is it to move editor markup after gallery markup? It's an accessibility issue, right?

clarkepaul:
There is a good side to the way it currently is set up. Markup wise I think it makes more sense having the edit panel positioned before the gallery items, it also means the gallery resizes on the remainder of the space based on what has already been used by the edit panel. On smaller devices it actually makes more sense for tabbing to go from the menu to the selected item as it consumes the entire screen with the gallery living underneath.

The bad part is that tabbing through can be confusing so if possible I would prefer to adjust the ordering of the tabindex for larger screens only rather than the structure of markup.

chillu:
> On smaller devices it actually makes more sense for tabbing to go from the menu to the selected item as it consumes the entire screen with the gallery living underneath.

I think only visible elements can be tabbed into, so that's a non-starter.

How about we just focus the first input field when opening the sidebar? That should guide screen readers enough as well. We're already doing this when opening a page from the CMS tree.

scott1702:
> How about we just focus the first input field when opening the sidebar?

+1

Having the editor markup after the gallery would be annoying as a user would have to tab through all the images in the folder to get to the editor (would be helped a little by pagination). As it currently is it's also unclear that the user should tab through all the elements in the page again to get to the editor (or shift+tab). I agree that focusing on the first input field when opening the sidebar would be the best approach here.

clarkepaul:
[...]

PR #473, opened 2017-05-12
Asset edit form tab index

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/177

### DIFF
--- a/code/Forms/FileFormFactory.php
+++ b/code/Forms/FileFormFactory.php
@@ -72,7 +72,8 @@ class FileFormFactory extends AssetFormFactory
         // Update details tab
         $tab = Tab::create(
             'Details',
-            TextField::create("Title", File::singleton()->fieldLabel('Title')),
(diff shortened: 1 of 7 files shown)


code/Forms/FileFormFactory.php
```diff
@@ -72,7 +72,8 @@ protected function getFormFieldDetailsTab($record, $context = [])
         // Update details tab
         $tab = Tab::create(
             'Details',
-            TextField::create("Title", File::singleton()->fieldLabel('Title')),
+            TextField::create("Title", File::singleton()->fieldLabel('Title'))
+                ->setAttribute('autofocus', true),
```
flamerohr: I don't think having the `autofocus` html5 attribute is a good idea when you're also using the `autoFocus` property in the `FormBuilder`, I could see this leading to potential race conditions down the line.

flamerohr:
restarting tests :D

Issue #183, opened 2016-06-02
Make "sort" dropdown keyboard accessible


chillu:
The "sort by..." item on the top right of admin/assets can't receive a keyboard focus. It's using Chosen.js which should already be keyboard friendly.

sheadawson:
`.gallery__sort {
  .chosen-container-active {
    .chosen-single {
      outline: $gray-dark dotted thin; // firefox, ie
      outline: -webkit-focus-ring-color auto 4px; // webkit
    }
  }
}`

Didn't quite get this finished, but here's a start

bergice:
Looks like this is resolved.

PR #481, opened 2017-05-19
MINOR: Add missing focus outline to file sort

lukereative:
Related Issue: #183


client/src/containers/Gallery/Gallery.scss
```diff
@@ -93,7 +93,13 @@
   display: inline-block;
   vertical-align: middle;
   padding: $btn-padding-y $btn-padding-x;
-  margin-bottom: -4px;
+
+  .chosen-container-active {
+    .chosen-single {
+      outline: 1px dotted #212121;
```
robbieaverill: Possibly worth using a variable here if you can @lukereative

flamerohr: yup, we have `$component-active-border-color` that could be used, it's a light blue colour, as opposed to the black-grey defined here :)

PR #487, opened 2017-05-23
Use new Injector API

unclecheese:
Dependency: https://github.com/silverstripe/silverstripe-admin/pull/70
https://github.com/silverstripe/silverstripe-admin/pull/87

### DIFF
--- a/client/src/components/ProportionConstraintField/ProportionConstraintField.js
+++ b/client/src/components/ProportionConstraintField/ProportionConstraintField.js
@@ -1,5 +1,5 @@
 import React, { PropTypes, Component, Children, cloneElement } from 'react';
-import { withInjector } from 'lib/Injector';
+import { inject } from 'lib/Injector';
 
 class ProportionConstraintField extends Component {
 
@@ -35,7 +35,7 @@ class ProportionConstraintField extends Component {
   }
 
   render() {
-    const FieldGroup = this.context.injector.get('FieldGroup');
+    const { FieldGroup } = this.props;
     return (
       <FieldGroup {...this.props}>
         {this.props.children.map((child, key) => (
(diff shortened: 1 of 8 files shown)


client/src/components/ProportionConstraintField/tests/ProportionConstraintField-test.js
```diff
@@ -1,21 +1,29 @@
...
 
-jest.mock('lib/Injector', () => ({
-  // eslint-disable-next-line react/prop-types
-  getComponentByName: () => ({ children }) => <div>{children}</div>,
-}));
-
 import React from 'react';
 import ReactTestUtils from 'react-addons-test-utils';
 import ProportionConstraintField from '../ProportionConstraintField';
+import Injector, { provideInjector } from 'lib/Injector';
```
flamerohr: What we've done with other FormFields was `export { TextField };` so we coudl get a vanilla component without any HOCs
I think it's a cleaner approach than trying to mock injector or other HOC in every test file

unclecheese: Yup, that's right. This can actually be deprecated. `inject()` maps dependencies as props, so we can just mock the dependency, not the whole injector. Will update.

tractorcow:
After merge and fix please restart https://github.com/silverstripe/silverstripe-framework/pull/6939

client/src/components/ProportionConstraintField/ProportionConstraintField.js
```diff
@@ -57,10 +57,18 @@ ProportionConstraintField.propTypes = {
   data: PropTypes.shape({
     ratio: PropTypes.number.isRequired,
   }),
+  FieldGroup: PropTypes.func.isRequired,
```
flamerohr: @unclecheese `PropTypes.element` would be more semantically correct. I've made the change

PR #494, opened 2017-06-01
Fix in mobile view, editor ends up with height 0 due to flexbox

flamerohr:
~Primarily fixes Campaign admin~
<details>

screen shot 2017-06-02 at 11 29 35 am

</details>

Ok, reinvestigated, because I thought this would be great to get in to beta1...
This fixes the Editor being hidden (zero height) in asset-admin mobile view.

The screenshot I provided to campaigns was actually a different pull request I was working on in parallel to this one.


tractorcow:
Has this been tested in safari? (noted the safari specific comment)

flamerohr:
Yes, I had checked this out in Safari (like you, because of that Safari comment) and it seems fine with or without the height rule :)

tractorcow:
Wait, why does asset-admin PR fix campaign admin? What if you have campaign admin without asset-admin installed?

tractorcow:
I can confirm that campaign admin without asset-admin installed is broken. do you have a missing PR somewhere?

tractorcow:
I'll assign back to you .:)

client/src/containers/Editor/Editor.scss
```diff
@@ -2,7 +2,6 @@
   background-color: $body-bg;
   z-index: 2;
   position: absolute;
-  height: 100%; // Safari fix for scroll, should be better fix
```
tractorcow: Need a fix for campaign-admin as well as asset-admin.

flamerohr:
> Wait, why does asset-admin PR fix campaign admin? What if you have campaign admin without asset-admin installed?

This seems like a bigger story than what I'd intend this fix to be for, there's a tight coupling between campaign-admin and asset-admin in the styles... perhaps we could create a new story to investigate and fix this for beta2 instead?

flamerohr:
@tractorcow reinvestigated this - this fixes the Editor being hidden (zero height) in asset-admin mobile view.
I think I was also fixing the campaign-admins screenshot in parallel and ended up getting that wire crossed.

flamerohr:
I think this https://github.com/silverstripe/silverstripe-admin/pull/116 will solve both asset-admin and campaign-admin cleanly :)

Issue #309, opened 2016-11-21
Insert Media modal button label


flamerohr:
If an existing image was selected, then the form button should show "Update" instead of "Insert", so it does not confuse the action of actually inserting the selected.

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-admin/pull/180
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/574

phalkunz:
Pull request: https://github.com/silverstripe/silverstripe-asset-admin/pull/499

flamerohr:
deferred this to beta2, as the fix presented currently was too complex for just a button label change

flamerohr:
Doable with the form injector work now in place :)

clarkepaul:
Isn't this issue irrelevant now that the images have a "Edit" and "Delete" action?

clarkepaul:
Ah ignore my previous comment as I was thinking this issue was about the button to trigger the modal and not the action within the modal.

PR #499, opened 2017-06-09
Pulls/4.0/309 insert update media button label

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/309


chillu:
A separate “file update” form (incl. PHP and schema) isn’t great, yes. It means another server-side roundtrip. I want to get to a place where the “insert image” and “insert link” form schemas are lazy loaded in the background when you first open the CMS, and then cached for a decent amount of time. The “insert link” and “insert image” buttons should present you a UI much quicker. At the moment it makes the CMS feel sluggish. For that reason, I don’t like splitting out “insert image” and “update image” into two separate forms just to change a button label

I agree with Damian that this isn't the right way to fix the bug, sorry :(

PR #500, opened 2017-06-09
Fix change to latest changes for behat-extension

flamerohr:


flamerohr:
Don't need this anymore :D

PR #503, opened 2017-06-16
BUG Ensure UploadField uploads to the controller with the correct Upload object

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/6912

### DIFF
--- a/code/Forms/UploadField.php
+++ b/code/Forms/UploadField.php
@@ -3,7 +3,10 @@
 namespace SilverStripe\AssetAdmin\Forms;
 
 use SilverStripe\AssetAdmin\Controller\AssetAdmin;
+use SilverStripe\Assets\File;
 use SilverStripe\Assets\Folder;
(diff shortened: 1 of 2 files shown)


dhensby:
Behat failure

code/Forms/UploadField.php
```diff
@@ -58,6 +69,10 @@ class UploadField extends FormField implements FileHandleField
     public function __construct($name, $title = null, SS_List $items = null)
     {
         $this->constructFileUploadReceiver();
+
+        // When creating new files, rename on conflict
+        $this->getUpload()->setReplaceFile(false);
```
flamerohr: isn't this false by default anyway? or is this to override a possible config setting which could set it to true?

tractorcow: This is just copied from asset-admin. We need this because we don't have overwrite-confirmation in uploadfield so this avoids that risk.

PR #506, opened 2017-06-20
WIP Feature/react form presentation

unclecheese:
See https://github.com/silverstripe/silverstripe-framework/issues/5630

Dependency: https://github.com/silverstripe/silverstripe-admin/pull/112


chillu:
BTW you raised the PR on the wrong repo :)

client/src/containers/Editor/Editor.js
```diff
@@ -128,16 +129,21 @@ class Editor extends Component {
 
     let error = null;
     if (this.state.loadingError) {
+      const message = this.state.loadingError.value;
       error = (
         <div className="editor__file-preview-message--file-missing">
-          {i18n._t('AssetAdmin.FILE_MISSING', 'File cannot be found')}
+          {(message.contains('Unexpected token < in JSON'))
```
tractorcow: Apparently this is necessary, but I find the output message hiding the underlying error (potential server error) a bit misleading. Is there no better solution here?

flamerohr: There possibly is, my intention for this is to extend the error message to show more than just "File not found" because sometimes the error is from the javascript-side.

The main trouble is that the PHP side is spitting out HTML rather than a JSON with error object, which would be the better way to address this.

PR #514, opened 2017-06-27
FIX: Use new resource reference

sminnee:
Relies on
https://github.com/silverstripe/silverstripe-framework/pull/7065


chillu:
Hmm, Behat breaks in master as well at the moment, but on different issues. This might be actual build failures here?

tractorcow:
I'll fix in separate PR.

Issue #519, opened 2017-07-03
Select / unselect “Limit to current folder and its sub-folders?” close the search panel


phalkunz:
![silverstripe_-_files](https://user-images.githubusercontent.com/318116/27776521-67309bd4-5fff-11e7-8161-ad79fbb0bbb4.png)

## Pull requests

- [x] https://github.com/silverstripe/webpack-config/pull/7
- [x] https://github.com/silverstripe/silverstripe-admin/pull/144
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/529

phalkunz:
Ok, this turns out to be not as simple as it looks. The problem is the search panel (`Search.js`) attach a click handler to the `document` which checks the DOM element that was clicked to determine whether to close the panel (when user clicks outside of the panel) or not. However, when the user clicks / toggles the checkbox, the panel (re-) render so the checkox DOM is no longer a descendant of the panel which in turns says that the checkox is outside of the panel thus closes the panel.

PR #529, opened 2017-07-05
FIX Limit to current folder checkbox click behaviour

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/519


client/src/components/Search/Search.js
```diff
@@ -265,64 +255,66 @@ class Search extends SilverStripeComponent {
...
-            type="text"
-            name="name"
-            placeholder={i18n._t('AssetAdmin.SEARCH', 'Search')}
-            className="form-control search__content-field"
-            onKeyUp={this.handleKeyUp}
-            onChange={this.handleChange}
-            value={searchText}
-            autoFocus
-          />
+      <ClickOutComponent onClickOut={this.hide} callerComponent={this}>
```
flamerohr: bind hide at the constructor `this.hide = this.hide.bind(this);` then you can remove the `callerComponent`

PR #545, opened 2017-07-14
Add edit command and its url test for file insert

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-admin/issues/66


client/src/entwine/TinyMCE_sslink-file.js
```diff
@@ -9,16 +9,20 @@ import ShortcodeSerialiser from 'lib/ShortcodeSerialiser';
...
-  // eslint-disable-next-line no-console
-  onclick: (editor) => editor.execCommand('sslinkfile'),
-});
+TinyMCEActionRegistrar
+  .addAction('sslink', {
+    text: i18n._t('AssetAdmin.LINKLABEL_FILE', 'Link to a file'),
+    // eslint-disable-next-line no-console
+    onclick: (editor) => editor.execCommand(commandName),
+  })
+  .addCommandWithUrlTest(commandName, /\[file_link/);
```
flamerohr: I think this will need a `^` at the start of the regex

Issue #550, opened 2017-07-17
Need option to not-inline protected thumbnails


tractorcow:
For certain backends (e.g. s3) it's more performant to use signed urls than inline content, as protected thumbnails can be served quicker directly through secure s3 urls.

We need an option on FileTypeCreator::resolveThumbnailField() to invoke getURL() even on protected thumbnails.

This option still needs to be OFF by default, but it should be easy to turn on via yml config.

Part of https://github.com/silverstripe/silverstripe-assets/issues/52

# Acceptance Criteria

 - When using s3 assets thumbnails are served via direct urls rather than inline content
 - https://github.com/madmatt/silverstripe-s3/pull/7 is implemented

# Tasks

 - [x] Add config to asset admin to make thumbnail method configurable
 - [x] Ensure s3 module disables inline thumbnails
 
# PRs

 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/552
 - [x] https://github.com/madmatt/silverstripe-s3/pull/9

PR #552, opened 2017-07-18
API Abstract inlining of urls into a configurable service

tractorcow:
Parent issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/550


code/Controller/AssetAdmin.php
```diff
@@ -1083,20 +1088,31 @@ public function getObjectFromData(File $file)
...
+            $thumbnail = $this
+                ->getThumbnailGenerator()
+                ->generateThumbnail($file, $width, $height);
+
+            // Generate links if client requests them
+            // Note: Thumbnails are always generated even if links are not
+            if ($thumbnailLinks) {
+                $object['smallThumbnail'] = $this
+                    ->getThumbnailGenerator()
+                    ->generateLink($smallThumbnail);
```
flamerohr: the thumbnail generation is neat, but could the API be done such as `$smallThumbnail->generateLink()`?
I guess that would end up having an instance of thumbnailGenerator with the generated image attached, which isn't the greatest for maintenance... this is more of a passing idea.

tractorcow: $smallThumbnail is a DBFile, but this thumbnail generator is asset-admin specific, so it probably can't be added as a DBFile method.

flamerohr: I mean such that `$generator->generateThumbnail()` could return a generator wrapper instead of the DBFile and could directly call `generateLink()` or `getFile()` from that?

tractorcow: It was the roundest square I could fit into that hole.

flamerohr:
@tractorcow I've added some amendments to this pull request, I think it simplifies the API to be more direct for what it's needed for.

Issue #532, opened 2017-07-05
Uploading larger files breaks the error message


clarkepaul:
Tested with large image and large video file, with both showing the same issue.

![image](https://user-images.githubusercontent.com/555033/27886613-028ec014-6231-11e7-91d3-5f8afb635731.png)

## Related

 * [Large folder breaks GraphQL fetch](https://github.com/silverstripe/silverstripe-asset-admin/issues/536)

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/553

chillu:
Excludes sanity check for PHP config vs. SS configured value. Please chekc if this is enforced, and create a ticket if it isn't. Ideally this graceful error handling would be handled fairly centrally.

Issue #551, opened 2017-07-17
Issue: Initial load of graphql on asset admin shows white screen


tractorcow:
The white screen is not indicative of any background progress, so it appears as though the system has errored.

We need some progress indicator / animation while the initial (and ideally all) graphql queries are running.

Tasks:

 - [x] Design loading background
 - [x] Implement on initial load
 - [x] Implement on subsequent graphql queries
 - [x] Test with slow backend (e.g. s3)

tractorcow:
Note: We already have a loading overlay. Probably redundant to add design.

tractorcow:
Confirm with @clarkepaul  just use existing loading indicator.

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-asset-admin/pull/558

PR #553, opened 2017-07-20
ENHANCEMENT: More robust upload validation, more responsive UI

unclecheese:
Fixes: https://github.com/silverstripe/silverstripe-asset-admin/issues/532

## Summary of changes

The primary change in this pull request to make the uploading UI use more optimistic rendering and provide instant feedback to the user. Currently, when large files or invalid file types are uploaded, a network request is made and the error originates server-side. These changes enforce client-side upload validation, which is built into the DropzoneJS library, with backend validation as a fallback. Further, the backend `UploadValidator` instance is the source of truth for any validation that happens both client side and server side.

### Bugfixes
* Uploads that fail validation on the server no longer blow up the UI
* Files deemed invalid by the Dropzone configuration no longer proceed with a network request
* Removing a queued item no longer selects the adjacent item

### Enhancements
* Client-side validation is enforced throughout, making use of Dropzone's `accept` handler in lieu of imperative validation in `addFile`. 
* Files are optimistically added to the UI, without regard to their valid or invalid state
* Preview generation does not block files being added to the gallery view, giving the user instant feedback. Images that need previews generate them asynchronously and backfill them.
* Dropzone configuration with respect to validation is derived from the `UploadValidator` state.
* Client-side `acceptedFiles` validation greys out invalid files in dialog box

### Minor
* Promote use of `object-rest-spread`


client/src/components/AssetDropzone/tests/AssetDropzone-test.js
```diff
@@ -168,24 +161,22 @@ describe('AssetDropzone', () => {
...
-        .catch((error) => {
-          expect(error instanceof Error).toBeTruthy();
-        })
-        .then(() => done());
+      item.dropzone._errorProcessing = jest.genMockFunction();
+      item.dropzone.addFile({ name: 'test', size: 100, type: 'text/plain' });
+      // The error gets called asynchronously
+      setTimeout(() => {
+        expect(item.dropzone._errorProcessing).toBeCalled();
+      }, 0);
```
flamerohr: I'd advise wrapping this in a `return Promise()` and call the `resolve()` inside `setTimeout()`. Jest has a funny way of ignoring `expect()` calls which are not contained in the test or a returned promise chain.
is there a way to tell when `addFile` is done directly?

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -17,6 +17,7 @@ import readFilesQuery from 'state/files/readFilesQuery';
 import deleteFileMutation from 'state/files/deleteFileMutation';
 import unpublishFileMutation from 'state/files/unpublishFileMutation';
 import CONSTANTS from 'constants/index';
+import configShape from '.././../lib/configShape';
```
flamerohr: we don't need relative paths for this :)
`from 'lib/conigShape';` is enough

client/src/containers/AssetAdmin/AssetAdminRouter.js
```diff
@@ -5,6 +5,7 @@ import AssetAdmin from 'containers/AssetAdmin/AssetAdmin';
 import { decodeQuery } from 'lib/DataFormat';
 import qs from 'qs';
 import CONSTANTS from 'constants/index';
+import configShape from '../../lib/configShape';
```
flamerohr: `from 'lib/conigShape';` is enough

tests/behat/src/FixtureContext.php
```diff
@@ -167,6 +167,18 @@ public function iAttachTheFileToDropzone($path, $name)
     }
 
     /**
+     * @Then I should see an error message on the file :file
+     */
+    public function iShouldSeeAnErrorMessageOnTheFile($file)
```
flamerohr: it doesn't look like `:file`/`$file` is used?

I would suggest using the label/title on the errored gallery-item and browsing the dom for that element, instead of using `:first-child` which limits the used case of this statement to only this test.
A good guide to follow is the `I click on the file named "folder1" in the gallery`

client/src/components/AssetDropzone/AssetDropzone.js
```diff
@@ -309,57 +326,61 @@ class AssetDropzone extends SilverStripeComponent {
...
     if (typeof this.props.canFileUpload === 'function' && !this.props.canFileUpload(file)) {
-      this.dropzone.removeFile(file);
-      return Promise.resolve();
+      return done(i18n._t('AssetAdmin.DROPZONE_CANNOT_UPLOAD'));
     }
 
     if (!this.props.canUpload) {
-      this.dropzone.removeFile(file);
-      return Promise.reject(new Error(i18n._t('AssetAdmin.DROPZONE_CANNOT_UPLOAD')));
+      return done(i18n._t('AssetAdmin.DROPZONE_CANNOT_UPLOAD'));
```
unclecheese: Wow. There are no fallback translations in the entire file. Adding them.

client/src/components/AssetDropzone/AssetDropzone.js
```diff
@@ -489,11 +510,13 @@ class AssetDropzone extends SilverStripeComponent {
 
 AssetDropzone.propTypes = {
   folderId: React.PropTypes.number.isRequired,
+  handleAccept: React.PropTypes.func,
   handleAddedFile: React.PropTypes.func.isRequired,
   handleDragEnter: React.PropTypes.func,
   handleDragLeave: React.PropTypes.func,
   handleDrop: React.PropTypes.func,
   handleError: React.PropTypes.func.isRequired,
+  handlePreviewLoaded: React.PropTypes.func,
```
unclecheese: It's there. Just below `handleSuccess`

flamerohr: oh, you're right :D I was sure I didn't a text search for it as well... carry on

tests/behat/src/FixtureContext.php
unclecheese: Nice. Have made use of `getGalleryItem` to make this more reusable.

flamerohr:
Merging on the basis that the failing behat test is an existing fail (intermittent) that will be fixed by https://github.com/silverstripe/silverstripe-asset-admin/issues/551

Issue #538, opened 2017-07-07
Editor form loading blocks on history API ajax request


tractorcow:
The form is loaded, but locked and uneditable until the `admin/assets/api/history` is loaded.

It would be better if the panel didn't load at all until clicked on, and the loading was limited to that panel.

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-admin/pull/171
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/565

tractorcow:
![image](https://user-images.githubusercontent.com/936064/27939455-df2a3020-6318-11e7-8257-76325bc840b6.png)

chillu:
I see this as a high priority since it actually make the whole UI feel sluggish on a major user action.

clarkepaul:
I take  it this isn't anything for me to worry about UX wise.

chillu:
@clarkepaul said that we should move the "history" tab into a secondary action anyway, because it needs different form actions (revert). Paul, can you please link the other issue here

clarkepaul:
The moving of file history issue https://github.com/silverstripe/silverstripe-asset-admin/issues/541 which should probably happen first.

flamerohr:
This seems to be a few things stacked up to give this experience:
* a history request is made before the form data request in certain circumstances
* Apache delays the important form data request until the history request is completed (what you observe in the screenshot)
* a history request could be made on unloading the old history list, un-intentional side-effect of having a `oldFileId !== newFileId` check.

To reproduce: open a file in editor, open different file, open the first file. This is all without closing the Editor.

PR #565, opened 2017-07-26
Fix history list to load last in the stack

flamerohr:
Fix moved loading overlay to the Gallery only, rather than the entire screen

Resolves some usability issues experienced in https://github.com/silverstripe/silverstripe-asset-admin/issues/538 not sure I'd be able to address that apache delay I mentioned


tractorcow:
Does this still trigger a background loading even if you never view the history tab? If we only view that tab rarely, we could save a lot of performance lost by lazy-loading it in.

flamerohr:
Yes this still retains that behaviour...
Unfortunately, I think to achieve "show only when this component shows" is not as easy as it is said... because technically it has "shown in the DOM" but hidden by the tab.
This will require more work with the `Tabs` component and tweaking it's "unmount on hide" feature rather than working on the `HistoryList` component and could cause undesirably regressions.

So for now, I think an improved non-blocking interface is a better point of call for this.
I also think apollo-client and graphql would benefit this greatly with its query caching system.

PR #568, opened 2017-07-31
FIX Center align folder preview icons in edit panel

### LINKED ISSUE #560: Folder icon is not centred on the edit panel
pasted_image_24_07_17__4_16_pm

robbieaverill:
Fixes #560


client/src/containers/Editor/Editor.scss
```diff
@@ -70,6 +70,11 @@
   }
 }
 
+.editor__file-preview--folder {
+  margin: auto;
```
flamerohr: isn't `margin: auto;` the same as `margin: 0;`?

robbieaverill: Nope - auto will centre align

tractorcow: Yes that's right; This change seems fine to me.

flamerohr: I mean, `auto` is useless in this case, because you have `width: 100%` as well :P

flamerohr: tested this out, and I think it'd be safer to target `margin-left: 0; margin-right: 0;` individually
you're removing a `margin-bottom: .76925rem;` unintentionally from that element

robbieaverill: Yeah it does look confusing, but the margin and width are calculated as a specific pixel width.  I could use `width: inherit` instead if that'd be less confusing?

robbieaverill: @flamerohr OK cool, updated and pushed

flamerohr: yeah, I guess `inherit` would've worked too :)

PR #570, opened 2017-08-01
API Introduce thumbnail loading mechanism

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/5544

In-progress images look like this (animated)

image

Images that failed after the specified retry period look like this

image

@clarkepaul can you please review the styles of the screenshots I posted?


clarkepaul:
Is it too late to use the button loading in place of the S loading? I've had feedback that it sends bad connotations having the loading S all over the place (ect. they think our system is slow rather than their hosting). So I'd be trying not to use the S for new scenarios but rather only use it in the context of loading panels as we currently do.

client/src/state/imageLoad/ImageLoadActionHandler.js
```diff
@@ -0,0 +1,204 @@
...
+  expiry: 0,
+};
+
+class ImageLoadActionHandler {
+
+  constructor(options, action) {
+    this.onStatusChange = null;
+    this.onRetry = null;
+    this.onReset = null;
+    this.onTimeout = null;
```
flamerohr: I had set these event callbacks to the `this.options` object, so they can be set by passing in options.
This makes the setters less relevant but still available if needed.

client/src/state/imageLoad/ImageLoadActionHandler.js
```diff
@@ -0,0 +1,204 @@
...
+
+  /**
+   * Change the status of a file
+   *
+   * @param {String} url
+   * @param {String} status
+   */
+  handleStatusChange(url, status) {
+    if (this.onStatusChange) {
+      this.onStatusChange(url, status);
```
flamerohr: had deleted `handleStatusChange()` as it was just a pass-through function

client/src/state/imageLoad/ImageLoadActionHandler.js
```diff
@@ -0,0 +1,204 @@
...
+
+    // Otherwise, stop further attempts
+    return null;
+  }
+
+  /**
+   * On failure we want to temporarily mark as failed, and retry later
+   *
+   * @param {String} url
+   * @return {Promise}|{null}
```
flamerohr: valid JSDocs is `{Promise|null}`

client/src/state/imageLoad/ImageLoadActionHandler.js
```diff
@@ -0,0 +1,204 @@
...
+   * @param {String} url
+   * @param {String} status
+   */
+  handleStatusChange(url, status) {
+    if (this.onStatusChange) {
+      this.onStatusChange(url, status);
+    }
+  }
+
+  handleTimeout(callback, delay) {
```
flamerohr: missing docblock

client/src/state/imageLoad/ImageLoadActions.js
```diff
@@ -0,0 +1,58 @@
...
+import ImageLoadActionHandler from './ImageLoadActionHandler';
+
+/**
+ * Default factory to test for image
+ *
+ * @param {String} url
+ * @param {function} resolve
+ * @param {function} reject
+ */
+const defaultImageFactory = (url, resolve, reject) => {
```
flamerohr: moved this to `ImageLoadActionHandler` as a default for constructor

client/src/state/imageLoad/ImageLoadActions.js
```diff
@@ -0,0 +1,58 @@
...
+      return null;
+    }
+
+    // Only process new urls, or urls that have been reset
+    const state = getState();
+    const currentFile = state
+      && state.assetAdmin
+      && state.assetAdmin.imageLoad
+      && state.assetAdmin.imageLoad.files
+      && state.assetAdmin.imageLoad.files.find((file) => file.url === url);
```
flamerohr: don't need all those prop.exists checks, as Reducer should've populated all those before action is called.

client/src/state/imageLoad/tests/ImageLoadActions-test.js
```diff
@@ -0,0 +1,201 @@
...
+
+jest.unmock('react');
+jest.unmock('deep-freeze-strict');
+jest.unmock('../ImageLoadActionHandler');
+jest.unmock('../ImageLoadActionTypes');
+
+import ImageLoadActionHandler from '../ImageLoadActionHandler';
+import IMAGE_STATUS from '../ImageLoadStatus';
+
+describe('imageLoadActions', () => {
```
flamerohr: this is more accurately describing `ImageLoadActionHandler` so better to give that description here

client/src/state/imageLoad/tests/ImageLoadActions-test.js
```diff
@@ -0,0 +1,201 @@
+/* global jest, describe, it, expect, beforeEach, jasmine */
```
flamerohr: Might be worth having a look through the `ImageLoadActionHandler-test.js` file altogether, there were a few things changed which I didn't list out here

flamerohr:
I'd like to add some unit tests for `loadImage` in `ImageLoadActions` before merging :)

tractorcow:
Rebased and requested tests added.

PR #574, opened 2017-08-04
Enhancement add middleware to update insert button label if form is dirty

flamerohr:
Please see https://github.com/silverstripe/silverstripe-asset-admin/issues/309 for more details

Relies on this https://github.com/silverstripe/silverstripe-admin/pull/180 to have a consistent effect for the middleware


client/src/boot/index.js
```diff
@@ -19,6 +20,26 @@ document.addEventListener('DOMContentLoaded', () => {
   Injector.component.register('HistoryList', HistoryList);
   Injector.component.register('ProportionConstraintField', ProportionConstraintField);
 
+  Injector.transform(
```
unclecheese: How do you feel about moving this into `applyTansforms()` like we did in `silverstripe-admin`? Just for consistency. There will be more of these transformations coming, I'm sure.

client/src/boot/index.js
```diff
@@ -19,6 +20,26 @@ document.addEventListener('DOMContentLoaded', () => {
...
+    'insert-media-modal',
+    (updater) => {
+      updater.form.alterSchema('AssetAdmin.EditForm', form => {
+        const schema = form.getState();
+        const schemaId = schema.schema.id;
+        const overrides = schema.stateOverride && schema.stateOverride.fields;
+        const customTitle = (schemaId === 'Form_fileInsertForm' && overrides)
```
unclecheese: I'm super annoyed (perhaps on your behalf) that this has to be so imperative. I wonder if we should have `FormBuilderLoader` append the schema ID on to the identifier? So you could match `AssetAdmin.EditForm.fileInsertForm` or something. Those identifiers should be as unique as realistically possible.

client/src/boot/index.js
flamerohr: Yeah, good idea, lets do that :)

client/src/boot/index.js
flamerohr: yeah, it was annoying to need to have the schemaId... In retrospect, perhaps the `schema.name` might've been cleaner `fileInsertForm`
I like the idea of appending this to the identifier, then we could utilise selectors better!

flamerohr:
@unclecheese Rebased and applied changes

I've added a `todo` note for the schemaName/schemaId usage, and I'll create a separate story for appending schemaName to the `identifier`

PR #582, opened 2017-08-10
Enhancement add option to hide the history tab, and hide by default

flamerohr:
Relies on https://github.com/silverstripe/silverstripe-framework/pull/7275

### DIFF
--- a/code/Forms/FileFormFactory.php
+++ b/code/Forms/FileFormFactory.php
@@ -17,6 +17,13 @@ use SilverStripe\Forms\TextField;
 
 class FileFormFactory extends AssetFormFactory
 {
+    /**
+     * History tab/form to be shown to the user or not
+     *
+     * @var bool
+     */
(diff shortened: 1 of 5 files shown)


flamerohr:
Yeah, I will need to look into it on Monday, I don't think behat can configure config variables like that... or can it?

If @tractorcow can confirm, perhaps the best course of action is to remove the behat test until the history tab refactor work happens.

code/Forms/FileFormFactory.php
```diff
@@ -17,17 +17,21 @@
 
 class FileFormFactory extends AssetFormFactory
 {
+    private static $show_history = false;
```
tractorcow: Can you please phpdoc this?

flamerohr:
Behat tests fixed with https://github.com/silverstripe/silverstripe-admin/pull/189

PR #587, opened 2017-08-16
FIX misleading button labels

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-admin/issues/169


client/src/entwine/TinyMCE_ssembed.js
```diff
@@ -19,12 +19,12 @@ const filter = 'div[data-shortcode="embed"]';
     init: (editor) => {
       editor.addButton('ssembed', {
         icon: 'media',
-        title: 'Insert Embedded content',
+        title: 'Insert media via URL',
```
flamerohr: Could we move these to the `i18n` translation files? :)
It'd be great to be able to text-collect these while we update the text

phalkunz: Cool, will do :)

Issue #528, opened 2017-07-05
Misleading error message when API loading fails ("folder cannot be found")


chillu:
This shouldn't happen in the first place, but if it does we should deal with it more gracefully.

![image](https://user-images.githubusercontent.com/111025/27845557-7288d010-6184-11e7-825e-192aab81ec8f.png)

## Related 

 * [Upload can't handle large amount of files](https://github.com/silverstripe/silverstripe-asset-admin/issues/139)
 * [Implement cachable asset store](https://github.com/silverstripe/silverstripe-assets/issues/43)
 * [Large folder breaks GraphQL fetch](https://github.com/silverstripe/silverstripe-asset-admin/issues/536)
 * [Pregenerate thumbs on upload](https://github.com/silverstripe/silverstripe-asset-admin/issues/462)
 
## PRs
 
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/590
 - [x] https://github.com/silverstripe/silverstripe-admin/pull/198

clarkepaul:
What are the scenarios for when this could happen?
I've experienced this when uploading a large file, while waiting navigate away to a different file's details—the view of the files and folders on the left fails.

newleeland:
I've also had occur when saving edited files when my session has timed out.

clarkepaul:
Error appears:
* When session times out
* Reached memory capacity
   * Individual file?
   * Folder?
   * Overall files area / system?

tractorcow:
>  I've also had occur when saving edited files when my session has timed out.

I think in this case, we need to ensure the CMS re-authentication dialog works with asset admin section

mister-todd:
@flamerohr what's thisblocked by?

PR #590, opened 2017-08-18
Improve error handling of ajax errors when providing form schema

tractorcow:
Parent issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/528

Requires https://github.com/silverstripe/silverstripe-admin/pull/198


client/src/containers/Editor/Editor.js
```diff
@@ -152,14 +152,12 @@ class Editor extends Component {
 
     let error = null;
     if (this.state.loadingError) {
-      const message = this.state.loadingError.value;
+      let message = this.state.loadingError.value;
+      if (this.state.loadingError.code === 404) {
+        message = i18n._t('AssetAdmin.FILE_MISSING', 'File cannot be found');
+      }
```
flamerohr: how about one step further and adding an `Unknown error occurred` message for when `message` is empty

flamerohr: Oh, I see it on the FormBuilderLoader end... Might pay to have something on this end as well, as loadingError may be handled by something else in future

client/src/containers/Editor/Editor.js
```diff
@@ -156,6 +156,9 @@ class Editor extends Component {
       if (this.state.loadingError.code === 404) {
         message = i18n._t('AssetAdmin.FILE_MISSING', 'File cannot be found');
       }
+      if (!message) {
+        message = i18n._t('AssetAdmin.UNKNOWN_ERROR', 'An unknown error has occurred');
```
tractorcow: Shall we use the existing Admin.UNKNOWN_ERROR key?

flamerohr: yeah, could do with the explicit dependency on Admin

code/GraphQL/FileFilterInputTypeCreator.php
```diff
@@ -88,6 +89,13 @@ public function filterList(Filterable $list, $filter)
         // ID filtering
         if (isset($filter['id']) && (int)$filter['id'] > 0) {
             $list = $list->filter('ID', $filter['id']);
+            
+            if ($list->count() === 0) {
+                throw new HTTPResponse_Exception(_t(
+                    __CLASS__ . 'FileNotFound',
```
tractorcow: String being concatenated needs to have a leading .

```diff
- __CLASS__ . 'FileNotFound',
+ __CLASS__ . '.FileNotFound',
`

File: client/src/state/files/readFilesQuery.js [hunk 4]
```diff
@@ -113,11 +114,16 @@ const config = {
 
     const filesLoading = (folder && !folder.children);
 
+    const errors = error && error.graphQLErrors &&
+      error.graphQLErrors.map((graphQLError) => (
+        `${graphQLError.message} ${JSON.stringify(graphQLError.locations)}`
```
tractorcow: The `.locations` component doesn't seem to be useful to the user. Shall we omit it?

flamerohr: hmm yeah, that locations data seems to be quite useless for devs as well (at least until it stops masking everything)

Issue #578, opened 2017-08-07
Bulk move of files


clarkepaul:
## AC's
As a user, I should:
* have a way to select multiple files (already built)
* be presented with an action to move multiple items which are intuitive
* have the ability to select multiple files to move collectively to a new folder location at any depth of the hierarchy of my choosing in the Files area
* be provided with a good indication of where the files have been moved
* be provided with a similar interaction as the moving of a single file https://github.com/silverstripe/silverstripe-asset-admin/issues/576

Out of scope: "Undo"

## Tasks
* [x] New batch action for bulk move
* [x] New dialog for "select destination" (treedropdown)
* [x] Display message (success or failure)
* [x] Message above the dropdown field in the modal
* [x] Behat tests
* [x] Add "Add folder" svg to fontastic
* [x] Update CMS with new font icons (May be completed with #576 ) https://github.com/silverstripe/silverstripe-admin/pull/190
* [x] i18n

## PRs

- [x] https://github.com/silverstripe/silverstripe-admin/pull/190
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/592
- [x] https://github.com/silverstripe/silverstripe-admin/pull/195

## Prototype
Move for files: https://invis.io/AKCZSFOWX#/247323908_Files_-_Edit_File_-_Move_Multi_File
Move for folders: https://invis.io/YRCZSG34B#/247323905_Files_-_Edit_File_-_Move_Multi_Folder

**Related**
https://github.com/silverstripe/silverstripe-asset-admin/issues/576

tractorcow:
Updated https://github.com/silverstripe/silverstripe-admin/pull/190

PR #592, opened 2017-08-20
FEATURE: Bulk move files

unclecheese:
~~Behat tests forthcoming.~~

Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/578


tests/behat/features/manage-files.feature
```diff
@@ -98,6 +98,33 @@ Feature: Manage files
...
+    Given I click "Move" in the "#Form_moveForm_action_move" element
+    Then I should not see the file named "file1" in the gallery
+      And I should not see the file named "file2" in the gallery
+      And I should not see the file named "testfile" in the gallery
+    When I press the "Navigate up a level" button
+      And I click on the file named "folder2" in the gallery
+    Then I should see the file named "file1" in the gallery
+      And I should see the file named "file2" in the gallery
+      And I should see the file named "testfile" in the gallery
+
```
unclecheese: Ideally, this should test that a folder is disabled when included in the list of selected items. I started down the path of behat testing that in `TreeDropdownField` and wasn't able to make much progress outside of really ugly solutions.

client/src/boot/applyTransform.js
```diff
@@ -21,11 +20,47 @@ const insertTransform = form => {
...
     (updater) => {
-      updater.form.alterSchema('AssetAdmin.EditForm', insertTransform);
+      updater.form.alterSchema('AssetAdmin.EditForm.fileInsertForm', insertTransform);
+    }
+  );
+
+  Injector.transform(
+    'move-form-disabled',
+    (updater) => {
+      updater.component('TreeDropdownField.AssetAdmin.MoveForm', ConnectedMoveTreeDropdownField);
```
flamerohr: I think this is a really neat way of achieving our goal, and demonstrates the power that the Injector API provides on the javascript side :D

client/src/containers/MoveModal/MoveModal.js
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+import i18n from 'i18n';
+import CONSTANTS from 'constants/index';
+import { deactivateModal, setNoticeMessage } from '../../state/gallery/GalleryActions';
```
flamerohr: can use absolute paths `state/gallery/GalleryActions;`
this is because `client/src` (and a couple of other directories) are setup as module folders, so we do not get into `../` relative-ancestor-hell

client/src/containers/MoveModal/MoveModal.js
```diff
@@ -0,0 +1,112 @@
...
+import FormBuilderModal from 'components/FormBuilderModal/FormBuilderModal';
+import configShape from 'lib/configShape';
+import moveFilesMutation from '../../state/files/moveFilesMutation';
```
flamerohr: absolute paths here would be good

client/src/containers/MoveModal/MoveModal.js
```diff
@@ -0,0 +1,112 @@
...
+                i18n._t('AssetAdmin.MOVED_ITEMS_TO', 'Moved %s item(s) to ')
+                , selectedFiles.length
+              )}
+              <a href="#" onClick={goToFolder}>{filename}</a>
+            </span>
+          ),
+        });
+        setTimeout(() => setNotice(null), 3000);
+        hide();
+      });
```
flamerohr: does this need a `.catch()` in case the move fails?

client/src/containers/MoveModal/MoveModal.js
```diff
@@ -0,0 +1,112 @@
...
+            <span>
+              {i18n.sprintf(
+                i18n._t('AssetAdmin.MOVED_ITEMS_TO', 'Moved %s item(s) to ')
+                , selectedFiles.length
+              )}
+              <a href="#" onClick={goToFolder}>{filename}</a>
+            </span>
+          ),
+        });
+        setTimeout(() => setNotice(null), 3000);
```
flamerohr: move the `3000` to a constant at the top of the file, or ideally the constants/index.js file.

I'd recommend storing the timer id for the `setTimeout()` on the object `this.noticeTimer = setTimeout(...)`, so you can `clearTimeout(this.noticeTimer);` in the `goToFolder()` above

client/src/containers/MoveModal/MoveModal.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+function mapDispatchToProps(dispatch) {
+  return {
+    hide() {
+      dispatch(deactivateModal());
+    },
+    setNotice(msg) {
+      dispatch(setNoticeMessage(msg));
+    },
```
flamerohr: for the sake of standardising, do you want to go with this route from now on?
traditionally, we've been keeping all dispatch actions (and graphql queries/mutations) in the actions object:
```
function mapDispatchToProps(dispatch) {
  return {
    actions: {
      gallery: bindActionCreators(galleryActions, dispatch),
    },
  };
}
```

tests/behat/features/manage-files.feature
flamerohr: I'll try things out tomorrow, I think we could have a clean way to test disabled items but I need to confirm that it works first

unclecheese: The cleanest solution would be to add such a step to `CMSFormContext`

tests/behat/features/manage-files.feature
```diff
@@ -98,6 +98,33 @@ Feature: Manage files
...
+    Given a "image" "assets/folder1/file2.jpg" was created "2012-01-02 12:00:00"
+      And a "folder" "assets/folder1/folder3"
+    When I click on the file named "folder1" in the gallery
+    And I check the file named "file1" in the gallery
+    And I check the file named "file2" in the gallery
+    Then I should see an ".bulk-actions__action[value='move']" element
+    And the ".bulk-actions-counter" element should contain "2"
+    And the ".bulk-actions__action[value='move']" element should contain "Move"
+    When I attach the file "testfile.jpg" to dropzone "gallery-container"
+    And I check the file named "testfile" in the gallery
```
unclecheese: Yeah, fair point. I copied some of this from the bulk delete files. I assume deletion of an incomplete upload should work exactly the same as an uploaded one, at least from a UI perspective. Not sure the same is true about moving. You shouldn't really be able to move an incomplete upload.

client/src/boot/applyTransform.js
```diff
@@ -21,11 +20,47 @@ const insertTransform = form => {
...
+const DisabledTreeDropdownField = (TreeDropdownField) => (props) => {
+  // eslint-disable-next-line react/prop-types
+  const { disabledIDs } = props;
+  const newProps = {
+    ...props,
+    findTreeByPath(tree, visible) {
+      const visibleTree = findTreeByPath(tree, visible);
+      return visibleTree ? {
+        ...visibleTree,
+        disabled: disabledIDs.includes(visibleTree.id),
```
unclecheese: Thanks! Hopefully you can see why I was so fixated on getting it right. It is an elegant solution, but it relies on improvements to the TreeDropdownField API, and highlights some limitations there.

client/src/boot/applyTransform.js
unclecheese: Thanks! Hopefully you can see why I was so fixated on getting it right. It is an elegant solution, but it relies on improvements to the TreeDropdownField API, and highlights some limitations there.

client/src/containers/MoveModal/MoveModal.js
unclecheese: I really don't like that practice. I think it encourages laziness, and gives your components too wide a scope. I think you should have to whitelist what actions your components have access to. Plus, there are a lot of cases where you actually want to do something in your action method beyond just invoking `dispatch()`, and in that sense it can clean up a lot of code in your component.

So, yes would be my vote, but that would be a massive cleanup effort to apply that standard. Good hackday project, maybe.

client/src/boot/applyTransform.js
flamerohr: Yeah, and the TreeDropdownField API additions were expected to a degree :)
I think we'll discover more of these as everyone continues finding more usecases

client/src/containers/MoveModal/MoveModal.js
flamerohr: fair points, so many hackday projects in our list already ;)

Maybe create a new story for this, it can go alongside the "simplify and cleanup Components to dumb and smart Components" stories

flamerohr: Oh, with regards to the `hide` property function, I guess I'd prefer the callback `onHide` for this case.
A bit more descriptive and less obscure purpose

client/src/containers/MoveModal/MoveModal.js
unclecheese: Ah, cool, didn't know those paths were resolved in webpack.

client/src/containers/MoveModal/MoveModal.js
unclecheese: Probably should. We're not doing it in `Gallery.handleMoveFiles`, so I'll add something there too.

client/src/containers/MoveModal/MoveModal.js
unclecheese: True. Add that to the list, too. Our `handleXXX` and `onXXX` semantics are all over the place.

client/src/boot/applyTransform.js
unclecheese: Good catch. Done.

client/src/boot/applyTransform.js
```diff
@@ -21,11 +20,48 @@ const insertTransform = form => {
   return form.getState();
 };
 
+const DisabledTreeDropdownField = (TreeDropdownField) => (props) => {
+  // eslint-disable-next-line react/prop-types
+  const { disabledIDs } = props;
+  const newProps = {
+    ...props,
+    findTreeByPath(tree, visible) {
+      const visibleTree = findTreeByPath(tree, visible);
```
flamerohr: with `findTreeByPath()` being an extendable prop, perhaps it'd be a good idea to check props for a previously injected `findTreeByPath()` function to use as well before using the library `findTreeByPath()`

Issue #563, opened 2017-07-24
Not all files can be placed into a WYSIWYG area (only images in SS4)


clarkepaul:
User can't place files of the type docs, audio, video into a WYSIWYG area (If they try they get an error). In SS3 you can select any file into the text editor (files other than images are placed as links) but with SS4 you can only place images. 

If we don't match SS3 functionality we need to provide clarity that certain files can't be placed by either not making them available or disabling the insert functionality (with notification this file type can not be inserted) 

## AC's
As a CMS user, I would like to:
- be notified when a particular file type is cannot be inserted onto a page.
- have the ability to insert an unsupported file as a link.
- clearly understand the difference between inserting content from "Files" and embedding content from offsite. (Related to https://github.com/silverstripe/silverstripe-admin/issues/169)

## Tasks
- [x] Add notification to side panel when selected file cannot be properly inserted
- [x] Enable files the are currently not supported to be inserted as links (as per design below)
- [x] Replace existing content under "Placement" tab with "Link text", "Filename" and "URL" read only fields.
- [x] Change text of "insert file" button to "Insert link"
- [x] Style notification message

## Prototype:
https://invis.io/4NCYDMWEW#/247738949_Page_-_TinyMCE_-_Insert_Media_-_Home

## Pull request:

- ~https://github.com/silverstripe/silverstripe-asset-admin/pull/585~
- [x] https://github.com/silverstripe/silverstripe-admin/pull/204
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/596

tractorcow:
You can add a link to a file. I know because I just did it. :)

![image](https://user-images.githubusercontent.com/936064/28565907-b46e1b18-7181-11e7-8dad-51561404de3b.png)

![image](https://user-images.githubusercontent.com/936064/28565925-be6eaed4-7181-11e7-821a-1fc8dae80cce.png)

tractorcow:
ok, so if you use "insert media" with an image file you get this error.

![image](https://user-images.githubusercontent.com/936064/28595427-ee1ae7d6-71e7-11e7-9ca3-d926413c4783.png)

Should the point of this issue be to address / fix this?

clarkepaul:
I'll add a bit more context above ^

tractorcow:
The above line is still incorrect.

> In SS3 you can place any file into the text editor (placed as links) but with SS4 you can only place images.

[...]

PR #596, opened 2017-08-22
Insert unsupported media as normal link

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/563


code/Forms/FileFormFactory.php
```diff
@@ -37,8 +37,18 @@ protected function getFormFieldTabs($record, $context = [])
         if ($this->config()->get('show_history')) {
             $tabs->push($this->getFormFieldHistoryTab($record, $context));
         }
+
+        $type = $this->getFormType($context);
+
+        // Unsupported media insertion will use insert link form instead
+        if ($this->getFormType($context) === static::TYPE_INSERT_MEDIA) {
```
flamerohr: could use `$type` here right?

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -329,12 +336,15 @@ jQuery.entwine('ss', ($) => {
...
-        'bad');
+      const data = this.getData();
+      const shortcode = ShortcodeSerialiser.serialise({
+        name: 'file_link',
+        properties: { id: data.ID },
+      }, true);
 
-      return false;
+      const linkText = data.Text || data.filename;
+      this.insertLinkInEditor({ href: shortcode }, linkText);
```
flamerohr: Please refer to the `sslink-file.js` file, I think you're missing other attributes that should be included for the link as well (the fields are there)
```
      // Add anchor
      const anchor = data.Anchor && data.Anchor.length ? `#${data.Anchor}` : '';
      const href = `${shortcode}${anchor}`;

      return {
        href,
        target: data.TargetBlank ? '_blank' : '',
        title: data.Description,
      };
```

PR #597, opened 2017-08-23
ENHANCEMENT Disclose file dependencies when deleting a single file

### LINKED ISSUE #591: Ask for confirmation before removing a file that's still used
# Overview

When a file is deleted from the files area, pages which use the file still think the file is available and placed on the page leaving broken images.

Example of deleted image on published page:
image
image

# Acceptance Criteria

 * When a file is deleted, show a generic warning if it is used on one or more pages (doesn't need to contain counts). Suggested message "Ensure files are removed from content areas prior to deleting them, otherwise they will appear as broken links."
 * The warning message is consistent for single file deletion from the file edit panel and bulk actions  
 * There is a warning that files might be in use when a user wants to delete multiple files (similar wording to single file delete)
 * Allow deletion after confirmation
 * ~~Direct the user to the existing "usage" tab on the "edit file" view for more details~~
 * ~~This logic applies to all UI triggers where a file can be deleted ("insert media", assets section)~~
 * The logic applies to batch operations
 * When an image is not found the image fallback text should be "File not found" not "Page not found" (as pictured above).

# Notes

 * The card doesn't include determining file usage across records. Usage is defined in the same way as shown on the "Usage" tab (which currently doesn't exist).
 * An alternative approach would be to allow users to delete but also remove all connections to those files so that we don't have broken images on pages
 
 ## PRs
 
  - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/597

# Related

 * [Show usage of files in "edit file" screen](https://github.com/silverstripe/silverstripe-asset-admin/issues/593)

tractorcow:
FIxes #591


clarkepaul:
Just made a request on the original issue, see if it's reasonable or not? Cheers

tractorcow:
@clarkepaul updated with the new confirmation message. I've also added the same message to the bulk delete warning.

client/src/containers/Editor/Editor.js
```diff
@@ -40,16 +41,33 @@ class Editor extends Component {
...
         this.props.onUnpublish(data.ID);
       }
       event.preventDefault();
       return;
     }
 
     if (name === 'action_delete') {
+      // Customise message based on usage
+      let message = i18n._t('AssetAdmin.CONFIRMDELETE', 'Are you sure you want to delete this record?');
+      if (this.props.file && this.props.file.inUseCount && this.props.file.inUseCount > 0) {
```
flamerohr: don't need `this.props.file.inUseCount` exists check and `> 0` as well, one or the other is enough to get the same results :)

flamerohr:
Broken behat tests due to the change in wording

tractorcow:
Pushed up with fixes to behat and adjusted the conditional.

PR #601, opened 2017-08-24
Fix move toolbar outside AssetDropzone

flamerohr:
This is because there is a race condition when the upload button is inside the AssetDropzone


tractorcow:
Tested and it seems ok.

PR #603, opened 2017-08-25
FEATURE: Group mouse selection

unclecheese:
Adds group selection of items using the mouse (lasso). Hold meta key to concatenate selection.

Resolves: https://github.com/silverstripe/silverstripe-framework/issues/7313


client/src/containers/Gallery/Gallery.js
```diff
@@ -368,6 +379,14 @@ class Gallery extends Component {
   }
 
   /**
+   * Toggle concatenating selected items based on the key event
+   * @param e
+   */
+  toggleSelectConcat(e) {
+    this.props.actions.gallery.setConcatenateSelect(!!e.metaKey);
```
flamerohr: `e.metaKey` returns a boolean anyway, so we don't need the casting right?

I think you'll need to add a `e.metaKey || e.ctrlKey` check, for mac and windows differences.

client/src/state/gallery/GalleryActions.js
```diff
@@ -32,6 +32,18 @@ export function selectFiles(ids = null) {
 }
 
 /**
+ * Toggles whether group selection is additive o
```
flamerohr: broken sentence in description

client/src/containers/Gallery/Gallery.js
```diff
@@ -820,7 +858,16 @@ class Gallery extends Component {
       sectionConfig: this.props.sectionConfig,
     };
 
-    return <GalleryView {...props} />;
+    return (this.props.view === 'table')
+      ? <TableView {...props} />
+      : <SelectableGroup
```
flamerohr: I think this failed linting, need to wrap in `( )` and indent correctly

client/src/state/gallery/GalleryActions.js
```diff
@@ -32,6 +32,18 @@ export function selectFiles(ids = null) {
...
+ * @param bool
+ * @returns {function(*): *}
```
flamerohr: Docblock:
```
@param {boolean} concat
@returns {function}
```

I don't think it needs the casting for the dispatch call either

client/src/containers/Gallery/Gallery.js
unclecheese: My `yarn lint` is coming back empty, and the travis tests have passed. Are you sure about this one?

flamerohr: strange
it is practiced in the majority of the code base too...

for reference, this is what I'm talking about https://github.com/airbnb/javascript/tree/master/react#parentheses

flamerohr:
After testing this out, two quirks that I'm looking to fix:
- `mouseup` only works if done inside the selectable area, kind of weird behaviour when even a pixel outside the selectable area.
- strange padding between the toolbar and gallery area, will look at decreasing the bottom padding on the toolbar and increasing the padding above the gallery area

flamerohr:
Added commit and would be good to merge either https://github.com/unclecheese/react-selectable/pull/38 or https://github.com/unclecheese/react-selectable/pull/33 to fix the mouseup and mousemove quirk

flamerohr:
Good to merge on green :)

Issue #483, opened 2017-05-19
Existing files in UploadField should allow opening of edit view


feejin:
When uploading an image for example, SilverStripe 3 has an "edit" option for inline editing of the file attributes. This is an important feature for accessibility, SEO etc. I also rely on it heavily for extensions such as focus point. Please bring it back in beta 2.

## Pull requests:
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/607

sminnee:
Have you seen this on SS4 or SS3.6?

feejin:
This was on a SilverStripe 4 dev-master branch from last week that @kinglozzer set up for internal testing. 
![screen shot 2017-05-22 at 09 28 19](https://cloud.githubusercontent.com/assets/1774129/26298986/11ae7806-3ed1-11e7-81c4-864821cf75b0.png)

chillu:
At the very least, we need to link to the edit view in `admin/assets` here. At the moment, you can't even tell which folder the file is in.

I agree that in-context editing of attached files was a strong point of the old upload field. It was also adding a lot of complexity. We should probably try to reuse the existing React UI components as much as possible for this.

@clarkepaul Can you please have a think on how you'd see this working?

clarkepaul:
Upload fields were designed to be selectable (opens the Files area with item selected in a modal) see https://projects.invisionapp.com/boards/VGX68PKC7PYB/ 
I don't recal the implementation of this so I'm not sure if it has broken since developemnt or wasn't included in the acceptence criteria. 

Outstanding designs (which I have noted):
- If a user doesn't have view/edit permissions of a file
- If a user doesn't have permissions to use/see the files area at all
- Error/success messages

tractorcow:
@clarkepaul at the moment the insert / select media modal has this form:

![image](https://user-images.githubusercontent.com/936064/29647207-288a3330-88dc-11e7-9a48-46e0fc71ab33.png)

clarkepaul:
@tractorcow Clicking an uploadfield item opens it in the Files area modal as you have pictured, nothing different really—same experience as far as I can see. Although I didn't realise we open the file in a new tab/window when editing the original but that is a separate issue and probably makes sense to leave as is until we have the "Usage" tab back again where people can link back to where it's used.

clarkepaul:
Ah yeah and don't forget hover states on the uploadfield item :)

chillu:
[...]

PR #607, opened 2017-08-30
Add edit to UploadField

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/483


client/src/components/UploadField/UploadField.js
```diff
@@ -113,11 +119,40 @@ class UploadField extends SilverStripeComponent {
...
   }
 
   /**
+   * Handler for editing an uploaded item
+   *
+   * @param {Object} event
+   * @param {Object} item
+   */
+  handleItemEdit(event, item) {
+    window.location.href = `/admin/assets/show/${item.parent.id}/edit/${item.id}`;
```
unclecheese: There has to be a better way to do this. We have `File::CMSEditLink()` but it's not available to `UploadField`. Not sure if it's worth adding.

flamerohr: Have you looked adding it to the uploadfield schema state props by updating `UploadField::getEncodedItems()`?

client/src/components/UploadField/UploadFieldItem.js
```diff
@@ -277,10 +328,12 @@ class UploadFieldItem extends SilverStripeComponent {
           ref="thumbnail"
           className={this.getThumbnailClassNames()}
           style={this.getThumbnailStyles()}
-        ></div>
+          onClick={this.handleItemClick}
+        />
         {this.renderFileDetails()}
         {this.renderProgressBar()}
         {this.renderErrorMessage()}
+        {this.renderEditButton()}
```
flamerohr: Probably weren't any mocks for this, but from a UX perspective, it would be good to also have a "ViewButton" call to action
especially on mobile, it is not obvious that you can just view without editing

@clarkepaul @newleeland what do you guys think?
I would almost lean for removing the edit button and just have the view button, the view form also has a "Edit original file" button/link anyway, although it'll be an extra click to get to.

clarkepaul: Are you sure the preview has an edit link still in this context, I can't remember seeing it? If so yes I'd agree then we can remove the edit link.

My intention wasn't to have any icons visible (you just click the bar, a pattern which will be become more standard as things progress), I didn't want to grow this issue but if we were to have multiple actions on a single item then we would have a [...] action which leads to a dropdown with multiple actions. I would like to keep it as much inline with where we intend it to be—a single action to view and other to remove (hence I didn't think the dropdown wouldn't make sense here).

flamerohr: Screenshot of the feature :)
To "view" you click the preview image/icon
and the edit/close buttons on the right next to each other
screen shot 2017-08-31 at 2 11 17 pm

client/src/components/UploadField/UploadField.js
unclecheese: I did.. are you sure that's the best place for it? I was wondering about just adding it in `AssetAdmin::getObjectFromData()`

flamerohr: I don't think `AssetAdmin::getObjectFromData()` is appropriate, as this is specific to `UploadField` currently
Maybe we can re-visit moving it there when there are more use cases for it?

client/src/components/UploadField/UploadFieldItem.js
unclecheese: Have removed the item click and replaced the edit icon with "view", which does what the item click used to.

PR #613, opened 2017-09-05
BUG Fix incorrect search form state loading

### LINKED ISSUE #612: Asset Search - 'Limit to current folder and its sub-folders?' function doesn't work.
When utilizing the 'Search feature' in the Asset Admin, enabling the "Limit to current folder and its sub-folders? - Checkbox" under 'Advanced', doesn't work as the search results show everything in the asset library that is related to the 'Search Parameter'.

Example -  
1. Create two Root Folders (Folder A & Folder B) in Asset Admin.
2. Create another folder within Folder A, such as 'Folder A2'
3. Create another folder within Folder B, such as 'Folder B2'
4. Upload files with a common name/search parameter to 'Folders A2 & B2' - e.g. Merrytest-640x420.jpg
5. Go into 'Folder B' and with the "Limit to current folder and its sub-folders?" feature enabled, search for '640x420'
6. The 'Search Results' display all the '640x420' assets from both 'Folder A2' & 'Folder B2'.

SS4 Build Version - master.2839

tractorcow:
Fixes #612


phalkunz:
I can see the limit to current dir option passed to the URL which is good but it doesn't seem to restrict the search just yet.

silverstripe_-_files

tractorcow:
I've fixed the incorrect search description text.

The issue is that if you search "all files" it'll redirect you to the root folder for the search results. After that, clicking "limit to current folder" is redundant because you aren't in the folder anymore. Separate issue IMO.

client/src/components/Search/Search.js
```diff
@@ -332,10 +333,12 @@ Search.propTypes = {
...
-function mapStateToProps(state) {
-  const formState = getFormState(state);
-  const form = getIn(formState, identifier);
-
+function mapStateToProps(state, ownProps) {
+  const formIdentifier = [
+    identifier,
+    ownProps.searchFormSchemaName,
+  ].join('.');
+  const form = getIn(getFormState(state), formIdentifier);
```
flamerohr: please check `CampaignAdmin => mapStateToProps()` for a more future proof way to fix the identifier... although it is quite verbose, so I'd welcome a helper function created:
```
    const schemaUrl = `${sectionConfig.form.campaignEditForm.schemaUrl}/${ownProps.params.id}`;
    const schema = state.form.formSchemas[schemaUrl];
    const schemaName = schema && schema.name;
    const selector = schemaName && formValueSelector(schema.name, getFormState);
    title = selector && selector(state, 'Name');
```

PR #619, opened 2017-09-07
BUG Fix display of oembed modal in some browsers

tractorcow:
Requires https://github.com/silverstripe/silverstripe-admin/pull/227


tractorcow:
Some of the travis issues will be fixed with https://github.com/silverstripe/silverstripe-asset-admin/pull/622

PR #637, opened 2017-09-19
FIX Update translations from "only these people" to "only these groups" to match CMS

robbieaverill:
We updated "only these people" to "only these groups" in the CMS and siteconfig with PRs like https://github.com/silverstripe/silverstripe-cms/pull/1880, this updates that as well.


tractorcow:
@flamerohr can you look at these npm test issues?

flamerohr:
@tractorcow just needed a restart ;)

Issue #620, opened 2017-09-07
BUG Pressing Enter key on the editor brings a file dialog box instead of submitting


phalkunz:
This issue is reproducible on Chrome, Firefox, and Safari. Surprisingly, it works as expected on IE 11.

Steps to reproduce:

1. Go to Files sections
2. Select a file
3. Change the title
4. While the cursor still in the input, press Enter key
5. Saving/form submitting should be triggered but instead, the browser brings up the file dialog box

## Pull request

- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/647

flamerohr:
Can we get a bit more detail on how to reproduce this?
Is this right after opening the file editor? After editing some fields? Any particular fields?

phalkunz:
@flamerohr I've updated the issue with steps to reproduce the issue

PR #647, opened 2017-09-29
FIX prevent opening file dialog when pressing enter key

phalkunz:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/620


client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -300,6 +300,7 @@ class PreviewImageField extends Component {
             id="preview-replace-button"
             onClick={this.preventDefault}
             className={this.getButtonClasses('replace')}
+            type="button"
```
flamerohr: maybe we should add the `type` to the button below as well, just in case it comes up in future?

PR #648, opened 2017-09-29
ENHANCEMENT Use new conditional thumbnail generation API

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/6457

Requires new API in https://github.com/silverstripe/silverstripe-assets/pull/74


_config/config.yml
```diff
@@ -8,6 +8,17 @@ SilverStripe\Admin\LeftAndMain:
   extensions:
     - SilverStripe\AssetAdmin\Controller\AssetAdminFieldsExtension
 ---
+Name: assetadminthumbnails
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\AssetAdmin\Model\ThumbnailGenerator.assetadmin:
+    class: SilverStripe\AssetAdmin\Model\ThumbnailGenerator
+    properties:
+      Generates: true
```
flamerohr: would this `generates` be case sensitive on some systems?

flamerohr:
I've added a commit to give a better look-and-feel to items which are missing thumbnails

_config/config.yml
tractorcow: No it'll be case insensitive on all; But just in case, it is the correct case (matches the `getGenerates` case).

PR #654, opened 2017-10-10
BUG fix selecting the same item multiple items in our redux store

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7459


tractorcow:
Approved can be merged. :)

PR #656, opened 2017-10-10
BUG dragging count indicator used wrong component

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-versioned/issues/56


tractorcow:
Rebased.

PR #658, opened 2017-10-10
BUG Fix to prevent folders being dragged onto itself

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-versioned/issues/57


tractorcow:
Rebased. :)

Issue #632, opened 2017-09-15
Sort performance is extremely bad if used repeatidly


clarkepaul:
Sort is pretty quick the first time it is used but slows down each time it is used. After about the tenth time it is practically unusable.

![image](https://user-images.githubusercontent.com/555033/30462526-4fd0d706-9a1a-11e7-8422-560f679602e4.png)

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/665

flamerohr:
For some reason the action `"PURGE_UPLOAD_QUEUE"` is being called an exponential number of times for each time it is sorted.

![Screen Shot 2017-10-16 at 9.01.04 AM.png](https://images.zenhubusercontent.com/575f920dbf93871ccd9c8fe1/736c4540-14c5-4756-a424-9144329ec464)

PR #665, opened 2017-10-15
BUG Fix sorting firing event multiple times

flamerohr:
also prevent activate from triggering when in uploading state

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/632


tractorcow:
Existing issue is unrelated. Will address.

PR #666, opened 2017-10-15
Experiments/listen to my story

flamerohr:
Requires https://github.com/silverstripe/webpack-config/pull/10 and https://github.com/silverstripe/silverstripe-admin/pull/307

This does not implement a separate storybook command - it relies on the storybook command in `silverstripe-admin`


flamerohr:
Behat broken for unrelated reasons, pulling out fixing to another pull request so this doesn't sit here

PR #667, opened 2017-10-15
BUG Add flag to check if record has been modified

flamerohr:
Required for https://github.com/silverstripe/silverstripe-admin/pull/308


tractorcow:
behat failure is pre-existing.

Issue #598, opened 2017-08-23
UploadField ignores Description


adrexia:
It seems like its no longer possible to provide a description field for a file upload. EG:

`$upload->setDescription('This is ignored');`

However the old 'setRightTitle', seems to work (though the alignment is a little off):

`$upload->setRightTitle('This is not ignored');`

![righhttitle](https://user-images.githubusercontent.com/984753/29640800-a7f46f28-88b4-11e7-90b6-4d1d2e0af22f.png)

tractorcow:
Fixed with #670

Issue #432, opened 2017-04-18
UploadField changes not triggering unsaved page navigation warning


tractorcow:
From #428

If modifying an uploadfield, and attempting to navigate to another page (entwine section) the "you have unsaved changes" notification isn't triggered.

PRs:

 - [ ] https://github.com/silverstripe/silverstripe-admin/pull/108
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/502

phalkunz:
@tractorcow  This is the cleanest I can come up without modifying `jquery.changetracker`.

```diff
diff --git a/client/src/components/UploadField/UploadField.js b/client/src/components/UploadField/UploadField.js
index 5fc31c2..db8725c 100644
--- a/client/src/components/UploadField/UploadField.js
+++ b/client/src/components/UploadField/UploadField.js
@@ -43,6 +43,10 @@ class UploadField extends SilverStripeComponent {
 
     if (filesChanged) {
       this.handleChange(nextProps);
+
+      if(typeof this.props.onChange === 'function') {
+        this.props.onChange(nextProps);
+      }
     }
   }
 
diff --git a/client/src/entwine/UploadField/UploadFieldEntwine.js b/client/src/entwine/UploadField/UploadFieldEntwine.js
index c299dee..e795aa0 100644
--- a/client/src/entwine/UploadField/UploadFieldEntwine.js
+++ b/client/src/entwine/UploadField/UploadFieldEntwine.js
@@ -32,12 +32,19 @@ jQuery.entwine('ss', ($) => {
       const store = window.ss.store;
       const client = window.ss.apolloClient;
       const props = this.getAttributes();
+      const $mock = jQuery(`<input type="hidden" id="mock-${props.id}" />`);
+      $(this.parents('form')).append($mock);
+      props.onChange = () => {
+        $mock.val(new Date()).trigger('change');
+      };
       ReactDOM.render(
         <ApolloProvider store={store} client={client}>
           <InjectableUploadField {...props} />
         </ApolloProvider>,
         this.parent()[0]
       );
+
+      $('.cms-edit-form')._setupChangeTracker();
     },
 
     /**
```

tractorcow:
Done!

PR #670, opened 2017-10-24
Fix apply non-destructive polyfill for the react version of UploadField

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/598


tractorcow:
Reference issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/432

Just making sure we don't re-introduce this bug :)

tractorcow:
Good job.

tractorcow:
Merge on green anyone please.

PR #673, opened 2017-10-25
Fix wrong mouse cursor for description text in upload field area

flamerohr:
Also added fix for overflowing error message in a gallery item


client/src/components/UploadField/UploadField.scss
```diff
@@ -62,8 +62,9 @@
   transform: translate(-50%, -50%);
   display: inline-block;
   text-align: center;
+  cursor: pointer;
```
tractorcow: droptext doesn't actually respond to clicks (only on parent, or on the individual a / button) so I would probably remove this line.

PR #685, opened 2017-11-06
BUG Fix buttons in upload field to be proper button types

flamerohr:
Made the button type explicit, it's type submit by default which is undesirable


unclecheese:
@flamerohr Behat tests fails on this. Want to have a look?

PR #690, opened 2017-11-07
Swapped out old variables for newer ones

clarkepaul:
Best efforts to replace older variables with newer ones without major visual differences. In cases where a variable was only used once I’ve put value directly in the stylesheet.
Requires https://github.com/silverstripe/silverstripe-admin/pull/346
Related issue https://github.com/silverstripe/silverstripe-admin/issues/201


dhensby:
I'm not sure inlining values when a variable is only used once is a good change.

1. Variables add a bit of extra context and remove the "magic number" feel when looking at the code
2. The variables may be used by 3rd parties to maintain consistent style / wanting to amend our core styles

clarkepaul:
Regarding removing the variable, I agree that this isn't ideally the best but there are loads of places we should be using variables which we aren't doing. For notifications styles there are three versions of styles for them and updating them is a larger piece of work which I am trying to avoid bloating this PR. There are also replacement shadow variables but its hard to test when I can't see how the changes impact the CMS. If I leave old variables there people will continue to use them but i would think the majority are not being used. All of the main CWP modules are going to be reviewed again to make sure they work with the latest release so these can be identified and updated to the newer variables. 

I realise things will break for someone somewhere but there must be a point we say goodbye to the old. I could try and find some replacements but that won't fix peoples issues with removed variables anyway. I don't really want to have all the old variables reference the new as the amount of variables is out of control (much more culling still to come I hope).

dhensby:
OK - I can understand that and I agree we should remove "old" (deprecated?) variables. But removing variables and moving to fixed numbers seems a step back.

If that's not something you want to address in this PR, that's ok and we can go from here.

clarkepaul:
I can give it another go and see what I can do (might take a few days to get around to it as only doing this as OS and not work). With the notification variables, it's going to be a bit harder as bootstrap uses a mixin of the base colours which would be hard to apply to our older notifications.

flamerohr:
Merging, broken behat tests are due to another pull request merged earlier today.

PR #700, opened 2017-12-01
Update version dependencies for 1.0 branch

tractorcow:


flamerohr:
Looks like selenium still isn't working :(

tractorcow:
Given we have at least fixed the phpunit tests, I think we can merge this, and fix behat in 4.1 @flamerohr

flamerohr:
We'll fix behat for 4.1 - or if it can be backported, that would be great too

Issue #683, opened 2017-11-02
Lost setAllowedMaxFileNumber()


zanderwar:
# Overview

UploadField still allows multiple uploads, but for some reason the ability to limit the amount of files has been dropped. Ref: https://docs.silverstripe.org/en/3/developer_guides/forms/field_types/uploadfield/#multiple-fileupload

# Acceptance Criteria

 * If a max file number is specified on an upload field, I get an error when I try to upload a single file beyond this limit
 * If a max file number is specified on an upload field, I get an error when I try to bulk upload multiple files which would push beyond this limit
 * In both cases, the file isn't uploaded in the first place (no need to clean up on the server)

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/707

tractorcow:
The removal was intentional; The existing feature set was pretty unwieldly.

I'll mark this as an enhancement request to re-implement this feature. :)

@chillu do we want to re-introduce this feature?

chillu:
Hmm, I think it's a useful feature, but I agree that it's `impact/low` for us at the moment. An alternative solution would be to put those constraints in a GridField (e.g. limiting how many `GalleryItem` a `Gallery` can have, where `GalleryItem has_one Image`). That doesn't work for everyone of course. But I can see it becoming more common with an increased use of content blocks

PR #707, opened 2017-12-12
Feature re-instate `setAllowedMaxFileNumber()` to the new UploadField

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/683

Screenshot of feature
screen shot 2017-12-13 at 11 03 07 am

Scenario for the screenshot above: I had two files already added, max files were set to three and I dragged three to the dropzone to upload - one succeeded, and two failed.


client/src/components/UploadField/UploadField.js
```diff
@@ -241,17 +241,17 @@ class UploadField extends Component {
       height: CONSTANTS.SMALL_THUMBNAIL_HEIGHT,
       width: CONSTANTS.SMALL_THUMBNAIL_WIDTH,
     };
-    const name = this.props.name;
+    const maxFiles = this.props.data.multi ? this.props.data.maxFiles : 1;
```
phalkunz: This says that if multi isn't defined, use `1` (as a default) but without set max number file upload doesn't show.

flamerohr: Thanks, I've pushed up a fix

Issue #708, opened 2017-12-12
AssetAdmin does not report any _unexpected_ errors, and just stops.


NightJar:
`SilverStripe\Assets\Upload_Validator::validate` runs a series of checks and returns false if one fails (true otherwise).
This appears to work reasonably well in common use cases (such as a file being over the legal limit as such), however does not if for example we have a custom rule. The file is sent off, and a perfectly valid response is sent back, however there is precisely zero feedback from the AssetAdmin, and it refuses to do any more business with invalid (or sometimes valid) uploads.

@unclecheese advises me that the current checks in `Upload_Validator` are recreated and performed on the client side. Simply prepending the `validate` method with

```php
$this->errors[] = 'Test bad response';
return false;
```

is enough to hack in an always repeatable occurrence of the error.

**tl;dr** AsssetAdmin ignores all server side validation errors.

unclecheese:
Resolved here: https://github.com/silverstripe/silverstripe-asset-admin/pull/709

tractorcow:
I'm reviewing the fix.

tractorcow:
The fix is good and can be merged on green.

PR #709, opened 2017-12-13
BUGFIX: Fix server error responses not displaying in UI

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/708


tractorcow:
Nice job. :D

tractorcow:
I've tested it and it works flawlessly. good job!

tractorcow:
This can be merged on green (behat tests excluded because they are currently messed up).

NightJar:
Only behat is red, so that's a green? :)

PR #710, opened 2017-12-14
BUG Try fixing behat by using chrome driver instead

tractorcow:
Just an experiment to see how well it works


tractorcow:
Need to setup chromedriver I think

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-asset-admin/pull/711 instead

Issue #179, opened 2016-05-31
Refactor <Gallery> and <Editor> into containers and presentational components


chillu:
Currently, these components have both state and render views, which is an antipattern - see https://medium.com/@learnreact/container-components-c0e67432e005#.jv7qgdyhi and https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.b5tp6c91z. 

This is setting a bad precedent on our own conventions (we have separate `containers` and `components` folders, but they're all over the show)

Note that kickassets does this pretty well, check `containers` vs. `views` folders: https://github.com/unclecheese/silverstripe-kickassets/tree/master/javascript/src/

chillu:
@flamerohr @unclecheese Do you think this is the next most important cleanup activity to work on after Chris' work on "Cleanup React Conventions" (https://github.com/silverstripe/silverstripe-framework/issues/6427)? What React code patterns are giving us the most grief in practice? I'm keen to sort out containers vs. components so we can create a pattern library that's not dependant on Redux/GraphQL and other plumbing, but that's broader (and more shallow) than this cleanup task.

flamerohr:
Yes, I think so :)
At the moment, it's very difficult to shift presentation elements around in these Components without the need to also jump through the redux state.
Also if we want to allow these components to become customisable (Injector), it's much easier to customise a presentation component when it doesn't handle a state.

Probably outside the scope of this work:
We could also look at how we're communicating across components when something needs to refresh, which sometimes need weird workarounds to achieve...
e.g. to refresh the Editor, you need to close then reopen the Editor and that causes the gallery to reload twice as well.

unclecheese:
Agreed. My biggest concerns are:

* More widespread use of Injector
* More reliance on redux for actions instead of passing handler props
* More reliance on redux for state instead of passing data props (injecting the store deeper in the tree improves rendering performance, too)
* Make components more single-concern
* More containers, as described above

chillu:
> We could also look at how we're communicating across components when something needs to refresh

This would be solved by more widespread use of GraphQL, right? The fact that form submissions aren't using GraphQL seems to be at the root of this issue.

@unclecheese Can you please write up a ticket or two about your proposed improvements? Something that'll fit in a ~2 day timebox.

chillu:
Sigh, not going to fit in the milestone

PR #712, opened 2017-12-16
Enhancement: add injector to gallery toolbar

blueo:
This is prototype for adding some injection points to the gallery section so the upload/create files buttons can be customised or added to.

Currently it is tricky to separate theses buttons because they effect changes that use the router/params. To this end, I have created an `adminDecorator` to allow passing of the router and location, along with common url-related functions to injected components (notably handleBrowse). This avoids needing to pass these objects down the component tree. 

I'd love some feedback on my approach to using the injector and the decorator concept / better ways to setup injected components.

current plans include:
- [ ]  shift more state into redux to allow cleaner separation of components
- [ ] inject the GalleryView component

relevant issues:
- https://github.com/silverstripe/silverstripe-asset-admin/issues/179


blueo:
happy new year @flamerohr :) I've removed the adminDecorator and changed GalleryToolbar so its buttons can be changed via props or new buttons can be added as 'children'.

client/src/components/GalleryToolbar/tests/GalleryToolbar-test.js
```diff
@@ -0,0 +1,94 @@
+/* global jest, describe, it, expect, beforeEach, Event */
+// mock sub-components, as they could rely on a Redux store context and not necessary for unit test
+jest.mock('components/FormAlert/FormAlert');
+jest.mock('components/AssetDropzone/AssetDropzone');
+jest.mock('components/BulkActions/BulkActions');
+jest.mock('containers/MoveModal/MoveModal');
+jest.mock('../Buttons/BackButton');
```
flamerohr: I think we don't need to mock BackButton if you can override it in the props already?
Could mock all the child components in the props where possible

client/src/containers/Gallery/Gallery.js
```diff
@@ -1124,8 +945,14 @@ function mapStateToProps(state, ownProps) {
     badges,
     concatenateSelect,
     loading,
+    sorters,
   } = state.assetAdmin.gallery;
 
+  // set default sort
+  if (!sort) {
+    sort = `${sorters[0].field},${sorters[0].direction}`;
+  }
```
flamerohr: how come this was moved from `defaultProps`?

client/src/containers/Gallery/Gallery.js
```diff
@@ -1148,9 +977,18 @@ function mapDispatchToProps(dispatch) {
...
+export {
+  Gallery as Component,
+  galleryViewPropTypes,
+  galleryViewDefaultProps,
+};
 
 export default compose(
+  inject(
+    ['GalleryToolbar'],
+    GalleryToolbar => ({ GalleryToolbar }),
```
flamerohr: this could be `null` for default behaviour :)

client/src/containers/TableView/TableView.js
```diff
@@ -362,7 +362,10 @@ class TableView extends Component {
 
 TableView.defaultProps = galleryViewDefaultProps;
 
-TableView.propTypes = galleryViewPropTypes;
+TableView.propTypes = {
+  ...galleryViewPropTypes,
+  sort: PropTypes.string.isRequired,
```
flamerohr: I see this belonging in the `galleryViewPropTypes` variable more than being standalone here and not required.

I'm pretty sure it was defined in a sharedPropTypes somewhere

client/src/containers/Gallery/Gallery.js
blueo: The default sort was created from a constant but I've moved the sort info into redux so the default can't be created without the context from the store.

client/src/containers/TableView/TableView.js
blueo: It was in the shared props but I moved it for a couple of reasons:
1. Not all the components with shared props needed it (ThumbnailView).
2. I wanted to make the proptype check stricter on those components that did need it. 

I could move it back to being an optional prop in the shared variable though if needed

client/src/components/GalleryToolbar/Buttons/tests/BackButton-test.js
```diff
@@ -0,0 +1,46 @@
...
+      },
+      handleMoveFiles: () => {},
+      onOpenFolder: () => {},
+      badges: [],
+    };
+  });
+
+  describe('render BackButton', () => {
+    it('should not render if parentId is not set', () => {
+      expect(Component({ ...props })).toBeNull();
```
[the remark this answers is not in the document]
blueo: huh so turns out `renderIntoDocument` doesn't play well with stateless components. (see https://stackoverflow.com/questions/36682241/testing-functional-components-with-renderintodocument). The options are to 1) leave it as it is or 2) add a wrapper. I'd prefer to leave it like this unless you already have a similar setup for other components?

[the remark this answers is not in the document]
flamerohr: huh...
I think we only have one or two stateless components and the rest as classes regardless of stateful-ness.
Could we convert the component is a class? (with eslint-disable if required)

I've been meaning to turn off the `react/prefer-stateless-components` rule for a while now.

client/src/containers/Gallery/Gallery.js
flamerohr: Sort wasn't stored in redux because it came from `react-router` url params - duplication of data.

Hmm... I didn't notice it being moved to redux.

client/src/containers/TableView/TableView.js
flamerohr: ah, I overlooked ThumbnailView not needing it now (used to be, until the sorter got move up a level)

cool, this stays then :D

client/src/containers/Gallery/Gallery.js
blueo: sorry 'sort' does come from the router but I moved 'sorters' as otherwise it was impossible to customise the sort dropdown. But the result is that the default 'sort' needs the redux info from 'sorters'

flamerohr: oh, you're right!

neat, ok I can see why it's needed here now.

client/src/components/GalleryToolbar/Buttons/tests/AddFolderButton-test.js
```diff
@@ -0,0 +1,19 @@
...
+  describe('handleCreateFolder()', () => {
+    it('should call onCreateFolder on click', () => {
+      const props = {
+        canEdit: true,
+        onCreateFolder: jest.genMockFunction()
+      };
+      const button = ReactTestUtils.renderIntoDocument(<Component {...props} />);
+      button.handleCreateFolder(new Event('click'));
+
+      expect(props.onCreateFolder).toBeCalledWith();
```
flamerohr: `.toBeCalled()` should be enough if no params are expected :)

flamerohr:
I had amended the props `handleMoveFile` and `handleSort` which I overlooked and should've been named `onMoveFile` and `onSort` :)

`handleMoveFile` and `handleSort` are valid class method names, but props should be `on` prefixed instead

Issue #705, opened 2017-12-11
Up / Upload / Add Folder buttons unresponsive most of the time


axllent:
All the upper-left buttons (up, upload, add folder) require multiple clicks in different areas to actually activate them. I can replicate this on both Google Chrome as well as Firefox (Ubuntu), and happens both in the responsive as well as desktop layout (demo in responsive mode to save space).

![peek 2017-12-12 12-30](https://user-images.githubusercontent.com/1463435/33859366-7b1d6e48-df38-11e7-87ce-9534ff75d72c.gif)

# Pull requests

 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/714

tractorcow:
Any errors in your JS console?
Live or dev mode?

axllent:
No errors in JS console, both live and dev, and I can confirm it happens here on another virtual machine running ie11. This has been happening since the very first alphas right through to current stable. I just tried a fresh install with no additional modules and it has the same effect.

tractorcow:
@flamerohr can you replicate?

flamerohr:
I have an idea of what you're talking about, sometimes having the buttons become unresponsive... however I haven't been able to replicate the issue consistently on my end :/

axllent:
Strange, I can replicate it consistently, in fact it's extremely hard to get any of those buttons to work. Every other button throughout the CMS works flawlessly, so it is related to these specific ones. 

So Linux & Windows, IE, Chrome & Firefox - so it's not an OS-thing or browser-thing - it must be something else. As I mentioned previously, no JS errors, nor any server response delays (ie: same with the browser file dialogue). It shouldn't matter, but this is on Apache.

phalkunz:
I saw that happened but I couldn't find a pattern that reproduces all the time either.

phalkunz:
Turns out it's caused by a feature that enables multiple-item selection by mouse. This issue can be reproduced by:

1. Hold the left mouse button on "Add folder" button
2. While still holding mouse button, move the mouse a little. This cause the UI to go into selection mode.
3. Release the mouse button

🐞 Observe that Create folder panel doesn't show up.

axllent:
[...]

PR #714, opened 2017-12-19
FIX mouse multi-section prevent buttons from working

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/705


phalkunz:
@flamerohr I've created an issue as an improvement to this https://github.com/silverstripe/silverstripe-asset-admin/issues/716

Issue #699, opened 2017-11-29
Bug: error occurs when checking a newly-uploaded file and publishing via "bulk actions"


colintucker:
## Overview

A JavaScript error "File selected for publishing cannot be found: <ID>" occurs in `asset-admin` when attempting to publish a newly-uploaded file via the "bulk actions" toolbar at the top of the panel. Refreshing the entire admin allows the files to be published in this manner.

## Expected Behaviour

Checking one or more newly-uploaded files via their checkboxes and then clicking "... > Publish" via the bulk actions toolbar should publish each selected file.

## Observed Behaviour

The SS loading spinner appears but never vanishes, and the admin appears to be frozen. The browser console shows the following error (asset ID will vary):

![image](https://user-images.githubusercontent.com/1853705/33353732-58fcbf7e-d503-11e7-9e4f-31665021d860.png)

## Steps to Reproduce

1. Within `asset-admin`, upload a file from your computer.
2. Without refreshing the admin, check the new file's checkbox in the bottom-right.
3. Click the ellipsis (...) in the "bulk actions" toolbar that appears at the top of the panel.
4. Click "Publish".

For me on the platform below, this can be replicated 100% of the time.

## Platform

- Windows 10 Pro x64 1703 15063.726
- Firefox Developer Edition 58.0b6 x64
- SilverStripe 4.0.0

## Pull requests

- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/715

tractorcow:
I've attempted to reproduce this on the current 4.0.x branch and it doesn't seem to be an issue.

Possibly this was fixed since 4.0.0 stable, or maybe it's windows-only.

@colintucker can you please check the HTTP respones for the admin/assets/api/createFile upload?

It should be JSON with the below format

![image](https://user-images.githubusercontent.com/936064/33403364-fa3acc66-d5c4-11e7-93d2-8918cb8be36d.png)

Possibly if files uploads are failing, subsequent attempts to publish could fail also.

colintucker:
Sure, here's a test file I just uploaded:

![image](https://user-images.githubusercontent.com/1853705/33408377-4a691f4e-d5ca-11e7-997e-0e9cb697cd36.png)

tractorcow:
We'll need to do a bit more work to attempt to reproduce it then. Thanks for the feedback so far!

unclecheese:
[...]

PR #715, opened 2017-12-19
Pulls/1/fix delete uploaded files

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/699

---

Refer to @unclecheese's comment for more details https://github.com/silverstripe/silverstripe-asset-admin/issues/699#issuecomment-352634380


client/src/containers/Gallery/Gallery.js
```diff
@@ -102,14 +102,6 @@ class Gallery extends Component {
 
       $select.off('change');
     }
-
-    // If props.files has changed, flush any uploaded files.
-    // The render() logic will merge queuedFiles into the props.files array
-    // until this is called, leaving completed uploads temporarily in the current view,
-    // even if they're technically not part of props.files.
-    if (!this.compareFiles(this.props.files, nextProps.files)) {
```
flamerohr: I think this is the only use of `compareFiles()` which means it's no longer needed.

Could we remove it and associated tests?

unclecheese: Yeah, good catch. That function is showing its age.

tractorcow: When is purgeUploadQueue() called now?

tractorcow: What about if folder ID changes? It looks like it only happens on sort otherwise.

phalkunz: @tractorcow `purgeUploadQueue()` is not being called from anywhere at the moment but @unclecheese said to keep it. I can't answer your second question but I'm sure @unclecheese can.

tractorcow:
As discussed at https://github.com/silverstripe/silverstripe-asset-admin/issues/699#event-1396843654 the only needed fix was the ID cast. Is that sufficient?

flamerohr:
Yup, lets do that, can we fix the casting on the server-side?

client-side is fine, but this leaves a tiny bit of tech debt for future development

tractorcow:
I've switched to 4.0 branch and switched the files comparison for a simple folder id check. :)

flamerohr:
@tractorcow nice one, but that still doesn't resolve my original comment :)
`compareFiles` is no longer needed/used, so it could be removed.

tractorcow:
Yeah but removing methods is semver breakage, so I've deprecated it instead.

tractorcow:
Ok I've removed it after confirming Gallery.js isn't public API. :)

Issue #718, opened 2017-12-28
Make UploadFieldItem decorateable


bummzack:
I'm still trying to update the [sortablefile](https://github.com/bummzack/sortablefile/issues/50) module to SS4… and sadly it's not that easy.

Decorating `UploadField` as outlined in the [docs](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/reactjs_redux_and_graphql/#customising-react-components-with-injector) now works, but that component is pretty much a blackbox and doesn't expose much via props.

To make files attached to the `UploadField` sortable, it would require a sortable container, as well as sortable children. There's currently no way to decorate the child components, as they don't seem to be exposed via `injector`. Ideally, the rendering of the child components would be exposed as a prop on `UploadField`, so that a decorator can pass in a custom function.

Example (not sure if that would work, but I hope the idea is understandable):
 
```js
renderChild(item) {
  const itemProps = {
    // otherwise only one error file is shown and the rest are hidden due to having the same `key`
    key: item.id ? `file-${item.id}` : `queued-${item.queuedId}`,
    item,
    name: this.props.name,
    onRemove: this.handleItemRemove,
    canEdit: this.canEdit(),
    onView: this.handleReplaceShow,
  };
  const renderer = this.props.childRenderer || function(FieldItemClass, props) {
    return <FieldItemClass ...props/>;
  }
  return renderer(UploadFieldItem, itemProps);
}
```

That way a decorator to `UploadField` could set the `childRenderer` property to customize `UploadFieldItem`

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-admin/pull/394
- [x] https://github.com/silverstripe/silverstripe-admin/pull/405
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/722
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7768

chillu:
Usually you'd pass in another child component prototype through props of the parent container, which adheres to the same "interface" (required props). Or if you're not in charge of creating the parent container, you'd do it via our custom `Injector`. 

Passing through a "renderer closure" (`this.props.childRenderer`) is more powerful since you can choose to pass through more props and state from the parent container (`<UploadField>`), but it also subverts the idea of having a "component interface" with the same props. And it can get a bit out of hand if applied at every level of the component hierarchy.

[...]

PR #722, opened 2018-01-16
Enhancement UploadField uses injector for child components

flamerohr:
Requires https://github.com/silverstripe/silverstripe-admin/pull/394

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/718 to become more extensible


client/src/components/UploadField/tests/UploadField-test.js
```diff
@@ -1,7 +1,5 @@
 /* global jest, jasmine, describe, it, expect, beforeEach */
 
-jest.mock('containers/InsertMediaModal/InsertMediaModal');
```
flamerohr: Having removed this line - I really like to option to pass a prop to override the injected component :D

unclecheese:
Have rebased to the `GalleryToolbar` changes.

flamerohr:
Experiencing some (heavy) loading issues with this now...

We've tracked down the issue to composing every render which is cache busting and re-instantiating almost every component per cycle 💥 

Fix found here https://github.com/silverstripe/silverstripe-admin/pull/405

unclecheese:
Behat test is failing on missing file history form. I can't replicate it locally, so I think it just needs a `    I wait for 2 seconds until I see the "#Form_fileHistoryForm" element`

flamerohr:
Found this bug in cms causing the behat to fail https://github.com/silverstripe/silverstripe-cms/pull/2079

unclecheese:
Merged. Rebasing.

flamerohr:
All green :D

PR #723, opened 2018-01-16
API Remove reference to removed constructExtensions()

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-framework/issues/6210

### DIFF
--- a/code/Forms/AssetFormFactory.php
+++ b/code/Forms/AssetFormFactory.php
@@ -56,7 +56,6 @@ abstract class AssetFormFactory implements FormFactory
 
     public function __construct()
     {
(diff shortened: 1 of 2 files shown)


tractorcow:
Need to merge in the phpunit / php version change.

robbieaverill:
Just came across this in 1.0 as well - do you think this is an API change? I can't see it breaking anything

tractorcow:
Not really I guess; it's more updating to respond to API changes elsewhere.

tractorcow:
This was actually done up in 4.0 branch so it's not necessary to do here also.

Issue #719, opened 2018-01-15
Freshly uploaded items always show in search


blueo:
The search results always include new assets immediately after uploading.

Steps to reproduce:
1. upload a new asset
2. open the search bar and perform a search to exclude the new file (eg filter for documents after having uploaded an image). 
3. note that search results include the new asset even though it should be excluded.

Interestingly the Graphql response does not include any files - so this is likely some kind of client side caching.

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/726

flamerohr:
that's correct, uploaded files are stored and listed on top of the current "page" or "results"

I guess it would make sense to clear that store when a search has been executed though... although, if you upload during a search then you'd still have that similar experience

blueo:
yeah it would useful when you upload during search but a little unexpected when you're not - maybe it is more about visually distinguishing them from the search results so it is clear where they come from

flamerohr:
yeah, we have plans to move uploaded items to a separate panel or something along those lines... in the future.

what I've suggested to clear is the reason I've marked this as a `bug` :)

clarkepaul:
Marking this as potential UX item as we might be able to suggest a mid-way solution.

flamerohr:
@clarkepaul sounds good, if it's a new UX element then I'd recommend a new ticket for that

a bug can be added to a patch release (e.g. 4.1.1) which means users get this bug fixed sooner
but a new UX element would be more suited to a minor release (e.g. 4.2.0)

PR #726, opened 2018-01-25
Fix remove uploaded items when executing or removing search

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/719


client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -223,6 +226,18 @@ class AssetAdmin extends Component {
...
 
+  /**
+   * Reset to non-search page
+   *
+   * @param event
+   */
+  handleClearSearch(event) {
+    this.props.actions.gallery.deselectFiles();
+    this.props.actions.queuedFiles.purgeUploadQueue();
+    this.props.actions.files.readFiles();
```
flamerohr: calling `readFiles()` (apollo's `refetch`) was the only way I could think of to refresh/break-cache the current page items...
otherwise we'll have instances of the uploaded item not showing if uploaded, then searching and clearing search.

unclecheese:
Test is failing on `git diff`. Did you maybe forget to build, @flamerohr ?

flamerohr:
css variables are different between `1.0` and `1`... hence the build difference

PR #732, opened 2018-02-01
API Enable upload / attach options on UploadField to be individually toggled on or off

### LINKED ISSUE #399: UploadField missing canUpload option
\SilverStripe\Forms\UploadField had a canUpload option that allowed the field to only be used to select from already uploaded assets.

\SilverStripe\Forms\UploadField has been removed and it's replacement, SilverStripe\AssetAdmin\Forms\UploadField, doesn't have a way to restrict users to only selecting from existing assets.

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/732

tractorcow:
Fixes #399


client/src/components/UploadField/UploadField.js
```diff
@@ -269,11 +292,29 @@ class UploadField extends Component {
...
+    if (this.canUpload()) {
+      options.push(
+        <button key="uploadbutton" type="button" onClick={this.handleSelect} className="uploadfield__upload-button">
+          {i18n._t('AssetAdmin.BROWSE', 'Browse')}
+        </button>
+      );
+    }
+    if (this.canAttach()) {
+      if (options.length) {
+        options.push(` ${i18n._t('AssetAdmin.OR', 'or')} `);
```
flamerohr: this is not good practice... it's a similar practice to having entries for singular/plural in the code...
have a translation with named placeholders would be a better practice, or omit this "or" conjoin.

tractorcow: It looks like this

image

flamerohr: ok, since translations work horribly with React, I'd suggested to move the spaces into the translation entry and remove the template string wrapping the translation

client/src/components/UploadField/UploadField.js
```diff
@@ -316,7 +361,7 @@ class UploadField extends Component {
         type="select"
         bodyClassName="modal__dialog"
         className="insert-media-react__dialog-wrapper"
-        fileAttributes={selectingItem ? { ID: selectingItem.id } : null}
+        fileAttributes={selectingItem ? {ID: selectingItem.id} : null}
```
flamerohr: the spaces are required by linter

tractorcow:
ok fixed @flamerohr

Issue #424, opened 2017-04-10
Improvements to selecting/dragging and moving of files


clarkepaul:
# Acceptance criteria

* When a file is being dragged the cursor should change from a pointer to a drag cursor. 
* Fade-out/remove original file being dragged in original location and fade-in the file attached to the cursor (too much emphasis on the old place and not enough on the file being dragged).
* Improve the highlighting of the drop-zones (back up folder, into folders) as they are hidden.
* Include intuitive keyboard interactions and commands for selecting items.
  * Cmd (Mac) or Ctrl (PC) and select to check/uncheck items
  * Select start, hold shift and select end item to select all items in between
* ~~Include a drag to select area option for selecting multiple items.~~ (tracked in https://github.com/silverstripe/silverstripe-admin/issues/127)

# Designs
Interactive prototype: https://invis.io/27C0T40G4#/237327492_Drag_N_Drop_Select
Flat style guide: ?

# Notes

 * Drag and drop of list items out of scope?
 * See [Add multi-select shortcuts for dragging files](https://github.com/silverstripe/silverstripe-admin/issues/127)

# Tasks
- [ ] Design review after development

# Pull requests
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/738

flamerohr:
Pointer: css rules to apply the correct cursor
Might need a boolean prop for `GalleryItem` to indicate that dragging is happening, so that it knows to face out.
This may require some jumping around, or maybe my I'm overestimating the complexity.

I've changed the effort required to reflect the slight difficulty that this could present

flamerohr:
@newleeland Some feedback, I agree with the first three points, it really needs an improvement there... what do you think about reducing the size of the items being dragged? That way it'd make it easier to make the folder areas easier to highlight/interact with.

About the keyboard/additonal interactions:
> Cmd (Mac) Cntrl (PC) and select to check/uncheck items

Are you talking about checking anywhere on the tile? So if the user is holding down ctrl/cmd and click on the thumbnail, it wouldn't open the file but instead select/deselect it?

> Select start, hold shift and select end item to select all items in between

I kind of wonder if this is still relevant with the select-box implemented...
This one is going to be tricky to understand what behaviour we would want, let alone implement.
So say we have this grid of items in the gallery:
```
  1  |  2  |  3  
  4  |  5  |  6
  7  |  8  |  9
```
[...]

PR #738, opened 2018-02-19
Improve styles for dragging and dropping files

flamerohr:
Also added a "shift+click" function, so you can select two files while holding shift and all files between them will also be selected

screen shot 2018-02-20 at 11 02 33 am

Please see https://github.com/silverstripe/silverstripe-asset-admin/issues/424 for more details.

cc: @newleeland for design review


client/src/components/GalleryItem/GalleryItemDragLayer.js
```diff
@@ -19,23 +22,31 @@ class GalleryItemDragLayer extends Component {
     if (!item.selected) {
```
unclecheese: Could use `const { item } = this.props` just for consistency.

client/src/components/GalleryItem/GalleryItemDragLayer.js
```diff
@@ -19,23 +22,31 @@ class GalleryItemDragLayer extends Component {
...
       return null;
     }
-    const selected = `${item.selected.length}`;
+    const shadows = [
+      item.selected.length > 1
```
unclecheese: `item.selected.length` has several references in this function. Maybe assign it to `selectionCount` just to keep things tidy?

unclecheese:
Looks great. I just made some minor changes to variable names and assignments. Overall really nice work on a pretty gnarly API!


Biggest issue is browser testing. All versions of IE and Edge are having issues with the drag layer. Basically, the native HTML5 drag layer is shown, but your custom folder designs are hidden behind it.

It doesn't completely kill the UX, but it does render most of this work moot for IE users, so might want to ask PO @chillu how much time you want to invest in fixing it.

dnd

unclecheese:
I remember having to navigate this on [kickassets](github.com/unclecheese/silverstripe-kickassets), and I believe this was the solution I used https://stackoverflow.com/questions/36379184/html5-draggable-hide-original-element#answer-36379727

chillu:
IE is still a sizeable portion of users. Google Analytics says around 8% of traffic on userhelp. I'd cap investigation into this at 2h, otherwise we'll leave it.

unclecheese:
The `git diff` NPM test is failing. Will rebuild and push up and see if that sorts it.

PR #739, opened 2018-02-20
Add warning when unpublishing owned files

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-versioned/issues/26


client/src/boot/applyTransform.js
```diff
@@ -53,6 +54,43 @@ const ConnectedMoveTreeDropdownField = compose(
...
+            'AssetAdmin.SINGLE_OWNED_WARNING_2',
+            'Ensure files are removed from content areas prior to unpublishing them. Otherwise, they will appear as broken links.'
+          ),
+          i18n._t(
+            'AssetAdmin.SINGLE_OWNED_WARNING_3',
+            'Do you want to unpublish this file anyway?'
+          )
+        ];
+        // eslint-disable-next-line no-alert
+        if (confirm(message.join('\n\n'))) {
```
tractorcow: Isn't this logic backwards? confirm() would be true if the user wanted to continue unpublishing.

client/src/state/files/createPublicationMutation.js
```diff
@@ -4,10 +4,20 @@ import { fileInterface, file } from 'lib/fileFragments';
...
-      ...FileInterfaceFields
-      ...FileFields
+  mutation ${operationName}($IDs:[ID]!, $Force:Boolean) {
+    ${mutationName}(IDs: $IDs, Force: $Force) {
+      ...on File {
+        __typename
+        ...FileInterfaceFields
+        ...FileFields
+      }
+      ...on PublicationError {
```
tractorcow: Had a discussion with @flamerohr about the way that we encode errors in graphql responses.

I don't think it's correct to call this an error; It's a confirmation required notice correct? I suggest to rename this to `PublicationConfirmationRequired`, or something to that effect.

code/GraphQL/PublicationMutationCreator.php
```diff
@@ -60,37 +64,59 @@ public function args()
...
         if ($files->count() < count($idList)) {
             // Find out which files count not be found
             $missingIds = array_diff($idList, $files->column('ID'));
-            throw new \InvalidArgumentException(sprintf(
-                '%s#%s items are not published or don\'t exist',
-                File::class,
-                implode(', ', $missingIds)
-            ));
+            foreach ($missingIds as $id) {
+                $result[] = new OperationError(
```
tractorcow: Is this coming out through the `PublicationError` object? This seems like an application error that should be thrown normally, not re-implemented as a "soft" error.

code/GraphQL/PublicationMutationCreator.php
```diff
@@ -60,37 +64,59 @@ public function args()
...
+                    'NON_EXISTENT',
+                    LogLevel::ERROR,
+                    [$id]
+                );
+            }
         }
 
-        $writtenFiles = [];
         foreach ($files as $file) {
             if ($this->hasPermission($file, $context['currentUser'])) {
```
tractorcow: If the user doesn't have permission to publish this file, selecting this option should disable the "publish" batch action.

Then any subsequent publication attempts should be real errors.

code/GraphQL/UnpublishFileMutationCreator.php
```diff
@@ -37,10 +40,33 @@ protected function hasPermission(File $file, Member $member)
...
-    protected function mutateFile(File $file)
+    protected function mutateFile(File $file, $args = [])
     {
+        if (!isset($args['Force']) || !$args['Force'] && $file->hasExtension(RecursivePublishable::class)) {
+            $owners = $file->findOwners();
+            if ($owners->exists()) {
+                return new OperationError(
+                    _t(
+                        __CLASS__ . 'OWNER_WARNING',
+                        'File "{file}" is used in {count} place|File {file} is used in {count} places.',
```
tractorcow: Do we really need to show all of these individual messages in the confirmation dialog? Don't we need a simple count?

tractorcow: See `ownedByFailures.map(failure => failure.Message).join('\n'),` in the confirmation dialog; This could get huge in a confirm box if unpublishing a large folder.

client/src/boot/applyTransform.js
unclecheese: Yup. I think I deleted the `!` during linting.

unclecheese:
We're not suppressing real errors. If the resolver throws an exception, it still comes back as an error, using the standard graphql/apollo behaviour. That hasn't changed.

I don't think the standard graphql errors are suited for what we're trying to do, though. Because we're using the errors to make decisions in the UI (confirm and force), the error layer simply isn't robust enough to handle that without some really ugly hacks. For one, we don't know the nature of the error.  A notice about ownership is treated the same as a database error. Additionally, we don't have things like IDs, which inform the UI how to force the next query after confirmation. Lastly, a resolver can only throw once, and because this is all happening in one resolver, that precludes us from doing any kind of split response of exceptions + successes.

The author of the graphql-php module recommends [modelling the errors into your schema](https://github.com/webonyx/graphql-php/issues/207) for this reason.

We need some concept of partial failures. This doesn't need to be as verbose as a list of each thing that failed (although that would be useful if we were doing any kind of UI more elegant than a browser alert), but it does need to be part of the response. Something like this might be a good halfway measure:

```
mutation unpublishFiles([IDs], SuppressNotice = false) {
  results {

  }
  notice {
    Message
    [IDs]
  }
}
```

Where `notice` is a single type, not a list of types, that aggregates all the failures and provides pertinent information back to the UI.

tractorcow:
> We're not suppressing real errors. If the resolver throws an exception, it still comes back as an error, using the standard graphql/apollo behaviour. That hasn't changed.

The problem I'm talking about is real errors that aren't thrown as exceptions but should be. Instead we try to bundle them into a response to show in the UX.

> Because we're using the errors to make decisions in the UI (confirm and force

We shouldn't be using errors of any type to make decisions in the UI; Finding a record with an owner isn't an error, it's simply a confirmation-required state. States outside of this situation (errors or otherwise) are not in the scope of this story.
 
> We need some concept of partial failures.

Possibly, but for now I recommend to keep the status quo. I.e. if records somehow got deleted or canUnpublish = false just skip them. Those are all errors that should be prevented prior to the graphql action.

> The author of the graphql-php module recommends modelling the errors into your schema for this reason.

I think this line is thinking is worthy of research and investigation in the future.

unclecheese:
> The problem I'm talking about is real errors that aren't thrown as exceptions but should be.

Which real errors are you talking about? The only "errors" that have been converted to notices are permission check failures, ID not found, and the new has-owner check. None of those IMO are worthy of killing the mutation, but they're all worthy of informing the user in some kind of elegant way. Any other type of network/database/query error will still behave the way it used to.

> We shouldn't be using errors of any type to make decisions in the UI;

Agreed. So let's not call them errors. If `PublicationError` it has to be `PublicationNotice` or whatever else, I'm totally with you on that. We just need some way of communicating in the response that some mutations didn't get executed and why that is. Whether we call those "errors" or not is immaterial.

>  if records somehow got deleted or canUnpublish = false just skip them.

As of now, those conditions throw. So what this PR does is not eliminate them all together, but soften them. If we want to eliminate them all together, we can do that, but if anything, I'd rather keep them as hard errors, because eliminating exceptions full stop seems out of scope for this story.

tractorcow:
Ok, I've added a bunch more tests, updated localisations, and fixed a couple of ownership lookup bugs (nothing major).

A problem I've found is that the OLD batch-unpublish warning is clashing with this new one.

This is what I see (clicking "OK" after each)

image

image

And then it does the unpublish.

PR #740, opened 2018-02-22
Add gallery item injector and extension points

blueo:


tractorcow:
@flamerohr can you please peer review this? :P

flamerohr:
yup, was planning to :)

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -183,22 +185,24 @@ class GalleryItem extends Component {
...
-      } else if (this.props.item.modified) {
+      } else if (item.modified) {
         flags.push(<span
           key="status-modified"
           title={i18n._t('File.MODIFIED', 'Modified')}
           className="gallery-item--modified"
         />);
       }
     }
+    flags = updateStatusFlags(flags, this.props);
```
flamerohr: hmm, this seems weird, maybe we could have an array formatted:
```
[
  { key: 'draft', message: i18n._t('File.DRAFT', 'Draft') }
]
```
Then at the end after `updateStatusFlags()` is called, we can render them to `<span>`

What do you think?
This will mean it's not possible to replace the `<span>` element...
But it would enable much easier updating messages/keys (e.g. otherwise I'd need to use `React.children` API to clone and modify, or outright replace/duplicate it)

blueo: Sounds good, what do you think of this approach (allows changing the span):
```Javascript
getStatusFlags() {
    let flags = [];
    const { item, updateStatusFlags } = this.props;
    if (item.type !== 'folder') {
      if (item.draft) {
        flags.push({
          node: 'span',
          key: 'status-draft',
          title: i18n._t('File.DRAFT', 'Draft'),
          className: 'gallery-item--draft',
        });
      } else if (item.modified) {
        flags.push({
          node: 'span',
          key: 'status-modified',
          title: i18n._t('File.MODIFIED', 'Modified'),
          className: 'gallery-item--modified',
        });
      }
    }
    flags = updateStatusFlags(flags, this.props);
    return flags.map(({tag, ...attributes}) => <tag {...attributes} />);
  }
```

flamerohr: yup, that looks good, except use correct object keys in the final map :)
e.g. uppercase the first character in the variable name `tag` to `node: Tag` or the react transpiler will treat that as a normal html element

`gallery-item--modified` is bad BEM for these flags, and it's outside the scope of this pull request.

blueo: ah copy paste fail :( must be friday.

PR #741, opened 2018-02-27
Enhancement expose extension point for item props in uploadfield

flamerohr:
Fix check container before unmounting react-dom


client/src/components/UploadField/UploadField.js
```diff
@@ -383,6 +384,10 @@ class UploadField extends Component {
       canEdit: this.canEdit(),
       onView: this.handleReplaceShow,
     };
+    const itemProps = (typeof this.props.getItemProps === 'function')
```
tractorcow: It's always a function isn't it? It has a default. This check can be left out then?

flamerohr: hmm true, I had originally not have the defaultProp

tractorcow:
Can be merged on green.

PR #751, opened 2018-03-15
Add ability to use additional Toolbar buttons on the Gallery

oilee80:
In a 3rd party JS file you can add a new Component that will be rendered

```js
import React, { Component, PropTypes } from 'react';

class NewButton extends Component {
    render() {
        return (
          <button
            className="btn btn-secondary font-icon-sync btn--icon-xl"
            type="button"
          >
            <span className="btn__text btn__title">Button Label<span>
          </button>
        )
    }
}

window.assetAdmin = window.assetAdmin || {};
window.assetAdmin.gallery = window.assetAdmin.gallery || {};
window.assetAdmin.gallery.extraToolbarButtons = window.assetAdmin.gallery.extraToolbarButtons || [];
window.assetAdmin.gallery.extraToolbarButtons.push(NewButton);
```

### DIFF
--- a/client/src/containers/Gallery/Gallery.js
+++ b/client/src/containers/Gallery/Gallery.js
@@ -495,6 +495,19 @@ class Gallery extends Component {
     this.props.actions.gallery.setEnableDropzone(enabled);
   }
 
+  renderExtraToolbarButtons() {
+    if (window.assetAdmin && window.assetAdmin.gallery && window.assetAdmin.gallery.extraToolbarButtons) {
+      let extraButtons = window.assetAdmin.gallery.extraToolbarButtons;
+      return (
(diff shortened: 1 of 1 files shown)


flamerohr:
Sorry, but you can add extra buttons through `Injector` https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/reactjs_redux_and_graphql/#transforming-services-using-middleware

I don't agree with using globals to add extra Components either, this will end up cluttering the global namespace greatly

flamerohr:
For reference, the work to enable adding new buttons to toolbar is here https://github.com/silverstripe/silverstripe-asset-admin/pull/712

PR #780, opened 2018-05-07
Compliance with Apollo 2.x upgrade

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-admin/issues/423


client/src/state/files/deleteFilesMutation.js
```diff
@@ -1,20 +1,30 @@
...
-      resultBehaviors: dataIds.map(dataId => ({
-        type: 'DELETE',
-        dataId,
-      })),
+      update: (store) => {
+        const variables = readFilesConfig.options(ownProps).variables;
+        const data = store.readQuery({ query: readFilesQuery, variables });
+        // Query returns a deeply nested object. Explicit reconstruction via spreads is too verbose.
+        // This is an alternative, relatively efficient way to deep clone
+        const newData = JSON.parse(JSON.stringify(data));
```
flamerohr: yeah, this makes sense since the data is most primitives or objects/arrays :)

definitely simpler than needing an entire library like lodash

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -118,7 +118,7 @@ class ThumbnailView extends Component {
     const badge = badges.find((badgeItem) => badgeItem.id === item.id);
     let props = {
       sectionConfig,
-      key: item.id,
+      key: item.id || item.queuedId,
```
flamerohr: might be a good idea to prefix `queuedId` (or both), it's unlikely but possible to have overlapping keys

tests/behat/features/manage-files.feature
```diff
@@ -93,6 +93,7 @@ Feature: Manage files
       And I check the file named "testfile" in the gallery
     Then the ".bulk-actions-counter" element should contain "3"
       And I press the "Delete" button, confirming the dialog
+      And I wait for 5 seconds
```
flamerohr: oh boy, are we able to have retries for the "I should not see" below somehow?

client/src/containers/Gallery/Gallery.js
```diff
@@ -229,7 +229,6 @@ class Gallery extends Component {
    * @returns {Promise}
    */
   handleBulkDelete(event, items) {
-    this.props.actions.gallery.setLoading(true);
```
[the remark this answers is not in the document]
unclecheese: Hmm... Probably, but the second `setLoading` is in a callback to a promise that is defined as a prop, so we have no guarantee what state transformations are happening in that function, so you could look at the second call as kind of a reset.

I guess if someone were using a different `onDelete`, they would probably have to take ownership of the state clean up.

client/src/containers/ThumbnailView/ThumbnailView.js
unclecheese: True. Keys don't have to be semantic. Just unique. :)

PR #784, opened 2018-05-18
Add Refresh ability on the Gallery & Toolbar

oilee80:
Adds a new `handleRefreshPage` method to AssetAdmin JS which is then passed into the `Gallery` component and from there the `GalleryToolbar`

It works by passing updating the variables passed via GraphQL to include a random `cache_buster`. Just asking it to make the same request again doesn't work as the variables are the same so the response was being returned from an internal cache in the `apollo-client`

**NOTE**: Needing to add this so that a "sync" button (https://github.com/MadeHQ/silverstripe-cloudinary) that causes assets on the server to update from a third party store can then get the updates without having to navigate into and out of folders or close and then re-open the asset browser


oilee80:
Hiya @unclecheese, @flamerohr 
Just checking if you could merge this in or if there is a better way to refresh the Gallery without closing and re-opening after the server has updated

unclecheese:
Hi, @oilee80. This seems to be a domain-specific concern related to your Cloudinary module.

What you can do, ideally on the 4.2 release, is add this behaviour with Injector. The Gallery toolbar is now extensible due to @blueo 's work, and you can add arbitrary actions into it.  Further, with SS 4.2 now shipping with Apollo 2.5, you'll be afforded a better API for refreshing query state rather than using a `cachbuster` workaround.

Either myself or @flamerohr can help you out with that in a separate discussion (maybe Slack?), but in the interim, a good starting point is the docs here: https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/reactjs_redux_and_graphql/

unclecheese:
Relevant PR/discussion. https://github.com/silverstripe/silverstripe-asset-admin/pull/712

flamerohr:
Apollo provides a refetch callback which forces a new request to be made, I don't remember the specifics at the moment, but I remember it was called `readFiles()` in either the AssetAdmin or Gallery props.

Highly recommend checking out the Injector work and see if that can be utilised for your refresh button :)

maxime-rainville:
Sounds like this PR is no longer needed or that there's a better way of achieving the stated goal.

Shall we close it?

unclecheese:
Closing as it seems to be domain-specific and we haven't had any follow-up from the author.

PR #792, opened 2018-06-14
Update to use new recipe branching number

tractorcow:
Parent issue: https://github.com/silverstripe/recipe-core/issues/24

From 4.2 onwards all recipes will use the same numbering as the core framework version.


chillu:
Confirmed the test failure is the same as in the base branch

PR #793, opened 2018-06-14
Cannot embed YouTube link on CWP #2168

caffeineinc:
The embed library allows the use of custom dispatchers, but our Model in asset admin restricted the use of custom configurations and prevented custom adapters. This fix allows `Embed` to be customised and allows for a custom dispatcher to be provided. 

The custom dispatcher must be configured for use on CWP where `Embed` makes a request to get metadata about a link, this call must go through the CWP proxy, which this enables through configuration.

e.g:
```
if(!\SilverStripe\Control\Director::isDev()){
    if (defined('SS_OUTBOUND_PROXY') && defined('SS_OUTBOUND_PROXY_PORT')) {
        $config = [
            CURLOPT_PROXY => SS_OUTBOUND_PROXY,
            CURLOPT_PROXYPORT => SS_OUTBOUND_PROXY_PORT
        ];

        $dispatcher = new \Embed\Http\CurlDispatcher($config);
        EmbedResource::setDispatcher($dispatcher);
    }
}
```


code/Model/EmbedResource.php
```diff
@@ -23,9 +40,17 @@ class EmbedResource
 
     public function __construct($url)
     {
-        $this->embed = Embed::create($url);
```
robbieaverill: For config we prefer ->get() for props these days.

Also the dispatcher looks like a good candidate for dependency injection with $depencies and a non static getter and setter. It looks like this is optional? We can omit the implementation config here and add it in CWP config instead.

Let me know what you think

caffeineinc: Yes, configuration is optional, as it's not required by default. CWP requires an overloaded dispatcher to set the CURL props:
```$config = [
            CURLOPT_PROXY => SS_OUTBOUND_PROXY,
            CURLOPT_PROXYPORT => SS_OUTBOUND_PROXY_PORT
        ];```
Without it it can't make requests to youtube / external links. The problem really is you can't overload anything if you don't provide an interface for it, and unless CWP is reimplementing a new asset admin model for Embed, (there's no dependency injection here) there's no _great_ way of doing it. 

This change allows for ```
if(!\SilverStripe\Control\Director::isDev()){
    if (defined('SS_OUTBOUND_PROXY') && defined('SS_OUTBOUND_PROXY_PORT')) {
        $config = [
            CURLOPT_PROXY => SS_OUTBOUND_PROXY,
            CURLOPT_PROXYPORT => SS_OUTBOUND_PROXY_PORT
        ];

        $dispatcher = new \Embed\Http\CurlDispatcher($config);
        EmbedResource::setDispatcher($dispatcher);
    }
}``` which at least gives the option to use a custom configured dispatcher and solves the CWP use case.

I can make the change to PSR / get but do you think you'd ride with this, and make the configure in CWP ?  I always think forcing an implementation `Embed::create($url);` makes it very inflexible. The other option would be using dependency injection for `EmbedResource`. Open to ideas.

robbieaverill:
I've updated the PR to allow a DispatcherInterface to be defined with Injector configuration (YAML)

PR #798, opened 2018-06-27
Install 5.x-dev recipe when testing master.

maxime-rainville:
Parent issue silverstripeltd/open-sourcerers#31


maxime-rainville:
The main problem is with the JS tests that completely fail. But the master branch is already like that. So I'll merge :see_no_evil:

PR #799, opened 2018-07-02
MINOR: Trim URL from media embed form and make the field type URL

lukereative:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/845


maxime-rainville:
I had a look to test. Failure is not related to this change. Merging

PR #801, opened 2018-07-04
MINOR: Correctly store overrides and use better alignment form input

lukereative:
Fixes https://github.com/silverstripe/silverstripe-assets/issues/131
Implements https://github.com/silverstripe/silverstripe-cms/issues/706


lukereative:
I've added a fix for the dimensions not working at all. Should be working fully now.

maxime-rainville:
I don't think the behat test failure is related to what you've done.

PR #802, opened 2018-07-09
MINOR Fix behat tests to work with new dropdown button.

maxime-rainville:


dhensby:
does this need to be back-ported to 1.2? I think these tests fail there too...

maxime-rainville:
Good point. Should have of that.

PR #803, opened 2018-07-10
FIX Making subsite extension illegal to ensure tests pass with Subsites

### LINKED ISSUE #787: Permission test failure when installed with subsites
**Context: CWP 2.0.0, SilverStripe 4.1.1, subsites**

The following test fails when run with subsites installed. Ideally we can isolate this somehow:

* SilverStripe\AssetAdmin\Tests\Controller\AssetAdminTest::testSaveOrPublish

ScopeyNZ:
Resolves #787

### DIFF
--- a/tests/php/Controller/AssetAdminTest.php
+++ b/tests/php/Controller/AssetAdminTest.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\AssetAdmin\Tests\Controller;
 
+use SilverStripe\AssetAdmin\Forms\FolderFormFactory;
 use SilverStripe\AssetAdmin\Tests\Controller\AssetAdminTest\FileExtension;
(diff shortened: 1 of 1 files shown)


tests/php/Controller/AssetAdminTest.php
```diff
@@ -28,6 +29,10 @@ class AssetAdminTest extends FunctionalTest
      */
     protected $session = null;
 
+    protected static $illegal_extensions = [
+        FolderFormFactory::class => ['SilverStripe\Subsites\Extensions\FolderFormFactoryExtension'],
```
robbieaverill: We'd probably prefer to use `FolderFormFactoryExtension::class`, even if it's aliased during import

ScopeyNZ: Sweet as. Forgot that it doesn't kick up a fuss for imports that don't exist. Fixed.

robbieaverill:
Behat failures are pre-existing

Issue #779, opened 2018-04-27
Update CMS/PL Popover CSS to align with popover design


newleeland:
CSS of popovers withing the CMS and Pattern Library are not consistent with design documation. This is because of several issues.
- Update to Bootstrap 4 stable, has caused some CMS bootrap CSS commenting redundant
- Originally Custom code made to Popovers originally, to make them as dropdown menus. e.g. campaigns
- SCSS Variable additions/removal/changes in BS4 Stable.

Popover Design:
https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/components/5ae267cb98330300118a7851

### Related scss files
- [ ] Update bootstrap CSS with latest bootstrap on src/styles/_bootstrap-variable.scss 
```### Update src/styles/_bootstrap-variable.scss
// Removed legacy commented variables which shouldn’t be used anymore ($popover-inner-padding, …
$popover-arrow-outer-width:         ($popover-arrow-width + 1px) !default;
$popover-inner-padding:             1px !default;

// Popovers (with //changes)

$popover-font-size:                 $font-size-base !default;  // SS custom
$popover-bg:                        $white !default;
$popover-max-width:                 276px !default;
$popover-border-width:              $border-width !default;
$popover-border-color:              rgba($black, .2) !default; // SS custom
$popover-border-radius:             $border-radius-lg !default;
$popover-box-shadow:                0 .25rem .5rem rgba($black, .2) !default; // SS custom

$popover-header-bg:                 darken($popover-bg, 3%) !default; // SS custom
$popover-header-color:              $headings-color !default;
$popover-header-padding-y:          .5rem !default; // SS custom
$popover-header-padding-x:          .75rem !default; // SS custom

$popover-body-color:                $body-color !default;
$popover-body-padding-y:            $popover-header-padding-y !default;
$popover-body-padding-x:            $popover-header-padding-x !default;

$popover-arrow-width:               1rem !default; // SS custom
$popover-arrow-height:              .5rem !default; // SS custom
$popover-arrow-color:               $popover-bg !default;

$popover-arrow-outer-color:         fade-in($popover-border-color, .05) !default;
```

- [ ] Potentially removing custom (hacky) code for component/PopoverField/PopoverField.scss in favour of using vanila bootstrap 4. There will need to be investigation into where this code is being used. First impression look like legacy code from BS4a where popover is being used as dropdown menu. Suggest replace with BS dropdown.
[...]

PR #804, opened 2018-07-10
Add legacy popover styles for asset-admin more actions menu

robbieaverill:
This will be removed in favour of a dropdown menu at some point.

I don't have enough time at the moment to replace the use of popovers in asset-admin with dropdowns right now, so this introduces a non-breaking patch for it in the meantime.

Requires https://github.com/silverstripe/silverstripe-admin/pull/565

Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/779


dhensby:
Test failure is pre-existing

Issue #782, opened 2018-05-09
BUG: Freshly uploaded files don't receive optimistic updates


unclecheese:
### Steps to reproduce
* Upload a file
* Select it
* Choose "publish" in the batch actions
* See that the red dot, indicating DRAFT is not removed

Suspect this is because the uploading process circumvents Apollo, making it invisible to the cache.

brynwhyman:
I can't recreate this in version `1.4.x-dev`. 

![publish-file](https://user-images.githubusercontent.com/15868440/67167936-46b33000-f3fb-11e9-8f00-21b04b91c43f.gif)

Issue #781, opened 2018-05-09
Leverage new Apollo 2.x APIs


unclecheese:
Once https://github.com/silverstripe/silverstripe-asset-admin/pull/780 is merged, we need to start taking advantage of the new [imperative cache API](https://www.apollographql.com/docs/react/advanced/caching.html). This will go a long way to improving the user experience and overall performance of the asset-admin module.

## Tasks
- [ ] `moveFilesMutation` needs to use cache read/writes
- [ ] `createFolderMutation` needs to use cache read/writes
- [ ]  Remove imperative (and unnecessary) `readFiles` call in `handleMoveFilesSuccess`, and instead rely on optimistic updates via cache
- [ ] Remove imperative (and unnecessary) `handleBrowse` call in `handleCreateFolder` and instead rely on optimistic updates via cache

## More information
See `deleteFilesMutation` in the above PR as an example of how this will work.

PR #808, opened 2018-07-12
MINOR Open file details after the last file has been uploaded

### LINKED ISSUE #489: Insert Media > Upload File doesn't get auto-selected
One of the criteria mentioned by @sminnee for the new Insert Media dialog in #173, was that after an image as uploaded, it would automatically be selected to allow the user to then click the Insert button.

But this doesn't appear to be the case currently. Perhaps something broke it?

lukereative:
Fixes #489


client/src/containers/Gallery/tests/Gallery-test.js
```diff
@@ -309,26 +309,6 @@ describe('Gallery', () => {
       expect(props.actions.queuedFiles.removeQueuedFile).not.toBeCalled();
     });
 
-    it('should openFile if type is "insert-media"', () => {
```
maxime-rainville: Why is this test no longer relevant?

lukereative: Because we decided that it should open the uploaded file on both views (asset-admin and insert media) so we no longer check for that.

lukereative:
@maxime-rainville Turns out the failing test wasn't actually related to my change (it was failing locally for me on `1` branch as well), the test was incorrectly clicking the delete button while uploading which had the same selector as the checkbox element it was trying click, I made the step more specific so it only clicks the checkbox not any action that the file has. However I did modify the behaviour so the details panel doesn't open if you have other files checked and you upload.

maxime-rainville:
Sorry I'm repeating myself :-P

lukereative:
I think that should be fixed as a separate issue, I think there's some todo's in the code and maybe some issues in the backlog that I've seen regarding updating newly uploaded files that might be relevant #781 #782

PR #814, opened 2018-07-27
BUG fixes #769 #756

### LINKED ISSUE #756: Remove "replace image" functionality from the Insert from files modal
In the "Insert from files" modal you have the actions to be able to replace a file although you cannot save the replacement so there is no ability to actually change the replace the image. We should just remove the functionality completely.
pasted_image_3_04_18__10_10_am

### LINKED ISSUE #769: Shouldn't be able to drag a file to the edit panel unless you can replace the image that way
It looks like you can drag one of the files to the edit panel when in fact you cannot drop a file in this location currently.

I think we shouldn't allow the drag functionality in the edit panel at all as this functionality is provided elsewhere and limiting the functionality would be easiest to maintain.

Img of thumb being dragged to the edit panel:
screenshot_17_04_18__11_13_am

The other option is to provide the functionality to drop content there—then we should redesign the "drag-drop" image to sit on top of the whole edit panel and be clearer that its replacing the file, as I said I don't prefer this option.

lukereative:
Fixes #769 
Fixes #756


robbieaverill:
Side note: it'd be great if you could amend the commit to describe what you're changing =)

robbieaverill:
For future reference, cow only uses the first line of a commit message in changelogs. This will end up with `Fixes #769 and #756` under the bugfixes section, which doesn't really describe anything. Better to put these at the end of the message instead so Git users can read it

unclecheese:
Thanks, @robbieaverill .. I'll update the commit message before merge.

unclecheese:
Updated commit message, got tests passing. Good to merge.

PR #824, opened 2018-08-22
WIP: FIX Existing URL query strings should be preserved

ScopeyNZ:
Currently when adding an upload field to the CMS fields of a versioned object it's possible that the "createFileEndpoint" may be provided with URL query params - eg. from `VersionedStateExtension`. This PR makes sure any existing query is preserved (and prevents things like `/my/base/link?stage=Stage/2`).


client/src/containers/Editor/Editor.js
```diff
@@ -147,12 +147,21 @@ class Editor extends Component {
...
+    const { targetId, schemaUrlQueries } = this.props;
+    let { schemaUrl } = this.props;
+    let urlQueryString = schemaUrlQueries
       .map(query => `${query.name}=${query.value}`)
       .join('&')
       .trim();
+
+    if (schemaUrl.indexOf('?') !== false) {
+      const questionMarkIndex = schemaUrl.indexOf('?');
+      urlQueryString = schemaUrl.substring(questionMarkIndex + 1) + (urlQueryString || '');
```
ScopeyNZ: This should be `+=`

client/src/containers/Editor/Editor.js
```diff
@@ -147,12 +147,21 @@ class Editor extends Component {
   }
 
   render() {
-    let urlQueryString = this.props.schemaUrlQueries
+    const { targetId, schemaUrlQueries } = this.props;
+    let { schemaUrl } = this.props;
```
robbieaverill: You could combine these two and alias schemaUrl so it can be mutated, e.g.

```js
const { targetId, schemaUrlQueries, schemaUrl: baseSchemaUrl } = this.props;
```

Then mutate baseSchemaUrl instead

ScopeyNZ: `const` though.

robbieaverill: Lol yeah of course

ScopeyNZ:
I think this problem isn't limited to asset admin - as I'm having a very similar issue with recent history viewer work in elemental. Perhaps this is not the right place to fix this. I'm going to mark this as WIP until I figure out what's going on.

robbieaverill:
Yeah, just reproduced on elemental as well. I think this could be a problem solved outside of this module

robbieaverill:
It's worth noting that the backend counterpart for this code isn't really expecting to return any query strings, so we could probably just strip them off and ignore them. It'd be much better to return data in the schema body than in the URL for it

robbieaverill:
Related: https://github.com/silverstripe/silverstripe-admin/issues/607

blueo:
FWIW I'm experiencing this behaviour when using Better Navigator. It provides an 'edit in CMS' button that appends a stage param to it's link (eg `admin/edit/show/2?stage=Stage` when then creates the invalid URL above. There seem to be a number of places where the JS isn't checking/expecting URL params. Perhaps it would be good to have a standard JS lib for handling urls and preserving/ignoring params

ScopeyNZ:
This should be fixed by https://github.com/silverstripe/silverstripe-versioned/pull/173 now.

PR #833, opened 2018-09-11
NEW TinyMCE inline toolbar for images and embeds

lukereative:
Resolves silverstripe/silverstripe-cms#455

**Related PR** https://github.com/silverstripe/silverstripe-admin/pull/637


lukereative:
I always forget `ENH` isn't one, is it possible we could include it in future? I've changed it to `NEW` anyway.

robbieaverill:
Yeah man, PR a new piece into that regex =D

lukereative:
Done – https://github.com/silverstripe/cow/pull/109

PR #842, opened 2018-09-24
ensure ID is set after upload

lekoala:
otherwise, getObjectFromData will inspect owners and fail when calling inferReciprocalComponent (which needs an ID)

### DIFF
--- a/code/Forms/UploadField.php
+++ b/code/Forms/UploadField.php
@@ -157,6 +157,11 @@ class UploadField extends FormField implements FileHandleField
             return (new HTTPResponse(json_encode($result), 400))
                 ->addHeader('Content-Type', 'application/json');
         }
+        
+        // We need an ID for getObjectFromData
+        if(!$file->ID) {
+            $file->write();
+        }
 
         // Return success response
         $result = [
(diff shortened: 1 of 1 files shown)


code/Forms/UploadField.php
```diff
@@ -157,6 +157,11 @@ public function upload(HTTPRequest $request)
             return (new HTTPResponse(json_encode($result), 400))
                 ->addHeader('Content-Type', 'application/json');
         }
+        
+        // We need an ID for getObjectFromData
+        if(!$file->ID) {
```
robbieaverill: You'll need to fix the PSR-2 violation here =)

code/Forms/UploadField.php
```diff
@@ -157,6 +157,11 @@ public function upload(HTTPRequest $request)
             return (new HTTPResponse(json_encode($result), 400))
                 ->addHeader('Content-Type', 'application/json');
         }
+        
+        // We need an ID for getObjectFromData
+        if (!$file->ID) {
```
robbieaverill: Actually, maybe `$file->isInDB()` might be better to wrap this up instead?

robbieaverill:
Behat broke in #833, so merging anyway

PR #843, opened 2018-10-01
BUG: Fix outdated data in Apollo GraphQL cache when deleting/moving files

### LINKED ISSUE #809: State gets confused after moving, deleting, uploading files.
Asset-admin has an annoying tendency to mix up its state when doing file operation. Which leads to ghost file entries.

Here's a couple examples:
* When deleting a newly added file, it's still visible https://www.youtube.com/watch?v=c8ZVXdeHKLY&feature=youtu.be
* When moving a file from one folder to a different one, it's not visible in the new folder. https://www.youtube.com/watch?v=_VEV7GCnVz8

# PR's

- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/843

bergice:
Apollo GraphQL does not currently have a way to invalidate cache without clearing the whole store

Fixes #809


client/src/state/files/moveFilesMutation.js
```diff
@@ -22,6 +22,9 @@ const config = {
             folderId,
             fileIds,
           },
+          update: () => {
+            window.ss.apolloClient.resetStore();
```
ScopeyNZ: I think this might be too heavy handed. Even if it's not for asset-admin, as soon as we have some routing in the CMS then you're going to be resetting the store of completely unrelated parts of the CMS. 

Even for asset admin I think it makes sense to keep stored query results for individual files, as the sort order isn't really exposed within that query?

I think you just want to reset the store for all `readFiles` queries.

bergice: Apollo GraphQL currently doesn't support invalidating whole queries as it depends on the exact variable set provided. Based on the research I did, this is something they're prioritising at the moment.

unclecheese: Yup, unfortunately, this is the best worst solution. More info: https://github.com/apollographql/apollo-client/issues/621

unclecheese: @bergice Can you just leave a comment  on that `update` callback explaining the hack and that we need to revisit it once Apollo supports cache invalidation? It's basically technical debt, so we need to have some record of it.

PR #848, opened 2018-10-11
NEW Move the replace file into the more options action set

### LINKED ISSUE #757: Move the "Replace file" action to the "More options" action set
I would like to suggest we move the "Replace" file action to the more options action area as the placement doesn't blend in well with the UI and the iconography is misleading (I've experimented with other icons but without improvement). The action really needs text to explain what the action is and doesn't feel like it should have such a prominent placement. 

The "cancel" option which appears after the replacement has taken place can move to the notification, we don't need the magnifying icon as it is already on the image when hovered (even though its not visible to touch devices users can still click on the image to view the full version). We should be able to provide a quick mockup to illustrate the interaction and I'm open to any other ideas instead.

Existing view:
image

sachajudd:
Resolves #757


client/src/components/AssetDropzone/AssetDropzone.js
```diff
@@ -70,6 +70,15 @@ class AssetDropzone extends Component {
     }
   }
 
+  componentDidUpdate() {
+    // Reattach name to hiddenFileInput as dropzone recreates this element after each upload
+    const name = this.props.name;
```
robbieaverill: We'd prefer this to be destructured: `const { name } = this.props;`

client/src/containers/Editor/Editor.js
```diff
@@ -115,6 +121,13 @@ class Editor extends Component {
     });
   }
 
+  replaceFile() {
+    const hiddenFileInput = document.querySelector('.dz-input-PreviewImage');
+
+    // Trigger a click on Dropzone's hidden file input in order to upload an image
+    hiddenFileInput.click();
```
robbieaverill: It's probably worth checking that `hiddenFileInput` is resolved before calling click on it

code/Forms/FileFormFactory.php
```diff
@@ -382,6 +382,20 @@ protected function getUnpublishAction($record)
...
 
+    /**
+     * Get Replace file action
+     *
+     * @param File $record
+     * @return FormAction
+     */
+    protected function getReplaceFileAction($record)
+    {
+        $action = FormAction::create('replacefile', 'Replace file')
```
robbieaverill: Can you please make the field title translatable?

client/src/components/AssetDropzone/AssetDropzone.js
```diff
@@ -70,6 +70,15 @@ class AssetDropzone extends Component {
...
+
+    if (name) {
+      this.dropzone.hiddenFileInput.classList.add(`dz-input-${name}`);
```
robbieaverill: This isn't very "Reacty" - is this element controlled by something outside of the React components? If it's rendered by React then we should control it with React really. If it's rendered by something else then I guess it's OK, but would be good to justify it

robbieaverill:
Also looks the tests need some love =)

unclecheese:
Would be good to get this into the `4.3` release. Looks it's just some minor revisions from @sachajudd ? Have optimistically retargeted to `1.3`.

client/src/components/AssetDropzone/AssetDropzone.js
scott1702: Yeah the element is created by the [dropzone](https://www.dropzonejs.com), outside of the React architecture.

sachajudd:
Sounds good @unclecheese, I'll get on to looking at the tests 🙂

unclecheese:
Looking good. I'm going to add a behat test for it.

PR #850, opened 2018-10-19
FIX Use fixtured file title in test assertion

robbieaverill:
Fixes an odd global state issue that surfaces when running the recipe-cms testsuite in the CWP kitchen sink. Not reproducible in isolation. https://travis-ci.org/silverstripe/cwp-recipe-kitchen-sink/jobs/437571397#L1081

@dhensby suggested debugging the DB drivers to see if any rogue DDL commits are made, but TBH I don't have time to do that right now. It's super edge case and this patch fixes it the failing test.


dhensby:
Can we add some kind of todo / issue about this fixture state leak? I know it's difficult as we don't know where it's actually coming from... but it's worth noting at least.

Just a thought: it must be most likely from a test that also uses the fixture file as this one...

Issue #851, opened 2018-10-23
Component provided by UploadField doesn't work when ReactDND exists on the page


ScopeyNZ:
This has been brought to light with recent work on Elemental.

If you have an UploadField on a page that uses ReactDND, opening the file browse modal means that it sits there with the loading indicator forever (and the console has the helpful error: `Cannot have two HTML5 backends at the same time`).

When a page already has React components using the HOC provided by ReactDND to create a `DragDropContext` any additional contexts being added will not work (provided they use the usual `HTML5Backend`).

There's a whole issue on this in the ReactDND repo: https://github.com/react-dnd/react-dnd/issues/186

There's two possible solutions I can think of:

#### Add the ability to disable dragging and dropping on a gallery:

```patch
--- client/src/containers/Gallery/Gallery.js	(revision 902fec0f89bb49783b2dc5f64b4d0c24c33531fb)
+++ client/src/containers/Gallery/Gallery.js	(date 1540333484000)
@@ -819,6 +819,7 @@
       onCreateFolder,
       onOpenFolder,
       onViewChange,
+      canDrag,
     } = this.props;
 
     const props = {
@@ -911,13 +912,15 @@
       cssClasses.push('gallery__main--has-opened-item');
     }
 
+    const GalleryInner = canDrag ? 'GalleryDND' : 'div';
+
     return (
       <div
         className="flexbox-area-grow gallery__outer"
         ref={gallery => { this.gallery = gallery; }}
       >
         {this.renderTransitionBulkActions()}
-        <GalleryDND className={galleryClasses.join(' ')}>
+        <GalleryInner className={galleryClasses.join(' ')}>
           {this.renderToolbar()}
           <SelectableGroup
             enabled={this.props.view === 'tile' && this.props.type === 'admin'}
@@ -947,7 +950,7 @@
               {this.renderGalleryView()}
             </AssetDropzone>
           </SelectableGroup>
-        </GalleryDND>
+        </GalleryInner>
         { this.props.loading && [
           <div key="overlay" className="cms-content-loading-overlay ui-widget-overlay-light" />,
           <div key="spinner" className="cms-content-loading-spinner" />,
```

#### Provide our own HOC that provides a Drag Drop context:

Provided as a solution in the attached issue: https://github.com/react-dnd/react-dnd/issues/186#issuecomment-282789420

I don't know if this will work - I haven't tested it.

This should probably live in admin.

maxime-rainville:
[...]

PR #853, opened 2018-10-25
NEW UploadFields can opt-out of having drag and drop functionality

ScopeyNZ:
See #851

ReactDND does not support having multiple drag and drop zones in a single DOM.

This PR allows opting out of DND - which might need to be done if developers choose to use DND in their own admin components.

Nothing changes by default - calling `->setDragEnabled(false)` on an `UploadField` will prevent drag and drop from being used.


ScopeyNZ:
Closing this issue in favour of an HOC option (coming soon)

PR #887, opened 2018-11-29
MINOR: Use new applyArgsToList function

unclecheese:
Reference: https://github.com/silverstripe/silverstripe-asset-admin/pull/886

Depends on: https://github.com/silverstripe/silverstripe-graphql/pull/191


unclecheese:
On further review, I don't think the change to GraphQL is necessary, and by association, neither is this PR. It serves a pretty clear edge case (no pun intended), and I'm not so sure doing it the way this is implemented in `1.3` is all that bad.

PR #888, opened 2018-12-03
Allow correct number of files to be uploaded with AssetDropzone

### LINKED ISSUE #821: UploadField: doesn't correctly count images when batch uploading
When batch uploading images to an UploadField that has a max file limit, the field return an error stating that it's too many images on an indeterminate number of the images, despite the total number of images being equal to or less than the limit.

```
        $image_field = UploadField::create(
            'Images',
            $product->fieldLabel('Images')
        );
        $image_field->setFolderName($folder)
            ->setAllowedMaxFileNumber(10)
            ->setRightTitle('Limted to 10 items at a time');
```

uploadfield-error

I've tried uploading varying numbers of images from 2 to 20 (one always works fine) and it always uploads less than the 10 image limit.

If i keep re-adding the files then it will upload them normally and stop me once i reach the limit.

# Pull Requests
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/888

bergice:
Just ensuring `maxFiles` stays the same while uploading multiple files.

Fixes #821


code/Forms/UploadField.php
```diff
@@ -321,7 +321,7 @@ public function validate($validator)
         if ($maxFiles < 1 || $count <= $maxFiles) {
             return true;
         }
-        $validator->validationError($this->getName(), _t('', 'Bobby'));
+        $validator->validationError($this->getName(), _t('', "You can only upload $maxFiles file" . ($maxFiles==1?'':'s') . '!'));
```
maxime-rainville: OMG ... can't believe we missed that one.

`_t()` is the SS3 method. We should be using `i18n::_t()` instead. The first parameter needs to be a key that will be targeted in the translation.

You need to parametrise the string, otherwise people won't be able to override the number in other languages. 

The exclamation mark is a bit aggressive. I would use something like "Limited to %s files". @clarkepaul @newleeland do we have a preference for the wording here?

ScopeyNZ: WRT `_t` - I don't think we plan to deprecate support for that function. I know that I've seen a bunch of PRs still using `_t`. In JS it must be `i18n._t` but I believe it's still valid to use just `_t` in PHP.

Technically you would also have to have a plural and a non-plural version of the string too - although I'm not sure how much support we have for that. A quick search only showed subsites.

It seems to me like "1 file" is a common use case so it'd be good to get the right wording.

Issue #891, opened 2018-12-11
UploadFieldItem title attribute on div.uploadfield-item__error-message


ivoba:
When the validator returns an error the UploadField renders a red Error kind of button.
This cuts longer error message for design reasons which is fine.
But an additional title attribute with the error message would make it possible to hover over the message and still read the whole string.

Should not be much of deal, what do you think?

chillu:
Do you have a screenshot of the current state, and maybe a mockup of how you'd fix it?

ivoba:
I extended UploadField with a json-schema validation, error messages are a bit longer now.
Current State:
![bildschirmfoto vom 2018-12-12 09-36-46](https://user-images.githubusercontent.com/471254/49857310-89c37080-fdf2-11e8-902e-7bc911d333bb.png)

Improved State:
![bildschirmfoto vom 2018-12-12 09-39-05](https://user-images.githubusercontent.com/471254/49857341-98aa2300-fdf2-11e8-9d34-392628b6cb28.png)

robbieaverill:
Fixed with #900 - thanks @ivoba!

PR #898, opened 2019-01-07
added title attribute to UploadField error

ivoba:
This should fix https://github.com/silverstripe/silverstripe-asset-admin/issues/891

This is a lucky shot PR :) as it seemed to be an easy adjustment.  
I did not build the client to test it and im also not sure if this is correct in react context.
Any advice welcome.

### DIFF
--- a/client/src/components/UploadField/UploadFieldItem.js
+++ b/client/src/components/UploadField/UploadFieldItem.js
@@ -226,7 +226,7 @@ class UploadFieldItem extends Component {
 
     if (message !== null) {
       return (
-        <div className="uploadfield-item__error-message">
+        <div className="uploadfield-item__error-message" title="{message}">
           {message}
         </div>
       );


robbieaverill:
Assuming the `message` inside that element is truncated via CSS, then this is a good change in my books. You'll need to build the dist files though - `yarn build` in the module (requires node 10 for the `1` branch).

client/src/components/UploadField/UploadFieldItem.js
```diff
@@ -226,7 +226,7 @@ class UploadFieldItem extends Component {
 
     if (message !== null) {
       return (
-        <div className="uploadfield-item__error-message">
+        <div className="uploadfield-item__error-message" title="{message}">
```
robbieaverill: Actually, I think it needs to be this:
```suggestion
        <div className="uploadfield-item__error-message" title={message}>
```

ivoba:
@robbieaverill i made a new branch for a new PR but seems i am not able to build.
I ran composer install & yarn install & yarn build (on Node 6) but i get a lot of module not found errors like
`Cannot find module 'components/FormAlert/FormAlert' from 'ThumbnailView-test.js'`

I can close this PR and make a new one as described above. Would you mind building the client?

robbieaverill:
@ivoba you will need to install the dev version of the admin module e.g. 1.x-dev, but yeah I'm happy to build it for you!

PR #904, opened 2019-01-21
WIP Remove ImageBackendFactory Injector override from ImageThumbnailHelper

maxime-rainville:
`ImageThumbnailHelper` overrides `ImageBackendFactory` with `InjectionCreator`. This seems to increase the memory usage systematically when resizing big pictures. Not quite sure why, but I think some object hang around when `InjectionCreator`.

This doesn't actually show when calling `memory_get_peak_usage` because the memory is actually used by some process outside of PHP that's responsible for resizing the image.


maxime-rainville:
Oups ... wrong remote.

Issue #862, opened 2018-11-01
UploadField doesn't correctly use form state


ScopeyNZ:
### Version

1.x / 1.3.x / Probably all the versions

### Steps to reproduce

- Add an UploadField to a form
- Load the form through togglable FormBuilderLoader but set the `refetchSchemaOnMount` property to `false`.
- Mount the form, check the upload field is empty.
- Add a file using the form
- Toggle the form so it is unmounted, then toggle it again so it is remounted
- Note the UploadField is empty

In this case, the value prop of the field (provided by redux-form) is actually correct:  

![image](https://user-images.githubusercontent.com/3260989/47883893-b7340a80-de93-11e8-922a-24d719c2c21e.png)

But the files prop is wrong: 

![image](https://user-images.githubusercontent.com/3260989/47883913-c9ae4400-de93-11e8-9024-c3ccc709530d.png)

I assume part of the issue here is the file info is not persisted...

# Pull Requests

* [x] #863

unclecheese:
Does this mean that I could be editing a page, upload a file, leave the page without saving, and then come back to it later to find the uploaded (unsaved) file still there?

ScopeyNZ:
Yes - As long as you have told `FormBuilderLoader` to *not* refetch schema on mount.

PR #922, opened 2019-03-14
FIX Revert 69341e86 (#863)

### LINKED ISSUE #919: Wrong Image in UploadField
In an Element with a has_one Image the Image "cache-hangs" in the CMS, specifically in UploadField. I've added a LiteralField with a Thumbnail and the path beside the UploadField to make the problem better visible. The LiteralField and also the Thumbnail thingy in the Elemental-GF are correct but on the second Element 'BBB' the Thumbnail in UploadField is wrong. It even shows the wrong Image if I click on the edit-eye in the UploadField. Check the short Screencast below.
```
class ElementTextImage extends BaseElement {
  private static $has_one = [
    'Image' => Image::class
  ];
  private static $field_labels = [
    'Image' => 'Bild',
  ];

  private static $owns = [
    'Image'
  ];

  private static $table_name = 'ElementTextImage';

  function getCMSFields() {

    if ($uploadField = $fields->dataFieldByName('Image')) {
      $uploadField->allowedExtensions = array('jpg', 'png');
      $uploadField->setFolderName('TextImageElement');
      $uploadField->setDescription('min. 1400x1800px');
    }
    $fields->addFieldToTab('Root', LiteralField::create('image-debug', $this->Image()->CMSThumbnail() .'<br>'. $this->Image()->URL), 'Image');

    return $fields;
  }
}
```
wrong-image

```
silverstripe/admin                         1.3.2 
silverstripe/asset-admin                   1.3.2 
silverstripe/assets                        1.3.2 
silverstripe/cms                           4.3.2 
silverstripe/config                        1.0.12
silverstripe/framework                     4.3.2 
silverstripe/graphql                       3.0.2 
silverstripe/recipe-cms                    4.3.2 
silverstripe/versioned                     1.3.2 
silverstripe/versioned-admin               1.1.2 
```

ScopeyNZ:
This was causing issues where state was incorrectly persisted across entwine remounts of the same field but in different form contexts. The original issue (#862) was difficult to reproduce after this revert. If there are remaining issues they can be dealt with separately in a less breaking way.

Fixes #919


lerni:
I can confirm this fixes #922 👍

robbieaverill:
Reopen https://github.com/dnadesign/silverstripe-elemental/issues/490 if merged

ScopeyNZ:
> The original issue (#862) was difficult to reproduce after this revert. 

This is a bunch of baloney. It's simple to reproduce 😅 .

PR #927, opened 2019-03-21
FIX UploadField now ensures that file data is copied to redux store of value updates

ScopeyNZ:
This is a follow up to #922 . That PR reverts a previous attempt at fixing #862 . This PR attempts to fix it again by leaving the existing use of the redux action, but adding an additional use of it given that the value of the field is updated without the redux connected "files" state also being updated.

I've tested this in elemental and I've tested using "Featured Image" in the blog module to ensure the bug reported in #919 is not occuring

Closes dnadesign/silverstripe-elemental#490


client/src/components/UploadField/UploadField.js
```diff
@@ -76,6 +76,44 @@ class UploadField extends Component {
...
+      && value.filter(item => !prevValue.includes(item)).length === 0
+    ) {
+      // Then nothing to do
+      return;
+    }
+
+    // Now we need to check if the files array that we currently have suits the value
+    const fileIds = files.map(file => file.id);
+
+    // This is a similar condition to above, just checking the files array rather than the previous
```
robbieaverill:
```suggestion
    // This is a similar condition to above, just check the files array rather than the previous
```

ScopeyNZ:
Ok it wasn't working when we had redux storing form state while the form fields weren't mounted - ie. unsaved changes on forms that were previously in the DOM but are removed when they're hidden.

I've had to go back to the original "fix" - where we can't let the upload field blindly reset the field to it's original load state. The "initial" state will always be what we got from the server, but the very nature of this issue is that the user has already updated state at some point - the changes are unsaved. I've been more careful with my condition this time. If the existing redux state of the files within the field (`props.files`) exactly matches the IDs of the value of the field (`props.value.Files`) coming from `redux-form` then it will skip the redux update of files.

Issue #893, opened 2018-12-12
Folders sorting incorrectly across pagination


scott1702:
## Overview

The fix here https://github.com/silverstripe/silverstripe-asset-admin/pull/844 caused another issue where folders will only appear at the top of the page they're on, not across all pages. E.g. if I have 50 pages of files/folders and I have a folder called 'Uploads', this folder may not appear until page ~45. This has a fairly big impact on the UX of asset-admin as content authors will struggle to find their folders.

## Acceptance Criteria

The below is documenting what already should be there before the regression was introduced.

 * [x] Shows all folders first in default sort
 * [x] Paginates across folders and files
 * [x] Shows folders visually separate from files if both types are on the same page
 * [x] Works with fluent
 * [x] Doesn't show folders first when changing sort order to "recently edited")

## PRs
 - [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/936

robbieaverill:
So I understand that folders used to be shown first, before files, but now that has regressed and is no longer the case. This means that if you have one folder (Uploads) and 45 pages of files, it won't show up until page 45.

If that's the case, I'll triage as a high impact bug/regression

chillu:
There's two ways we could do this. Examples with 50 folders and 10 files

1. Fix the regression

```
page 1: folder 1-20
page 2: folder 21-40
page 3: folder 41-50, files 1-10
```

2. Paul's alternative suggestion for new UI:

```
page 1: folder 1-10, files 1-10
page 1*: folder 1-50, files 1-10
```

We'd need either loading of all folders (cap at 100?), or we need infinite scrolling.

tomaszpirc:
Both solutions have serious drawbacks when handling large assets libraries. Maybe the most intuitive solution would be using for folders the same Treeview that is used for Pages. This would:
- bring more coherence into both sections;
- allow accessing and listing between folders separately from files (avoid both awkward situations from these proposals);
- offer users a familiar interface (used by Wins and OS)

What do you think?

dnsl48:
In my opinion any option will heavily depend on a project workflow.
I believe we have to revert the regression and bring back the original (4.0) behaviour first, then we could make a step back, rethink this, then gradually introduce new functionality or postpone the change until SS5.

clarkepaul:
[...]

Issue #820, opened 2018-08-13
Folder sort breaks existing group sorts (eg Fluent Localisation sorts)


blueo:
Currently the `FolderTypeCreator` sorts assets by folder first (line 163), in doing so it first saves the current order-by list then clears all other sorts to add a group sort by folder. However, if there is an existing grouped sort, such as as one set by the fluent module (https://github.com/tractorcow/silverstripe-fluent/issues/208), then this will get the same index (eg both sorts get `_SortColumn0` as their identifier and as a result, only the folder sort remains.

To reproduce:
install the fluent module and add the `FluentVersionedExtension` to `File`, make sure that 'Name' is in the localised fields. This will mean the default sort on Name will become a group sort. This leads to unstable sorting as the 'Name' sort is removed when the folder sort is applied.

tractorcow:
If we change folder connector sort to FQN `"File"."ClassName"` instead of just `"ClassName"` maybe that'll fix it? Fluent is sensitive about FQN column.

blueo:
Unless it is changed from a group filter I think the problem will still exist. The issue is that all filters are removed by the sort call and re-added. If a group query is included it is moved into the column select and given an index (SQLSelect.php line 360). So removing and adding messes up the index count and you get two `_SortColumn0` criteria.

PR #935, opened 2019-04-17
FEATURE folders always go first when ordering

dnsl48:
Fix for https://github.com/silverstripe/silverstripe-asset-admin/issues/893
Also fixes the original fluent issue (https://github.com/silverstripe/silverstripe-asset-admin/issues/820)


robbieaverill:
I kind of feel like this should be a patch change against 1.2

dnsl48:
Yeah, right I guess. I'll check if it works with 1.2.

dnsl48:
Done in https://github.com/silverstripe/silverstripe-asset-admin/pull/936

PR #936, opened 2019-04-17
FIX folders always go first when ordering

dnsl48:
Fix for https://github.com/silverstripe/silverstripe-asset-admin/issues/893
Also fixes the original fluent issue (https://github.com/silverstripe/silverstripe-asset-admin/issues/820)

## Notes
 - tested on silverstripe/installer 4.x-dev
 - tested on silverstripe/installer 4.2.x-dev
 - tested on cwp/cwp-recipe-kitchen-sink ^2

### DIFF
--- a/code/GraphQL/FolderTypeCreator.php
+++ b/code/GraphQL/FolderTypeCreator.php
@@ -166,21 +166,16 @@ class FolderTypeCreator extends FileTypeCreator
                 $existingOrderBys = [];
                 foreach ($query->getOrderBy() as $field => $direction) {
                     if (strpos($field, '.') === false) {
-                        // some fields may be surrogates added by extending augmentSQL (e.g. fluent)
+                        // some fields may be surrogates added by extending augmentSQL
                         // we have to preserve those expressions rather than auto-generated names
-                        // that SQLSelect::addOrderBy leaves for them (usually that's alike _SortColumn0)
-                        //
-                        // see related issues for more details:
-                        //  - https://github.com/silverstripe/silverstripe-asset-admin/issues/820
-                        //  - https://github.com/silverstripe/silverstripe-asset-admin/issues/893
+                        // that SQLSelect::addOrderBy leaves for them (e.g. _SortColumn0)
                         $field = $query->expressionForField(trim($field, '"')) ?: $field;
                     }
 
                     $existingOrderBys[$field] = $direction;
                 }
 
-                // Folders should always go first due to backwards compatibility
-                // See https://github.com/silverstripe/silverstripe-asset-admin/issues/893
+                // Folders should always go first
                 $dataQuery->sort(
                     sprintf(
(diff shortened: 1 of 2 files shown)


robbieaverill:
Minor point that I think this is a bug fix not a feature re: the commit prefix

dnsl48:
> Minor point that I think this is a bug fix not a feature re: the commit prefix
true. fixed

code/GraphQL/FolderTypeCreator.php
```diff
@@ -158,8 +158,53 @@ public function resolveChildrenConnection(
...
+
+                    $existingOrderBys[$field] = $direction;
+                }
+
+                // Folders should always go first due to backwards compatibility
+                // See https://github.com/silverstripe/silverstripe-asset-admin/issues/893
+                $dataQuery->sort(
+                    sprintf(
+                        '(CASE WHEN "ClassName"=%s THEN 1 ELSE 0 END)',
+                        DB::get_conn()->quoteString('SilverStripe\Assets\Folder')
```
robbieaverill: Do you think it might be worth using `Folder::class` here?

code/GraphQL/FolderTypeCreator.php
```diff
@@ -158,8 +158,53 @@ public function resolveChildrenConnection(
...
+                    $dataQuery->sort($field, $dir, false);
+                }
+
+                return $dataQuery;
+            });
+        }
+
         // Filter by permission
-        $ids = $list->column('ID');
+        // $ids = $list->column('ID');
```
robbieaverill: If this isn't needed maybe you could remove it

code/GraphQL/FolderTypeCreator.php
dnsl48: right. Just copy-pasted this bit from the old version.
fixed.

code/GraphQL/FolderTypeCreator.php
dnsl48: done

code/GraphQL/FolderTypeCreator.php
```diff
@@ -158,8 +158,52 @@ public function resolveChildrenConnection(
...
+                foreach ($query->getOrderBy() as $field => $direction) {
+                    if (strpos($field, '.') === false) {
+                        // some fields may be surrogates added by extending augmentSQL (e.g. fluent)
+                        // we have to preserve those expressions rather than auto-generated names
+                        // that SQLSelect::addOrderBy leaves for them (usually that's alike _SortColumn0)
+                        //
+                        // see related issues for more details:
+                        //  - https://github.com/silverstripe/silverstripe-asset-admin/issues/820
+                        //  - https://github.com/silverstripe/silverstripe-asset-admin/issues/893
+                        $field = $query->expressionForField(trim($field, '"')) ?: $field;
```
unclecheese: One of these issues is closed and the other is the one that is resolved by this PR. I'm going to remove these links from the comments.

PR #941, opened 2019-05-07
BUG Update RemoteFileFormFactoryTest to rely on translation codes rather than actual translation

maxime-rainville:
This only fixes the unit tests, not the behat failure.

Basically, we were target specific translation. I updated the test to rely on the actual translation code, so future translation change won't break the build.

### DIFF
--- a/tests/php/Forms/RemoteFileFormFactoryTest.php
+++ b/tests/php/Forms/RemoteFileFormFactoryTest.php
@@ -98,43 +98,43 @@ class RemoteFileFormFactoryTest extends SapphireTest
         return [
             [
                 ['fileurl_scheme_blacklist' => ['https']],
-                'https://www.google.com', 'This file scheme is not included in the whitelist',
+                'https://www.google.com'
             ],
             [
                 ['fileurl_scheme_whitelist' => ['http']],
-                'https://www.google.com', 'This file scheme is not included in the whitelist',
+                'https://www.google.com'
             ],
             [
                 ['fileurl_domain_blacklist' => ['www.amazon.com']],
-                'http://www.amazon.com', 'This file hostname is not included in the whitelist',
+                'http://www.amazon.com'
             ],
             [
                 ['fileurl_domain_whitelist' => ['www.google.com']],
-                'http://www.amazon.com', 'This file hostname is not included in the whitelist',
+                'http://www.amazon.com'
             ],
             [
                 # ipv4 blacklist
                 ['fileurl_domain_blacklist' => ['127.0.0.1']],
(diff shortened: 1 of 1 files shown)


tests/php/Forms/RemoteFileFormFactoryTest.php
```diff
@@ -145,10 +146,13 @@ public function providerTestRejectedURLs()
...
      */
-    public function testRejectedURLS($config, $rejectedURL, $rejectedMessage)
+    public function testRejectedURLS($config, $rejectedURL, $rejectedMessageCode)
     {
+        // Suppress the warning about us not provding a default translation for _t
+        i18n::config()->set('missing_default_warning', false);
+
         $this->expectException(InvalidUrlException::class);
-        $this->expectExceptionMessage($rejectedMessage);
+        $this->expectExceptionMessage(_t(RemoteFileFormFactory::class . '.' . $rejectedMessageCode));
```
ScopeyNZ: Can you leave in the defaults too? Related https://github.com/silverstripe/silverstripe-framework/pull/8949

dnsl48:
are you targeting `1` or `1.4`?

tests/php/Forms/RemoteFileFormFactoryTest.php
dnsl48: yes, I agree, if we could take out of equation the actual translation functionality that would make this test more _unitish_ :)

dnsl48: though, I don't believe we should leave the default messages in this test. We'll have to refactor RemoteFileFormFactory and put those into the class constants so they are reusable externally. Not sure if that's worth a shot.

robbieaverill: The proposed change here looks like a good stop gap. Ultimately @ScopeyNZ's proposed change in https://github.com/silverstripe/silverstripe-framework/pull/8949 is what we need to do to isolate translations in tests.

dnsl48: I don't see any value in having `$this->expectExceptionMessage` assertion in this test, except if tests actual translations. I reckon we should either remove it, or set `$usesTranslations = true;` in this test case.

ScopeyNZ: The only reason that I ask we maintain the default is that it makes it easier to overload the translation functionality. Having `_t` in tests doesn't feel right, and having translated exception messages also feels strange.

I agree with removing the message assertion. `$usesTranslations` doesn't exist yet so there's no point setting that.

unclecheese:
Brilliant. I love this fix.

PR #946, opened 2019-05-23
Enable ts.

maxime-rainville:


client/src/containers/ThumbnailView/ThumbnailView.tsx
```diff
@@ -1,7 +1,7 @@
 import i18n from 'i18n';
 import React, { Component } from 'react';
 import { inject } from 'lib/Injector';
```
maxime-rainville: This is what we are trying to fix

ScopeyNZ: Can you give more detail? Does it not like the weird webpack globals? I've been thinking about doing some fairly "simple" work to remove this annoying global and move injector to `@silverstripe/react-injector` - the import would become a valid `import { inject } from '@silverstripe/react-injector';`.

maxime-rainville: Sorry ... that was specifically meant for @dnsl48 who had context on the issue.

maxime-rainville: Basically, TS spit the dummy when you're trying to import a dependency expose from admin with the `expose-loader`.

That specific line throws that error.

maxime-rainville:
We had discussion about this internally and decided it wasn't achievable in the current release line.

Issue #929, opened 2019-04-10
New images disappear from files area 


clarkepaul:
I would say this issue has been around for some time. 
Steps to reproduce:
1. Upload some files to the files area
2. Navigate to a folder
3. and then navigate back to where you uploaded your images
4. The files are no longer visible
5. Refresh the CMS so they appear again

I presume we can detect if there are new images which can be shown here, a user would be very surprised the files disappear as was I.

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/949

PR #949, opened 2019-05-31
BUGFIX: Force refresh of files on upload when file is in edit view

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-asset-admin/issues/929


robbieaverill:
I can't reproduce the original problem, but I've tested this out and the assets area is navigating and uploading correctly

ScopeyNZ:
I can still reproduce this with folders. Steps:

- Add a folder
- Go into that folder
- Press the "go up a folder" button

This one seems like it's redux not being updated?

bergice:
Merged, failing test is unrelated.

Issue #852, opened 2018-10-24
Deleting assets triggers javascript error


jakxnz:
# Requirements

```
"silverstripe/recipe-plugin": "^1",
"silverstripe/recipe-cms": "4.3.x-dev",
"silverstripe/asset-admin": "1.3.x-dev",
```

```
silverstripe/admin                                     1.3.x-dev 4a81b53
silverstripe/asset-admin                               1.3.x-dev 4600d30
silverstripe/cms                                       4.3.x-dev d44e761
```

# Details

Deleting an asset in the asset admin triggers a javascript error, and mildly interrupts the sequence of states

Steps to recreate

1. Log in to the CMS
2. Browse to the "Files" (assets) section
3. Navigation to an existing asset
4. Click on the asset, to bring up the details/edit form
5. Expand the "Other actions" menu (...) and click the "Delete" button
6. Click "OK" to confirm the "Do you want" prompt

or

4. Check the checkbox, located at the bottom-right of the asset thumbnail
5. Click the "Delete" button, located at the bottom of the viewport
6. Click "OK" to confirm the "Do you want" prompt

Expected result:

The image is removed from view (and deleted from the filesystem in the background). A confirmation notice reads "1 folders/files were successfully archived."

Actual result:

The image remains in view, prompting a javascript error,  (but is deleted from the filesystem in the background). A confirmation notice reads "1 folders/files were successfully archived.". The asset is removed from view within 1-2 seconds.

```
TypeError: Cannot read property 'options' of undefined
    at update (bundle.js:1)
    at vendor.js:1
    at o (vendor.js:1)
    at vendor.js:1
    at t.performTransaction (vendor.js:1)
    at e.markMutationResult (vendor.js:1)
    at Object.next (vendor.js:1)
    at f (vendor.js:1)
    at h (vendor.js:1)
    at e.value (vendor.js:1)
```

## Verbose

```
TypeError: "a.config is undefined; can't access its "options" property"
update https://.../resources/vendor/silverstripe/asset-admin/client/dist/js/bundle.js:1:160067
markMutationResult https://.../resources/vendor/silverstripe/admin/client/dist/js/vendor.js:1:81625
o https://.../resources/vendor/silverstripe/admin/client/dist/js/vendor.js:1:103141
markMutationResult https://.../resources/vendor/silverstripe/admin/client/dist/js/vendor.js:1:81598
[...]

PR #951, opened 2019-05-31
FIX Reset apollo store when deleting a file to ensure item is removed from the UI

robbieaverill:
The optimistic cache update logic was broken since SilverStripe 4.3.x. While this is not a great workaround, it at least ensures that the functionality works correctly. Optimistic updates can be re introduced again later in a way that is compatible with injectable GraphQL queries.

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/852 and the Behat builds in 1.4 once merged up.

## Dependencies
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/889

## Todo
- [x] Get tests passing (should just be a matter of adding `getApolloHOC()`
- [ ] ~~Separate PR for the 1.4 branch using node 10.~~


unclecheese:
After some thought, there's a way to accomplish this without nuking the optimistic update. It requires a slight change to the `apolloGraphQLContainer` where the factory is actually a proxy object that allows the user to consume either the HOC or the AST (both injected).

I've added notes to the OP.

robbieaverill:
Thanks for your work @unclecheese, I'll come back to it this week if you don't beat me to it =)

unclecheese:
This is good to go. As mentioned on the dependent PR, we'll worry about the separate node 10 build on merge up. There will be conflicts in `dist/` anyway.

chillu:
Moving this to blocked, since @unclecheese wasn't comfortable putting this into the pending 4.4.1 release (it can go into a 4.4.2 soon after)

Issue #902, opened 2019-01-18
Image of folder while being dragged into another folder looks broken


newleeland:
The folder icon should not scale down to a mini icon when dragged, we can leave it at the standard size.
![Screen Shot 2019-01-18 at 4.05.24 PM.png](https://images.zenhubusercontent.com/596d365561b4ca3b0a74bd86/5472d666-f823-4032-827a-d7052bbb6949)

# Pull Request
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/952

maxime-rainville:
PR has been merged. Looks like this issue has been fixed. If it's not completely fixed, please reopen.

PR #952, opened 2019-05-31
Fix image being small when dragging a folder

bergice:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/902


robbieaverill:
Could you are a screenshot or a little screen capture to show this in action?

bergice:
@robbieaverill 

image

newleeland:
> @robbieaverill
> 
> image

Hey @bergice. Not quite sure how large you are recommending the dragged item should be. But we intentionally made the dragged item to be smaller than the original item. This was to ensure the items in the background are visible so that the user can see where they are dragging into. 

This was the original design:
Screen Shot 2019-06-04 at 10 45 53 AM

bergice:
@newleeland the dragged item size has not changed, just the icon inside it.

ScopeyNZ:
(Pre-existing behat failures)

PR #956, opened 2019-06-14
BUG: Deselect previous item selection when dragging to select items

bergice:
This fixes the following:

1. Select some files/folders
2. Drag to select some files/folders
3. Notice that the previous selection is still selected


client/src/containers/Gallery/Gallery.js
```diff
@@ -512,6 +512,8 @@ class Gallery extends Component {
    * @param event Event
    */
   handleGroupSelect(items, event) {
+    this.props.actions.gallery.deselectFiles();
```
robbieaverill: Any reason you haven't included this in the destructuring on the next line?

bergice: What do you mean?

bergice: Just realised this should work with the concatenation logic below.

robbieaverill: For reference, I meant add it to `const { foo, bar, deselectFiles } = this.props;` then call `deselectFiles()` directly, the destructuring from `this.props.actions.gallery` is already on the next line, so you'd just be adding another item to it

Issue #930, opened 2019-04-15
Null size images are wrongly inlined in GraphQL thumbnail responses


chillu:
When images return a null/zero size during thumbnail generation, they are wrongly inlined into the response, with the assummption that they're small enough for thsi. Since the size couldn't be computed, that's not the case, leading to massive asset admin GraphQL response payloads.

Original report and investigation from @frankmullenger

The responses generated for displaying the image metadata are massive (1-2MB). The reason the responses are massive is because thumbnail images are being inlined into the response:

Probably here:
https://github.com/silverstripe/silverstripe-asset-admin/blob/b4626ed2e629d4ec7c103751ab0c0f6f7d5e438d/code/Controller/AssetAdmin.php#L1221

Which calls here:
https://github.com/silverstripe/silverstripe-asset-admin/blob/5f0bacd05b9319b41c9cc972654a8762c2249d6f/code/Model/ThumbnailGenerator.php#L134

Because of setting here (which shouldn’t be used for large images) and means that protected image thumbnails are inlined:
https://github.com/silverstripe/silverstripe-asset-admin/blob/5f0bacd05b9319b41c9cc972654a8762c2249d6f/code/Model/ThumbnailGenerator.php#L58

## Pull Requests
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/964

unclecheese:
This seems to be a pretty key part of the problem here:

https://github.com/silverstripe/silverstripe-asset-admin/blob/5f0bacd05b9319b41c9cc972654a8762c2249d6f/code/Model/ThumbnailGenerator.php#L116

That always evaluates `false`. The negation is tripping up the way that evaluates. Should use a `<`.

chillu:
🤦‍♂

PR #964, opened 2019-06-28
ENHANCEMENT: Large thumbnails fallback on URL

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-asset-admin/issues/930

This PR is quite verbose, in the interest of preserving public APIs.

* New field resolvers for thumbnails (static, to be compatible with future GraphQL release with schema caching)
* `generateThumbnailLink()` now takes a `$graceful` parameter, backward compatible, which allows the thumbnail result to fallback on URL if too large

Separate PR for the master branch (2.0) will simply change the return value of the `generateThumbnailLink()` method.

### DIFF
--- a/code/GraphQL/FileTypeCreator.php
+++ b/code/GraphQL/FileTypeCreator.php
@@ -9,6 +9,8 @@ use SilverStripe\AssetAdmin\Forms\UploadField;
 use SilverStripe\AssetAdmin\Model\ThumbnailGenerator;
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Folder;
+use SilverStripe\Assets\Storage\AssetContainer;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\GraphQL\Manager;
 use SilverStripe\GraphQL\TypeCreator;
 use SilverStripe\GraphQL\Util\CaseInsensitiveFieldAccessor;
@@ -100,9 +102,11 @@ class FileTypeCreator extends TypeCreator
             ],
             'thumbnail' => [
                 'type' => Type::string(),
+                'resolve' => [static::class, 'resolveThumbnailFieldGraceful'],
             ],
             'smallThumbnail' => [
                 'type' => Type::string(),
+                'resolve' => [static::class, 'resolveSmallThumbnailFieldGraceful'],
             ],
             'width' => [
                 'type' => Type::int(),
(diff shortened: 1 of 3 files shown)


code/GraphQL/FileTypeCreator.php
```diff
@@ -261,4 +265,43 @@ public function setThumbnailGenerator(ThumbnailGenerator $generator)
...
+     * @deprecated 2.0
+     */
+    public static function resolveSmallThumbnailFieldGraceful(AssetContainer $object, $args, $context, $info): string
+    {
+        Deprecation::notice('2.0', 'Use resolveSmallThumbnailField');
+        $width = UploadField::config()->uninherited('thumbnail_width');
+        $height = UploadField::config()->uninherited('thumbnail_height');
+        return static::singleton()
+            ->getThumbnailGenerator()
+            ->generateThumbnailLink($object, $width, $height, true);
```
robbieaverill: I believe this could return null via `File::getURL()` - without strict typing it probably doesn't matter, but maybe worth casting as a string just in case

code/GraphQL/FileTypeCreator.php
```diff
@@ -261,4 +265,43 @@ public function setThumbnailGenerator(ThumbnailGenerator $generator)
...
         return $this;
     }
+
+    /**
+     * @param AssetContainer $object
+     * @param $args
+     * @param $context
+     * @param $info
+     * @return string
+     * @deprecated 2.0
```
robbieaverill: Deprecations should be from when the notice should be raised, e.g. `@deprecation 1.5.0` (it should be a version number). Same for line 279 =)

unclecheese: I think we actually want it deprecated in 2.0 and removed in 2.1. Otherwise, we're going to be calling a deprecated method internally right?

code/GraphQL/FileTypeCreator.php
unclecheese: Ah, yup, good catch.

code/GraphQL/FileTypeCreator.php
robbieaverill: You won't be able to remove it in a minor release, but if you want to deprecate in 2.0 then this line needs to be `@deprecated 2.0.0` =)

code/GraphQL/FileTypeCreator.php
```diff
@@ -261,4 +265,43 @@ public function setThumbnailGenerator(ThumbnailGenerator $generator)
...
+     * @param AssetContainer $object
+     * @param array $args
+     * @param array $context
+     * @param ResolveInfo $info
+     * @return string|null
+     * @deprecated 2.0:2.1
+     */
+    public static function resolveSmallThumbnailFieldGraceful(AssetContainer $object, $args, $context, $info): ?string
+    {
+        Deprecation::notice('2.0', 'Use resolveSmallThumbnailField');
```
maxime-rainville: I'm getting this deprecation warning thrown in my face when running my site in DEV mode

unclecheese: I can't replicate this. The `2.0` should suppress it from anything under `2`, right? Unless you're on `master`?

unclecheese:
Have added a check for null thumbnails, and have moved the Deprecation to the `master` PR. Unfortunately we can't get rid of those methods until `3`.

code/GraphQL/FileTypeCreator.php
maxime-rainville: I was on SS 4.5.x with the kitchen sink. I'll have another go and I'll dig down to see what version 2.0 is being compared to.

maxime-rainville: For my local test, it's comparing "2.0" to "4.0". Not quite sure how this is suppose to work, but it looks like it's a global value, not a module specific one.

PR #965, opened 2019-06-28
ENHANCEMENT: Large thumbnails fallback on URL

unclecheese:
API breaking version of https://github.com/silverstripe/silverstripe-asset-admin/pull/964

### DIFF
--- a/code/GraphQL/FileTypeCreator.php
+++ b/code/GraphQL/FileTypeCreator.php
@@ -263,4 +263,43 @@ class FileTypeCreator extends TypeCreator
         $this->thumbnailGenerator = $generator;
         return $this;
     }
+
+    /**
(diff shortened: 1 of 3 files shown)


emteknetnz:
This pull request hasn't had any activity for a while. Are you going to be doing further work on it, or would you prefer to close it now?

unclecheese:
Needs to stay open. It's a PR to master so review doesn't have to be too picky.

ChloeSartorelli:
This didn't make it in for CMS 5, and there's no branch for CMS 6 yet so it will be closed when the master branch is deleted.
Given that the change was already made in a way that preserves APIs, I don't see the need for this PR - if it is still needed for some reason, an issue should be raised for it which will be appropriately triaged.

Issue #931, opened 2019-04-15
'Files in use' tracking incredibly slow with large numbers of (nested) folders


MasonD:
The `filesInUseCount` field of the `Folder` graphql type eventually ends up calling `nestedFolderIDs`, which is incredibly slow with large numbers of folders (in fact it runs approximately one SQL query per folder).

On a project with ~26000 folders, the vast majority of which live somewher under a single top-level folder, we see response times for any graphql query that fetches this folder of upwards of two minutes. Patching `filesInUseCount` to always immediately return an empty DataList speeds up the slow sections of our asset admin by 500 times.

If it's not feasible to increase the performance of `nestedFolderIDs`, perhaps a configuration option to disable `filesInUseCount` is necessary, so large projects can feasibly use the asset admin. 

# Sub tasks
* [x] Add a state to track confirmation modal on front end
* [x] Display modal to confirm deletion of files
* [x] Create GraphQL end point for retrieving file usage for a list of files
* [x] Fetch file usage when display delete confirmation modal
* [x] Adjust confirmation message based on file usage
* [x] Move the delete file logic and hook it into the confirmation modal
* [x] Display loading indicator while the file usage data is being fetch
* [x] Add unit test to test state reducers
* [x] Add test to confirmation modal component
* [x] Update behat test
* [x] Write some unit test for the new GraphQL endpoint
* [x] Stop fetching fileInUseCount on regular readFiles request
* [ ] ~~Stretch goal: Use toast notifications instead of alert boxes~~
* [x] Apply the same logic to the regular single file delete.

# Pull requests
* [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/999

dnsl48:
Thank you for reporting this and taking time to analyse the problem and provide a good detailed explanation of the problem.

> Perhaps a configuration option to disable filesInUseCount is necessary

Yes, this definitely feels like a non essential feature

> On a project with ~26000 folders ... speeds up the slow sections of our asset admin by 500 times

Do you remember approx numbers in seconds/minutes/hours?

MasonD:
No problem! It was a big issue affecting the usability of our assets since we upgraded to silverstripe 4, so getting it fixed for ourselves ASAP was a priority.

[...]

PR #978, opened 2019-08-16
FIX Optimise AssetAdminFile::nestedFolderIDs

ScopeyNZ:
Attempts to solve #931

Previously the code would fetch children for each folder. Given a structure like this:

```
1
⎿ 1.1
   ⎿ 1.1.1
      ⎿ 1.1.1.1
⎿ 1.2
   ⎿ 1.2.1
   ⎿ 1.2.2
⎿ 1.3
   ⎿ 1.3.1
⎿ 1.4
⎿ 1.5
   ⎿ 1.5.1
```

A query would run to find children of each folder (12 queries). This change means it now fetches folders by "level". So it would be reduced to 4 queries for the above structure.

### DIFF
--- a/code/Controller/AssetAdminFile.php
+++ b/code/Controller/AssetAdminFile.php
@@ -206,20 +206,22 @@ class AssetAdminFile extends DataExtension
     /**
      * Get recursive parent IDs
      *
-     * @param int $parentID
+     * @param int|int[] $parentIDorIDs
      * @param int $maxDepth Hard limit of max depth
      * @return array List of parent IDs, including $parentID
      */
-    public static function nestedFolderIDs($parentID, $maxDepth = 5)
+    public static function nestedFolderIDs($parentIDorIDs, $maxDepth = 5)
     {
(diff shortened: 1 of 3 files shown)


ScopeyNZ:
My own test failed 😖

emteknetnz:
Have tested fix against a dataset with ~7000 nested folders, can confirm the root folder loads  substantially quicker.

Not sure what the original functionality AssetAdminFile::getFilesInUse() actually did though so cannot confirm if there was any change there.

ScopeyNZ:
I wrote the test before I changed the functionality of the method (woo TDD) so hopefully nothing was impacted 😅

ScopeyNZ:
Hmmmn. Tests are passing locally :\

ScopeyNZ:
Ok https://github.com/silverstripe/silverstripe-assets/pull/291 was the culprit :)

NightJar:
Thanks @ScopeyNZ, @emteknetnz  :)

PR #980, opened 2019-08-29
BUG Add a status column to the table view

maxime-rainville:
The Table view didn't have a way to view the status of individual files.

# Parent issue
* https://github.com/silverstripe/silverstripe-assets/issues/286

# Related PRs
* https://github.com/silverstripe/silverstripe-admin/pull/861
* https://github.com/silverstripe/webpack-config/pull/37


client/src/containers/TableView/TableView.js
```diff
@@ -245,6 +256,35 @@ class TableView extends Component {
     );
   }
 
+  /**
+   * Renders the content for the status column
+   *
+   * @param {object} props
+   * @returns {Component|null}
+   */
+  renderStatus(props) {
```
maxime-rainville: This was inspired by [`getStatusFlags`](https://github.com/silverstripe/silverstripe-asset-admin/blob/bc17c1cc5c9febf4761c83b3704b93b05fb7f912/client/src/components/GalleryItem/GalleryItem.js#L191) on `GalleryItem`.

`getStatusFlags` has a `updateStatusFlags` hook that looks like it's meant to allow module like Fluent to add their own status flags. However, it's not clear how you would go about using it or that it even works.

If allowing module to add their own flags is a key requirements, I think we need to rethink how this is done and come up with a consistent way of doing for both the gallery view and the table view.

client/src/containers/TableView/TableView.js
```diff
@@ -4,6 +4,7 @@ import Griddle from 'griddle-react';
 import i18n from 'i18n';
 import { galleryViewPropTypes, galleryViewDefaultProps } from 'containers/Gallery/Gallery';
 import { fileSize } from 'lib/DataFormat';
+import VersionedBadge from 'components/VersionedBadge/VersionedBadge';
```
robbieaverill: It's better to use the `inject` HOC to provide these components as props, that way they're extensible

client/src/containers/TableView/TableView.js
```diff
@@ -245,6 +256,35 @@ class TableView extends Component {
...
+        flags.push({
+          key: 'status-modified',
+          status: 'modified'
+        });
+      }
+    }
+
+    flags = flags.map(({ ...attributes }) => <VersionedBadge {...attributes} />);
+
+    return <span>{flags}</span>;
```
robbieaverill: So this would return an empty span for folders - was that deliberate?

client/src/containers/TableView/TableView.js
maxime-rainville: Switched to use the injector.

client/src/containers/TableView/TableView.js
maxime-rainville: I've updated it to return `null` when there's no flag. There's a possible future where we update this again so other module can specify their own flags and where someone decide they want to flag folders.

bergice:
LGTM, just gotta get the tests green @maxime-rainville

PR #984, opened 2019-09-03
BUG Bust thumbnail cache by appending the vid at to URL in asset admin

### LINKED ISSUE #970: File replace does not show current version
### Steps to reproduce

- Upload a file in CMS
- Publish the file
- Select replace file
- Upload new file
- Save or Publish
- View the file 
- Error - The first uploaded version of the file is shown 
- Error - Last changed date time is not reflected

### Notes:
- Investigate how the thumbnails are being handled too. From the comments:
> If I replace the file and publish it right away, the main file gets published but the thumbnails don't get regenerated. If I first save the replace file and then publish it, the updated thumbnails get published along with the main file.

# Pulls request
* [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/984
* [x] https://github.com/silverstripe/silverstripe-assets/pull/325

maxime-rainville:
Because the live version of an image doesn't change now, when you publish a new version of an existing file, your browser will want to show you the cache version.

I'm getting around that by adding the version ID as a get parameter in Asset-Admin only.

This partially fix #970, although it's not the biggest problem.

# Parent issue
* #970


maxime-rainville:
Note that this change shouldn't affect front end URL. They won't get the vid param.

client/src/components/GalleryItem/tests/GalleryItem-test.js
```diff
@@ -167,6 +167,42 @@ describe('GalleryItem', () => {
...
+        ...props.item,
+        category: 'image',
+        url: 'myUrl',
+        thumbnail: 'data:image/jpeg;base64,0000',
+        version: 123
+      };
+
+      item = ReactTestUtils.renderIntoDocument(
+        <GalleryItem {...props} />
+    );
```
Cheddam:
```suggestion
      );
```

client/src/components/GalleryItem/tests/GalleryItem-test.js
```diff
@@ -167,6 +167,42 @@ describe('GalleryItem', () => {
...
+        ...props.item,
+        category: 'image',
+        url: 'myUrl',
+        thumbnail: 'myThumbnailUrl',
+        version: 123
+      };
+
+      item = ReactTestUtils.renderIntoDocument(
+        <GalleryItem {...props} />
+    );
```
Cheddam:
```suggestion
      );
```

blueo:
just a thought - would LastEdited be safer if say someone didn't want versioned assets?

maxime-rainville:
The sad truth is that there's already a lot of tie in with versioned in asset and asset-admin. Adding a Version ID to the URL is not the straw that will break that camel's back.

client/src/components/GalleryItem/tests/GalleryItem-test.js
maxime-rainville: Fixed.

client/src/components/GalleryItem/tests/GalleryItem-test.js
maxime-rainville: Fixed

Issue #977, opened 2019-08-14
UploadField hard limit of 30s timeout on uploads


HARVS1789UK:
Seen in SilverStripe `4.4.2`

`UploadField` makes use of dropzonejs via `AssetDropzone`.

As per the [dropzonejs config documentation](https://www.dropzonejs.com/#config-timeout), dropzone implements a default timeout of 30 seconds for uploads.

There seems to be no way to change this default timeout through the `UploadField` class, and I am therefore left with a file upload field in the CMS which is not able to upload any files which take > 30 seconds to upload (in my case, videos of circa 100MB and up).

You can replicate this by setting up a fresh `4.4.2` install, adding a `has_many` -> `Files` to a DataObject and then configuring an `UploadField` to accept larger file sizes 
*Note* You will need to raise the default `php.ini`values for:
- `max_input_time`
- `max_execution_time`
- `max_post_size`
- `upload_max_filesize` 

to allow for larger files and longer times.
If your upload speed is so good it is hard to take > 30s to upload a file, I suggest using [Chrome's 'Network throttling` feature](https://css-tricks.com/throttling-the-network/) to simulate a slower connection.

The HTTP request to `upload` will be marked as `(cancelled)` after exactly 30s of processing the upload. This indicates to me that the client has cancelled the request.

The same behaviour can be seen in @kinglozzer's module `bigfork/silverstripe-dropzone` with the default dropzonejs config, but can be resolved by using his methods as below:

```
->setDropzoneConfig([
    'maxFilesize'   => (int)ini_get('upload_max_filesize'),
    'timeout'       => (int)ini_get('max_input_time') * 1000
])
```

It would be good to make these dropzonejs options configurable on `UploadField`, probably via the `AssetDropzone` JS  [referenced here](https://github.com/silverstripe/silverstripe-asset-admin/tree/1/client/src/components/AssetDropzone#options-object---required)

adiwidjaja:
I had the same problem. I am leaving a comment with my server error message for anyone searching for the same error: Partial results are valid but processing is incomplete: [client xxx] mod_fcgid: can't get data from http client.

HARVS1789UK:
@robbieaverill - Is there any roadmap for a fix for this?

[...]

Issue #925, opened 2019-03-21
UploadField max file size validation has no effect for large files


kinglozzer:
Tested with 1.3.3

```php
/** @var Upload_Validator $validator */
$validator = $uploadField->getValidator();
$validator->setAllowedMaxFileSize('1m');
```

I’d expect this to prevent the upload, but it actually allows the file to be uploaded before validation is run server-side. That, while a little frustrating for CMS users who have slow connections, does work for smaller files, however...

If `post_max_size` is set to 10mb, the field is set to 1mb, and the user attempts to upload a 40mb file, they receive a useless error message: a 400 error code and `There has been an error` (or other webserver error message). Error logs show:

> PHP Warning:  POST Content-Length of 40500753 bytes exceeds the limit of 10485760 bytes in Unknown on line 0, referer:

Attempting to upload files via `AssetAdmin` does produce error messages prior to uploading, so client-side validation is obviously possible _somehow_.

### Expected results
- If the set UploadField max file size is exceeded, there is client-side validation alerting the user
- The file is prevented from being uploaded/does not reach the server 

## PRs

 - [x] https://github.com/silverstripe/silverstripe-assets/pull/341
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1017

dnsl48:
Done. Tested in IE11 and Chromium

PR #985, opened 2019-09-03
Remove Default DropzoneJS Timeout of 30s

HARVS1789UK:
Resolves issue 977 (see - https://github.com/silverstripe/silverstripe-asset-admin/issues/977) where file uploads of over 30s have the network request fail (marked as "(cancelled)" in dev tools) due to DropzoneJS having a default client-side timeout of 30 seconds.


ScopeyNZ:
It would be a reasonably "quick win" solution to have this in config. See:

https://github.com/silverstripe/silverstripe-asset-admin/blob/1.2/code/Controller/AssetAdmin.php#L259-L268

These appear to automatically filter down to `AssetDropzone` via the `options` prop. See:

https://github.com/silverstripe/silverstripe-asset-admin/blob/b318caa0f52c3b704b679064499dc434f1c234f1/client/src/containers/Gallery/Gallery.js#L893

and

https://github.com/silverstripe/silverstripe-asset-admin/blob/b318caa0f52c3b704b679064499dc434f1c234f1/client/src/containers/Gallery/Gallery.js#L940

kinglozzer:
@ScopeyNZ Is there an equivalent option for `UploadField`?

ScopeyNZ:
Ah good point. Hmmmn that's kind of annoying.

kinglozzer:
I think we should introduce that to the `UploadField` API though (it’d also solve https://github.com/silverstripe/silverstripe-asset-admin/issues/925), let’s create a separate issue for that

ScopeyNZ:
Are you raising an issue for that or should I?

kinglozzer:
@ScopeyNZ I think you have a better grasp of client-side stuff than me, so go for it 😃

ScopeyNZ:
Done: https://github.com/silverstripe/silverstripe-asset-admin/issues/986

PR #995, opened 2019-09-09
Allow thumbnail generator to skip non-image files before file size check

michalkleiner:
Even though `$generated = $this->generateThumbnails($file);` checks for the file being an image, some files (like large PDFs and/or videos) produce warnings being too large before arriving at that point, unnecessarily polluting the migration log.

This checks for the file type being an image early in the loop.

### DIFF
--- a/code/Helper/ImageThumbnailHelper.php
+++ b/code/Helper/ImageThumbnailHelper.php
@@ -92,10 +92,15 @@ class ImageThumbnailHelper
         $inspectedCount = 0;
         $generatedCount = 0;
         foreach ($files as $file) {
+            // Skip if file is not an image
+            if (!$file->getIsImage()) {
+                $this->logger->debug(sprintf('File is not an image: %s', $file->Filename));
+                continue;
+            }
             // Skip if file is too large
             if ($maxSize > 0 && $file->getAbsoluteSize() > $maxSize) {
                 $this->logger->warning(sprintf(
-                    'File too large for thumbnail for generating thumbnail: %s',
+                    'File too large for generating its thumbnail: %s',
                     $file->Filename
                 ));
                 continue;


michalkleiner:
The Travis failure seems unrelated to the change.

michalkleiner:
@maxime-rainville or @robbieaverill, mind having a look at this small tweak? Thanks

code/Helper/ImageThumbnailHelper.php
```diff
@@ -92,10 +92,15 @@ public function run()
         $inspectedCount = 0;
         $generatedCount = 0;
         foreach ($files as $file) {
+            // Skip if file is not an image
+            if (!$file->getIsImage()) {
```
robbieaverill: Could you change `$files = File::get()` to `Image::get()` instead?

michalkleiner: Technically yes, I was thinking about it, however, if there are any other subclasses of File that are not subclasses of Image and identify as an image using the getIsImage method, this wouldn't match them.

I can update it to that and update some of the debug/warning messages to suit that as well, might be cleaner.

Shall I go ahead?

maxime-rainville: > if there are any other subclasses of File that are not subclasses of Image and identify as an image using the getIsImage method

That's kind of a weird edge case, but it doesn't hurt to cover it. I would be fine with either approach.

michalkleiner:
@maxime-rainville what's your preference in regards to what Robbie commented on? Shall we only target `Image` class in this helper?

maxime-rainville:
I think your solution is a tiny bit safer.

michalkleiner:
Rebased, tests are now all green.

robbieaverill:
Thanks @michalkleiner!

kinglozzer:
@anhld As this was merged into the `1` branch and not `1.4`, it’ll be available in 1.5.x (i.e. SilverStripe 4.5) when that’s released.

michalkleiner:
Uh oh... @kinglozzer would it be fine to open a new PR with the same code against 1.4?

Issue #943, opened 2019-05-15
UX: "Used on" tab with "File Link" records is confusing to users and broken


unclecheese:
### Poor UX issue

![Screenshot 2019-05-15 12 05 13](https://user-images.githubusercontent.com/654636/57740001-d8f72f80-7709-11e9-9bc3-15d909994c03.png)

In the above screenshot, we see a table of the record's title (e.g.`getTitle()`)  and its content type (`singular_name()`). Because file link tracking is done through `many_many_through`, these `FileLink` objects come up as first class content types when they should be invisible to content editors.

An ideal fix would be one that detects these special linking dataobjects and instead presents the user with information about the dataobject that this file is on, rather than the ID of its linking object.

CC @newleeland @clarkepaul

### Broken functionality
Uploading a file, immediately shows it's being used throughout the CMS, when it's not. See Clakepaul's comment below

# Pull Requests
- [ ] ~https://github.com/silverstripe/silverstripe-assets/pull/328~
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/998

robbieaverill:
Related: https://github.com/silverstripe/silverstripe-cms/issues/2428 cc @scott1702

clarkepaul:
I think this issue is getting worse, I uploaded different images without using them on any page and this was the result. Does the amount of links relate to the size of the image and amount of thumbs produced?
![image](https://user-images.githubusercontent.com/555033/59727218-e6fc1b00-9288-11e9-9919-86629a036a21.png)
![image](https://user-images.githubusercontent.com/555033/59727249-04c98000-9289-11e9-9193-085c349401d1.png)

chillu:
This also affects the Bambusa demo, and makes it harder for us to show the "campaigns" feature in a meaningful way

PR #998, opened 2019-09-11
Fix Used On tab showing file links

bergice:
Resolves silverstripe/silverstripe-asset-admin#943

### DIFF
--- a/code/Extensions/UsedOnTableExtension.php
+++ b/code/Extensions/UsedOnTableExtension.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace SilverStripe\AssetAdmin\Extensions;
+
+use SilverStripe\Assets\Shortcodes\FileLink;
+use SilverStripe\Core\Extension;
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\DataObject;
+
+/**
+ * Hides File Links on the "Used On" tab when viewing files
+ */
+class UsedOnTableExtension extends Extension
+{
+    public function updateUsage(ArrayList &$usage, DataObject &$record)
+    {
+        $usage = $usage->exclude('ClassName', FileLink::class);
+    }
+}
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -7,6 +7,9 @@ SilverStripe\Assets\File:
 SilverStripe\Admin\LeftAndMain:
   extensions:
     - SilverStripe\AssetAdmin\Controller\AssetAdminFieldsExtension
+SilverStripe\Admin\Forms\UsedOnTable:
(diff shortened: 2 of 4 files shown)


code/Extensions/UsedOnTableExtension.php
```diff
@@ -0,0 +1,25 @@
...
+use SilverStripe\ORM\DataObject;
+
+/**
+ * Class UsedOnTableExtension
+ *
+ * Hides File Links on the Used On tab when viewing files
+ */
+class UsedOnTableExtension extends Extension
+{
+    public function updateUsage(ArrayList &$usage, DataObject &$record) {
```
maxime-rainville: Maybe use the more generic `SS_List` here. DataList tend to be more efficient because you can use SQL for your filtering. In the future, we may want to switch to a different list type. 

```suggestion
    public function updateUsage(SS_List &$usage, DataObject &$record) {
```

code/Extensions/UsedOnTableExtension.php
```diff
@@ -0,0 +1,25 @@
...
+        $usage = $usage->filterByCallback(function($userId, $user) {
+            return !in_array($userId, [
```
maxime-rainville: `$userId` will be a `DataObject` instance (why is it called `$userId`?). `FileLink::class` will return a string. So we're checking that a DataObject instance is not  in an array of strings.

I'm somewhat confused as to why PHP thinks that a class instance is equal to the string representation of its class name. I don't think we should rely on this behaviour.

You can achieve the same results with something like this: 
`$usage = $usage->exclude('ClassName', FileLink::class);`

code/Extensions/UsedOnTableExtension.php
```diff
@@ -0,0 +1,25 @@
...
+namespace SilverStripe\AssetAdmin\Extensions;
+
+
+use SilverStripe\Assets\Shortcodes\FileLink;
+use SilverStripe\Core\Extension;
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\DataObject;
+
+/**
+ * Class UsedOnTableExtension
```
maxime-rainville:
```suggestion
```

code/Extensions/UsedOnTableExtension.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Hides File Links on the Used On tab when viewing files
```
maxime-rainville:
```suggestion
 * Hides File Links on the "Used On" tab when viewing files.
```

bergice:
> Looks good ... just need to fix some linting issues.

Should be good to go once travis is done.

PR #999, opened 2019-09-11
Only retrieve file usage count when deleting files

maxime-rainville:
This updates the delete file logic to fetch the file usage data just-in-time, rather than getting it on every readFileQuery request.

To do this we introduce a new slice of state dedicated to displaying a confirm modal and displaying different messages based on the usage count of the files to delete.

This also merge the bulk delete and single file delete logic.

# Parent issue
* #931


client/src/boot/registerQueries.js
```diff
@@ -1,11 +1,12 @@
 import Injector from 'lib/Injector';
-import { fileInterface, file, folder } from 'lib/fileFragments';
+import { fileInterface, file } from 'lib/fileFragments';
 import readFilesQuery from 'state/files/readFilesQuery';
+import readFileUsageQuery from 'state/files/readFileUsageQuery';
 
 const registerQueries = () => {
   Injector.query.registerFragment('FileInterfaceFields', fileInterface);
   Injector.query.registerFragment('FileFields', file);
-  Injector.query.registerFragment('FolderFields', folder);
```
maxime-rainville: This may be a breaking change. We may need to keep this around.

client/src/constants/index.js
```diff
@@ -22,63 +22,6 @@ export default {
       canApply: (items) => (
         items.every(item => item && item.canDelete)
       ),
-      confirm: (items) => new Promise((resolve, reject) => {
```
maxime-rainville: This is where the bulk of the confirmation logic for file deletion was. It's been kind of replicated with a modal: you should be getting the same messages bu in a modal.

If you get a different message under the same condition, then that should be treated as a bug.

client/src/boot/registerQueries.js
maxime-rainville: This may be a breaking change. We may need to keep this around.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -443,6 +445,8 @@ class AssetAdmin extends Component {
    * @param {array} ids
    */
   handleDelete(ids) {
+    this.props.actions.confirmDeletion.deleting();
```
maxime-rainville: This bit of code tells the redux store that we're in the process of deleting files. This allows the delete modal to still be rendered while we're setting `isOpen` flag to `false`, allowing the modal to be hidden with a pretty animation.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -474,7 +478,35 @@ class AssetAdmin extends Component {
         this.props.actions.files.readFiles();
 
         return deleteFiles;
-      });
+      })
+      .then((resultItems) => {
```
maxime-rainville: This bit was moved over from Gallery. It's responsible for displaying an alert to confirm files have been deleted.

Once we get toast notification working in React, we should swap them in here.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -694,7 +726,6 @@ class AssetAdmin extends Component {
         graphQLErrors={this.props.graphQLErrors}
         createFileApiUrl={createFileApiUrl}
         createFileApiMethod={createFileApiMethod}
-        onDelete={this.handleDelete}
```
maxime-rainville: This could be considered a breaking change.

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
```diff
@@ -0,0 +1,113 @@
...
+  ),
+  injectGraphql('readFileUsageQuery'),
+  withApollo
+)(BulkDeleteConfirmation);
+
+/**
+ * Decide whether to render the Deletion Modal based on the information in the redux store.
+ * This avoid firing off a GraphQL query for nothing.
+ */
+const ConditionalModal = ({ showConfirmation, files, ...props }) => (
```
maxime-rainville: I ended up having two components chain together to decide if we should render the modal:
* ConditionalModal decides if our redux store is in a state that will allow us to run our GraphQL query
* ConnectedModal runs the GraphQL query and shows the outcome in BulkDeleteConfirmation.

Part of me thinks there has to be a better way of doing this, but I'm not sure.

client/src/containers/Editor/Editor.js
```diff
@@ -49,28 +50,7 @@ class Editor extends Component {
     }
 
     if (name === 'action_delete') {
-      // Customise message based on usage
```
maxime-rainville: This logic is being run when trying to delete a file individually via the editor window. This change will allow us to rely on the same logic used for bulk deleting, therefore making the process of deleting file more consistent.

client/src/containers/Gallery/Gallery.js
```diff
@@ -221,42 +221,6 @@ class Gallery extends Component {
...
   }
 
-  /**
-   * Delete a list of items
-   *
-   * @param {Event} event
-   * @param {Array} items
-   * @returns {Promise}
-   */
-  handleBulkDelete(event, items) {
```
maxime-rainville: Moved to AssetAdmin.

Again, removing this method could be considered and API change.

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
```diff
@@ -0,0 +1,113 @@
...
+import DeletionModal from './DeletionModal';
+import BulkDeleteMessage from './BulkDeleteMessage';
+import { getFileInUseCount, getFolderInUse } from './helpers';
+import fileShape from 'lib/fileShape';
+import i18n from 'i18n';
+
+/**
+ * Wires the Redux store and Apollo result set with the DeletionModal.
+ */
+const BulkDeleteConfirmation = ({
```
maxime-rainville: I'm thinking maybe the logic that does the actual deletion of files should be in this file, but I'm not 100% sure.

client/src/lib/fileFragments.js
```diff
@@ -25,15 +25,8 @@ const file = `
     size
     smallThumbnail
     thumbnail
-    inUseCount
     version
   }
 `;
 
-const folder = `
```
maxime-rainville: Removing this could be considered an API change.

client/src/state/confirmDeletion/ConfirmDeletionActions.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function confirm(files) {
+  return {
+    type: ACTION_TYPES.CONFIRM_DELETION_ASK,
+    payload: { files },
+  };
+}
+
+/**
+ * Ask for permission to delete
```
maxime-rainville:
```suggestion
 * The GraphQL request for deleting files has started.
```

client/src/state/confirmDeletion/ConfirmDeletionActions.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param {Object[]} List of files we want to delete
```
maxime-rainville:
```suggestion
```

client/src/state/confirmDeletion/ConfirmDeletionActions.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function deleting() {
+  return {
+    type: ACTION_TYPES.CONFIRM_DELETION_CONFIRM,
+    payload: { },
+  };
+}
+
+/**
+ * Ask for permission to delete
```
maxime-rainville:
```suggestion
 * User has changed their mind and doesn't want to delete files any more.
```

client/src/state/confirmDeletion/ConfirmDeletionActions.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param {Object[]} List of files we want to delete
```
maxime-rainville:
```suggestion
```

client/src/state/confirmDeletion/ConfirmDeletionActions.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function cancel() {
+  return {
+    type: ACTION_TYPES.CONFIRM_DELETION_CANCEL,
+    payload: { },
+  };
+}
+
+/**
+ * Ask for permission to delete
```
maxime-rainville:
```suggestion
 * Deletion workflow was completed and we're returning to our regular state.
```

client/src/state/confirmDeletion/ConfirmDeletionActions.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param {Object[]} List of files we want to delete
```
maxime-rainville:
```suggestion
```

client/src/state/confirmDeletion/ConfirmDeletionActions.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function reset() {
+  return {
+    type: ACTION_TYPES.CONFIRM_DELETION_RESET,
+    payload: { },
+  };
+}
+
+/**
+ * Ask for permission to delete
```
maxime-rainville:
```suggestion
 * Modal has closed, we can stop rendering it.
```

client/src/state/confirmDeletion/ConfirmDeletionActions.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param {Object[]} List of files we want to delete
```
maxime-rainville:
```suggestion
```

tests/behat/features/manage-files.feature
```diff
@@ -71,14 +71,19 @@ Feature: Manage files
     Then I should see the "Form_fileEditForm" form
       And I should see the file status flag
 
+  @delete
```
maxime-rainville:
```suggestion
```

tests/behat/features/manage-files.feature
```diff
@@ -71,14 +71,19 @@ Feature: Manage files
...
     Then I should see the "Form_fileEditForm" form
     When I press the "Other actions" button
-      And I press the "Delete" button, confirming the dialog
+      And I press the "Delete" button
+      And I wait for 1 second
+    Then I should see a modal titled "Confirm file deletion"
+      And I press the Delete button inside the modal
     Then I should not see the file named "file1" in the gallery
 
+  @delete
```
maxime-rainville:
```suggestion
```

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
```diff
@@ -0,0 +1,113 @@
...
+          handler: onCancel
+        },
+      ];
+    }
+  }
+
+  // If we're in the process of canceling/deleting results, we set isOpen to false.
+  // This is just a way to get the nice modal out transtion.
+  // When tell the modal to call the `reset` action when it's done closing.
+  const isOpen = !['canceling', 'deleting'].includes(transition);
```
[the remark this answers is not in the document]
maxime-rainville: I don't think our style guide is meant to be applied to our code ... just the doc. Most of the libraries and PHP/JS/CSS syntax we use will be using Yankee English. My assumption was that we should be using American spelling in our own code, otherwise will inevitably end up with a mix-match of both.

tests/behat/features/manage-files.feature
```diff
@@ -76,7 +76,10 @@ Feature: Manage files
       And I click on the file named "file1" in the gallery
     Then I should see the "Form_fileEditForm" form
     When I press the "Other actions" button
-      And I press the "Delete" button, confirming the dialog
+      And I press the "Delete" button
+      And I wait for 1 second
```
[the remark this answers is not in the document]
maxime-rainville: Some of those requests will be very quick. It's quite possible the spinner won't have time to render or be seen by behat.

I'm not even sure we need the delay at all. I added just to be on the safe side and make sure never got into a race condition.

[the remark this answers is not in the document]
NightJar: Sorry, to be clearer on this - I did not mean the UI component that "shows a thing that spins while something the user waits for an asynchronous event to complete" - I meant the Behat feature `spin functions`
http://docs.behat.org/en/v2.5/cookbook/using_spin_functions.html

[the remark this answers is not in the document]
maxime-rainville: Cool ... I didn't know about that.

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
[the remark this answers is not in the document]
NightJar: Sorry, I thought I saw this used in a translation string somewhere... but it appears I may have been befuddled with `transition`, rather than `translation` :)
100% correct, you could write variable names in French for all that matters (not that I recommend this, for maintainability).

There certainly is a mixmatch of languages in our codebase already - British English is generally preferred I thought though, excepting where it can't be used (e.g. CSS & `color`).
c.f. `silverstripe/framework` `SilverStripe\View\ViewableData::customise` (as opposed to customize)

On a side note, some words are far more insidious via culture pollination though, such as Learnt ✅ vs Learned ⛔️ so perhaps we shouldn't care because language evolves and nothing really matters.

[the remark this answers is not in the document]
NightJar: But on the other hand, perhaps these two items in the array that are strings but are used as identifiers rather than text could be converted to `constants` as opposed to looking like free-form strings?
That way wherever they're used could import the same constants, and it could be a _little_ more obvious that they're an option of a defined set, rather than free-from as such.

tests/behat/features/manage-files.feature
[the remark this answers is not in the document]
dnsl48: We don't need to manually wait because of this: https://github.com/silverstripe/silverstripe-behat-extension/blob/4/src/Context/SilverStripeContext.php#L289

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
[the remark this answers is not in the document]
maxime-rainville: Yeah ... there's a good case for making them constants. The only reason I didn't is because they are only use in this file and in reducer function, I don't expect them being used any where else.

I'll make them constant.

tests/behat/features/manage-files.feature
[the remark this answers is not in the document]
maxime-rainville: I've removed the delay.

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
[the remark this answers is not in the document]
maxime-rainville: Done.

maxime-rainville:
I've addressed everyone's feedback. Are your guys going to have another in-depth pass on this? You didn't have a lot of bad things to say.

tests/behat/features/manage-files.feature
```diff
@@ -167,18 +171,22 @@ Feature: Manage files
   Scenario: I cannot delete a folder containing a file that is in use
     When I check the file named "folder3" in the gallery
     Then I press the "Delete" button
-      And I see the text "currently in use" in the alert
-      And I see the text "before you can delete the folder" in the alert
-      And I confirm the dialog
+      And I wait for 1 second
```
dnsl48: this delay too

tests/behat/features/manage-files.feature
```diff
@@ -167,18 +171,22 @@ Feature: Manage files
...
     Then I should see the file named "folder3" in the gallery
 
   @modal
   Scenario: I can delete a file that is in use with a warning
     When I click on the file named "folder3" in the gallery
       And I check the file named "file1" in the gallery
     Then I press the "Delete" button
-      And I see the text "file is currently in use" in the alert
-      And I confirm the dialog
+      And I wait for 1 second
```
dnsl48: I guess this one as well :)

dnsl48:
> You didn't have a lot of bad things to say.

because there's not too much bad in this PR :smile:

maxime-rainville:
Removed those last 2 waits.

PR #1009, opened 2019-09-24
FEATURE Enable 'upload_folder_id' config to set dialog folderid

### LINKED ISSUE #933: No way to set default folder for TinyMCE image insert dialog
There is currently no way to set the default folder for the `TinyMCE_ssmedia` plugin. All the necessary features seem to be there (`InsertMediaModal` has a `folderId` prop which it uses as the default folder), it just is not exposed to the user.

What I mean by this is that there's currently no way to use the Injector to customise the `InsertMediaModal` component that is used by `TinyMCE_ssmedia` in order to use this prop. Despite the fact that `InsertMediaModal` is registered with Injector, `TinyMCE_ssmedia` imports the component directly. It calls [`loadComponent(InsertMediaModal)`](https://github.com/silverstripe/silverstripe-asset-admin/blob/master/client/src/entwine/TinyMCE_ssmedia.js#L15) to wrap the *imported* component, instead of calling `loadComponent('InsertMediaModal')` which would obtain the component from Injector's component factory. Thus, updating `'InsertMediaModal'` with Injector has no impact on the modal rendered when you click on the insert file button on a TinyMCE editor.

This seems like an easy fix: change the `loadComponent` call to passing a string, but I'm unsure about contribution guidelines/process, or if there was other reasoning for loading `InsertMediaModal` in this manner.

tractorcow:
Fixes #933

Actually was kind of quick to implement, so I just did it. ;P


tractorcow:
I've added an option to tinymceconfig over at https://github.com/silverstripe/silverstripe-framework/pull/9262

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -238,6 +240,25 @@ jQuery.entwine('ss', ($) => {
       return Promise.resolve();
     },
 
+    /**
+     * Get default upload folder
+     *
+     * @returns {number}
```
unclecheese: Return type should include `null`

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -238,6 +240,25 @@ jQuery.entwine('ss', ($) => {
...
+     */
+    getFolderId() {
+      const $field = this.getElement();
+      if (!$field) {
+        return null;
+      }
+
+      const config = jQuery($field).data('config');
```
unclecheese: `$field` should be a `jQuery` object (as you've indicated with the `$`), so this second call to `jQuery` is redundant.

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -238,6 +240,25 @@ jQuery.entwine('ss', ($) => {
...
+      if (typeof config.upload_folder_id === 'number') {
```
unclecheese: Do we want to coerce this, or are you sure the backend won't produce something like `"240"`?

Could do
```js
const id = Number(config._upload_folder_id);
return isNaN(id) ? null : id;
```

tractorcow:
@unclecheese updated and rebased based on your feedback.

client/src/entwine/TinyMCE_ssmedia.js
tractorcow: Done. :D

client/src/entwine/TinyMCE_ssmedia.js
tractorcow: Done.

client/src/entwine/TinyMCE_ssmedia.js
tractorcow: Done.

PR #1011, opened 2019-09-25
Add `Download file` action to file meatballs button

bergice:


client/src/containers/Editor/Editor.js
```diff
@@ -131,6 +137,19 @@ class Editor extends Component {
     }
   }
 
+  downloadFile() {
+    function downloadURI(uri, name) {
+      let link = document.createElement("a");
+      link.download = name;
+      link.href = uri;
+      document.body.appendChild(link);
```
kinglozzer: Do we need a `link.target = '_blank';`? We could also make use of the HTML5 `download` attribute to trigger a download instead of just viewing the file

code/Forms/FileFormFactory.php
```diff
@@ -400,6 +400,25 @@ protected function getReplaceFileAction($record)
...
+    /**
+     * Get Download file action
+     *
+     * @param File $record
+     * @return FormAction
+     */
+    protected function getDownloadFileAction($record)
+    {
+        // Check if record exists and user has correct permissions
+        if (!$record || !$record->isInDB() || !$record->canEdit()) {
```
kinglozzer: Should `$record->canEdit()` be `$record->canView()` here?

client/src/containers/Editor/Editor.js
```diff
@@ -48,6 +48,12 @@ class Editor extends Component {
       return;
     }
 
+    if (name === 'action_downloadfile') {
```
maxime-rainville: Might be worth re-factoring this method to use a switch statement and use actual constants for the action names.

But that's more of a nice to have.

maxime-rainville: Might be worth re-factoring this method to use a switch statement and use actual constants for the action names.

But that's a nice to have.

client/src/containers/Editor/Editor.js
```diff
@@ -131,6 +137,19 @@ class Editor extends Component {
     }
   }
 
+  downloadFile() {
+    function downloadURI(uri, name) {
+      let link = document.createElement("a");
```
maxime-rainville:
```suggestion
      const link = document.createElement("a");
```

client/src/containers/Editor/Editor.js
```diff
@@ -131,6 +137,19 @@ class Editor extends Component {
     }
   }
 
+  downloadFile() {
```
maxime-rainville: Having an inline function here seems needlessly convoluted. Could have been written like this and achieve exactly the same result.

```js
  downloadFile() {
    const {url, name} = this.props.file;
    const link = document.createElement("a");
    link.download = name;
    link.href = url;
    document.body.appendChild(link);
    link.click();
    document.body.removeChild(link);
  }
```

code/Forms/FileFormFactory.php
```diff
@@ -400,6 +400,25 @@ protected function getReplaceFileAction($record)
...
+            return null;
+        }
+
+        $action = FormAction::create('downloadfile', _t(__CLASS__ . '.DOWNLOAD_FILE', 'Download file'))
+            ->setIcon('install');
```
maxime-rainville:
```suggestion
            ->setIcon('p-download');
```

I think p-download would be more appropriate here. But maybe check with @silverstripeux to see what they want. They may want to create a brand new icon

code/Forms/FileFormFactory.php
```diff
@@ -412,6 +431,7 @@ protected function getPopoverActions($record)
             // add the unpublish and replace file actions to the start of the array
             array_unshift($actions, $this->getUnpublishAction($record));
             array_unshift($actions, $this->getReplaceFileAction($record));
+            array_unshift($actions, $this->getDownloadFileAction($record));
```
maxime-rainville: Not sure this is the best order. 
image

Probably another thing to consider running past @silverstripeux.

client/src/containers/Editor/Editor.js
```diff
@@ -131,6 +137,19 @@ class Editor extends Component {
...
+    function downloadURI(uri, name) {
+      let link = document.createElement("a");
+      link.download = name;
+      link.href = uri;
+      document.body.appendChild(link);
+      link.click();
+      document.body.removeChild(link);
```
maxime-rainville: The `click()` on the link will give it the focus. After you remove the element it goes in a weird place.

Please reset the focus to a sensible element after you're done.

code/Forms/FileFormFactory.php
sachajudd: 'p-download' is used primarily as a site tree icon. Looking below at the screenshot you provided the icon looks a little out of place. I'll look into our fonts and see if there's an appropriate alternative and add to admin. For the meantime it's fine and shouldn't be a blocker.

sachajudd: 'down-circled' could potentially work better. image

code/Forms/FileFormFactory.php
sachajudd: I think potentially 'Replace file' and 'Download file' could be swapped around.

client/src/containers/Editor/Editor.js
bergice: It's mostly there to make the code more readable and easier to refactor down the road.

bergice:
> Have you given some thoughts about trying to make the Download action an actual link? This would seem like the ideal approach. Would probably require more work, but it could be reuse in other places.

It can be done with a `LiteralField`. See: `vendor/silverstripe/asset-admin/code/Forms/FileFormFactory.php:103` - but it would require some CSS changes as well.

maxime-rainville:
Using a `LiteralField` might be a quick way to hack it in, but the correct way to do it probably involves updating something the FormBuilder in admin and updating it so that if a specific attribute on the FormAction gets specified, it renders it as a link instead of an actual button. If you feel like trying it out, we could have a look ... but it's very much a nice to have.

client/src/containers/Editor/Editor.js
maxime-rainville: Focus still acts a bit weird ... the problem here is that `previousFocus` will always be the Download button you just clicked. But that will get hidden when the `...` menu is automatically collapse.

I think it would make more sense to give the focus to the `...` button.

maxime-rainville:
I tested in Chrome, Firefox and IE11. It all works fine. Note that in IE11, it wants to display the images rather than download them. You might be able to force IE to download the files by adding a `download` attribute to the link, but I'm not sure.

bergice:
> I tested in Chrome, Firefox and IE11. It all works fine. Note that in IE11, it wants to display the images rather than download them. You might be able to force IE to download the files by adding a `download` attribute to the link, but I'm not sure.

I think this relies on browser settings and response headers in order to work.

kinglozzer:
> Note that in IE11, it wants to display the images rather than download them. You might be able to force IE to download the files by adding a download attribute to the link, but I'm not sure.

Apparently IE doesn’t support the `download` attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Browser_compatibility. We’d have to set up a new endpoint that adds the `Content-Disposition` header, so I think we should just leave it as a legacy browser quirk

PR #1013, opened 2019-09-26
BUG Honour AssetAdminFile::insert_width and AssetAdminFile::insert_height when inserting a new image

maxime-rainville:
This updates ImageFormFactory to respect the default insert_width and insert_height dimension.

# Parent issue
* #917

Issue #917, opened 2019-02-26
Max $insert_width is being ignored in 4.3


purplespider:
Inserting a large image into the WYSIWYG editor usually defaults to inserting it at a max width of 600px or max height of 400px. But in 4.3 it seems to ignore these settings and inserts it at its original size instead, which is very annoying when dealing with large images, e.g. photos uploaded from a camera/phone, as you have to remember to manually set the width/height before inserting.

To reproduce:

1. In a SilverStripe 4.3 WYSIWYG, click **Insert Image**.
2. **Upload** a large image (or select on an existing one)

Before clicking "**Insert File**" note that the Width and Height are set to the image's full width and height, rather than the 600 or 400 px max:

![](http://spdr.me/Mv87rW+)

# Pulls request
* [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/1015

purplespider:
Looks like this issue was caused by the renaming of the width and height fields in this commit: https://github.com/silverstripe/silverstripe-asset-admin/commit/333acfc5cbefbe15d5c91a4e920aaabc6f040009

purplespider:
Any chance in this 4.3 regression getting fixed for 4.4? @maxime-rainville 

It affects any CMS user who adds a large image into the WYSIWYG field.

maxime-rainville:
Sorry, we're already under crunch time. We won't have time to fix it in the 4.4 release.

clarkepaul:
My preference would be to show a warning if an image was uploaded at a large size, and then another warning if a large image was used on a page (if a person was uploading directly from the page it only needs to be a single warning). How does that sound @purplespider ? I'll create a separate issue for this and link once created.

The 600px default width is actually annoying and a bad default which trips up content authors as to why an image is so small or pixelated.

purplespider:
@clarkepaul I'm not sure if we're on the same page here. The most important part is that when a CMS user inserts an image into the WYSIWYG editor, no matter how large the original image is, it should not initially appear larger than the width of the WYSIWYG editor, as this makes it frustratingly tricky for the user to resize the image to their desired size, as the drag handles will be hidden, way off the side of the screen, and they will need to be dragged multiple times to get the image down to a sensible size. This is what is happening due to a 4.3 regression. See this video: http://spdr.me/z13sKu

[...]

PR #1015, opened 2019-09-27
BUG Honour AssetAdminFile insert dimentions when inserting a new image

maxime-rainville:
This updates ImageFormFactory to respect the default insert_width and insert_height dimension.

# Parent issue
* #917

# Supersedes
* #1013

### DIFF
--- a/code/Forms/ImageFormFactory.php
+++ b/code/Forms/ImageFormFactory.php
@@ -141,9 +141,9 @@ class ImageFormFactory extends FileFormFactory
     }
 
     /**
-     * Retrieve the approprite insert dimension for the image, if available.
+     * Retrieve the appropriate insert dimension for the image, if available.
      * @param Image $context
-     * @return array|false
+     * @return array|null
      */
     private function getInsertDimensions($context)
     {
@@ -156,6 +156,6 @@ class ImageFormFactory extends FileFormFactory
                 'Height' => $context->getInsertHeight(),
             ];
         }
-        return false;
+        return null;
     }
 }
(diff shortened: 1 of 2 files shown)


code/Forms/ImageFormFactory.php
```diff
@@ -55,13 +56,13 @@ protected function getFormFieldAttributesTab($record, $context = [])
         $tab->push(
             FieldGroup::create(
                 _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.ImageSpecs', 'Dimensions'),
-                TextField::create(
+                NumericField::create(
```
maxime-rainville: Technically, this doesn't fix anybug, but the Width and Height field only accepts numbers, so we probably should use number fields.

code/Forms/ImageFormFactory.php
```diff
@@ -131,6 +132,30 @@ public function getForm(RequestHandler $controller = null, $name = FormFactory::
...
+        $form = parent::getForm($controller, $name, $context);
+        // Set Width and Height to Insert dimensions if available.
+        if ($context['Record'] && $dimensions = $this->getInsertDimensions($context['Record'])) {
+            $form->loadDataFrom($dimensions);
+        }
+        return $form;
+    }
+
+    /**
+     * Retrieve the approprite insert dimension for the image, if available.
```
Cheddam:
```suggestion
     * Retrieve the appropriate insert dimension for the image, if available.
```

code/Forms/ImageFormFactory.php
```diff
@@ -131,6 +132,30 @@ public function getForm(RequestHandler $controller = null, $name = FormFactory::
...
+        if ($context &&
+            $context->hasMethod('getInsertWidth') &&
+            $context->hasMethod('getInsertHeight')
+        ) {
+            return [
+                'Width' => $context->getInsertWidth(),
+                'Height' => $context->getInsertHeight(),
+            ];
+        }
+        return false;
```
Cheddam: Might be better to return `null` here, as once we're comfortable with our minimum requirement of PHP 7.1 we can make the return type `?array`.

maxime-rainville:
Cool .. I'll merge on green.

Issue #961, opened 2019-06-24
600px default image size is not always the best default, add option for original size


clarkepaul:
When placing an image into the editor the file width is presented at a default of 600px (currently broken in the latest version as per #917). This is fine if the site wants a default of 600px to be used for the majority of the images but this number doesn't consider what the usable space of the page to be, or what the original size of the image was (could be 650px). This results in pixelated images which don't take up the full width of the content area. This becomes more important with responsive views were the size set in desktop might not be as usable at other screen resolutions. 

Issues:
- Badly resized images resulting in pixelation.
- Can't see the original size of the image once the file is placed in the editor (from the editor view) and wouldn't know if the image is being resized larger than the original image.
- It isn't well known that the default size of 600px can be removed (make fields empty) so that the image acts responsively using its original size (typically also using max-width 100% of the usable area within template CSS).

# Acceptance criteria
- [x] Files have the ability to be configured between a "Best fit" and "Original size". The "Best Fit" size should inherit the 600px default from $insert_width (or look to depreciate that functionality?)
- [x] The "Best Fit" option should be selected by default when adding a new file
- [x] The image popover in TinyMCE is updated to include a "Best fit" and an "Original size" option
- [x] The "Best fit" and an "Original size" options are also available in the 'edit file' view
- [x] The alignment options are removed from the image popover, the Edit icon is kept\

# Designs
You can find the new TinyMCE designs under [TinyMCE Edit image](https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/folder/components/5c68f96ffb26bb0018525d0a) and the dimensions Best fit / Original designs under [File details](https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/folder/components/5d70aad32041c840c590d9cd).

# Notes
- We'll need to have a closer look at what we can get out of the TinyMCE popover. Depending on the outcome, we may need to look at other options.
- Let's also investigate allowing the developer to configure the defaults.
- STRETCH GOAL: allow the size options to be configured to set defaults, e.g small, medium, large

# Pull request
* [ ] https://github.com/silverstripe/silverstripe-framework/pull/9276
* [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/1018
* [ ] silverstripe/silverstripe-admin#965

newleeland:
We potentially could use the popover thats in TinyMCE right now?

[...]

PR #1018, opened 2019-10-03
NEW Add HTML editor button to switch to image size preset

maxime-rainville:
This PR updates the WYSIWYG image menu to remove the alignment buttons and replace them with button to adjust the size of the image on the fly.

# Parent issue
* #961

# Depends on
* https://github.com/silverstripe/silverstripe-framework/pull/9276/
* https://github.com/silverstripe/silverstripe-admin/pull/965


client/src/components/ProportionConstraintField/ImageSizePresetList.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from 'reactstrap';
+
+/**
+ *
```
maxime-rainville:
```suggestion
 * Display a list of possible imge size presets that the user can click to quicky set the width and height of the image in the placement form.
```

sachajudd: image?

client/src/components/ProportionConstraintField/ImageSizePresetList.js
```diff
@@ -0,0 +1,38 @@
...
+      {imageSizePresets.map(
+    ({ width, text }) => (
+      <li key={text} className="image-size-preset-list__list-item">
+        <Button
+          color="link"
+          size="sm"
+          onClick={() => onSelect(width || originalWidth)}
+          disabled={(originalWidth < width) || currentWidth === (width || originalWidth)}
+        >
+          {text}
```
maxime-rainville: Need to insert something like "Set the image size to " here

Cheddam:
```suggestion
          <span className="sr-only">Set image size to</span> {text}
```

maxime-rainville: I'll add with a translation.

maxime-rainville: The `text` bit is translated server side, but `Set image size` to needs to be translated client side.

client/src/containers/InsertMediaModal/InsertMediaModal.js
```diff
@@ -153,6 +166,10 @@ InsertMediaModal.propTypes = {
   onClosed: PropTypes.func,
   className: PropTypes.string,
   actions: PropTypes.object,
+  imageSizePresets: PropTypes.arrayOf(PropTypes.shape({
+    text: PropTypes.string.isRequired,
```
dnsl48: I guess text could be optional and width required? (we could derive preset name from width, but not vise-versa). Lack of width only makes sense in a single use case - original size, which I don't think we should expose through configuration at all.

client/src/entwine/TinyMCE_ssmedia_sizepressets.js
```diff
@@ -0,0 +1,147 @@
...
+    });
+  });
+
+  /**
+   * Retrieve the current selection if it's an image.
+   * @returns {HTMLElement|undefined}
+   */
+  const image = () => {
+    const img = editor.selection.getNode();
+    return img && img.tagName === 'IMG' ? img : undefined;
```
dnsl48: is `tagName` guaranteed to always be in uppercase?

code/Forms/ImageFormFactory.php
```diff
@@ -130,17 +130,17 @@ public function getForm(RequestHandler $controller = null, $name = FormFactory::
...
                 ]);
             }
         });
 
         $form = parent::getForm($controller, $name, $context);
         // Set Width and Height to Insert dimensions if available.
-        if ($context['Record'] && $dimensions = $this->getInsertDimensions($context['Record'])) {
-            $form->loadDataFrom($dimensions);
-        }
+        $form->loadDataFrom($this->getInsertDimensions());
```
dnsl48: do we still need to load the empty data here?

dnsl48:
I think the main suggestion I have is - we should leave "Original" to be a non-configurable option which always exists and ensured on the front-end side (no need to configure it as a preset on backend at all)

client/src/components/ProportionConstraintField/ImageSizePresetList.js
maxime-rainville: Fixed.

client/src/entwine/TinyMCE_ssmedia_sizepressets.js
maxime-rainville: Yes https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName.

If we were serving XHTML we might get the lowercase version, but we're not.

code/Forms/ImageFormFactory.php
maxime-rainville: Yes, otherwise when you select an image in the front end, the form schema will be pre-filled with the original size rather than our best fit size.

I'll remove the private method and set the blank value directly here instead.

code/Forms/ImageFormFactory.php
```diff
@@ -149,17 +149,11 @@ public function getForm(RequestHandler $controller = null, $name = FormFactory::
...
-            return [
-                'Width' => $context->getInsertWidth(),
-                'Height' => $context->getInsertHeight(),
-            ];
-        }
-        return null;
+        return [
+            'Width' => '',
+            'Height' => '',
+        ];
```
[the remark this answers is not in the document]
maxime-rainville: https://github.com/silverstripe/silverstripe-asset-admin/pull/1018#discussion_r336730614

code/Forms/ImageFormFactory.php
```diff
@@ -58,7 +58,7 @@ protected function getFormFieldAttributesTab($record, $context = [])
         $tab->insertAfter(
             'Alignment',
             FieldGroup::create(
-                _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.ImageSpecs', 'Dimensions'),
+                //                _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.ImageSpecs', 'Dimensions'),
```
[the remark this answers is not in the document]
maxime-rainville: Done.

client/src/components/ProportionConstraintField/ImageSizePresetList.js
```diff
@@ -0,0 +1,52 @@
...
+      {imageSizePresets.map(
+    ({ width, text }) => (
+      <li key={text} className="image-size-preset-list__list-item">
+        <Button
+          color="link"
+          size="sm"
+          onClick={() => onSelect(width || originalWidth)}
+          disabled={(originalWidth < width) || currentWidth === (width || originalWidth)}
+        >
+          <span className="sr-only">{srText(text)}</span>
```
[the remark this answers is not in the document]
maxime-rainville: Good idea. I've implemented it on the PHP side ... otherwise I would have had to fix it for the WYSIWYG popover as well.

client/src/components/ProportionConstraintField/ImageSizePresetList.js
```diff
@@ -0,0 +1,52 @@
...
+) : null);
+
+ImageSizePresetList.propTypes = {
+  onSelect: PropTypes.func,
+  imageSizePresets: PropTypes.arrayOf(PropTypes.shape({
+    text: PropTypes.string,
+    width: PropTypes.number
+  })),
+  currentWidth: PropTypes.number,
+  originalWidth: PropTypes.number,
```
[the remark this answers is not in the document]
maxime-rainville: I've added it to `originalWidth`. There's a scenario where `currentWidth` isn't set to begin width, so I would rather leave it optional.

maxime-rainville:
I've answered most of the feedback. I'm a bit ambivalent about removing about hardocding the "original size" preset in the front end lib. This would preclude people from reordering it or changing the description for it or defining it as the default options. I also think it adds an extra bit of complexity because you're introducing two ways of defining the same type of buttons.

client/src/entwine/TinyMCE_ssmedia_sizepressets.js
dnsl48: do you reckon it's not worth adding `toUpperCase` here to support both?

dnsl48:
> I'm a bit ambivalent about removing about hardcoding the "original size" preset in the front end lib.

I'm not feeling too strong about it to block the PR, so I'll just put my arguments about it here, then feel free to make your decision if you want to leave as is or patch accordingly.

----

- ability to omit `width` in the configuration
    - makes the config more complicated, adds an implicit configuration option through changing config structure rather than values
    - makes frontend code more complicated (I've left a couple notes about where the code could be simpler if `width` wouldn't be optional)
- if we handled "original" size separately in the code, the implementation could become simpler and more straightforward since we wouldn't have to check everywhere if `width` is defined or not
- from the front-end perspective, "Original" seems to be an "already known" option that doesn't really require neither `width` nor translation nor anything else from the backend, so it feels to be "natural" option for having, whereas actual sized presets are always optional by definition and feel like "extra" stuff
 - controlling ordering of "original" and "presets" through backend configuration feels to be an "easy win" feature, but in reality it's unnecessary coupling, especially if eventually we would want to change the UI, make it to be called something else (e.g. "reset") and stick out from the list of other options - we won't be able to do that easily. Controlling UI through backend configuration feels to be somewhat anti-patternish to me

maxime-rainville:
Yeah ... having to check if width is Zero is a bit of annoyance. However, if I treated it has a system option, I would still need to implement it has a separate option, which might not be trivial.

If I was redoing it again, I probably would try to think of a better way of handling this case. Since, it's not an absolute clear-cut case that one way is better than the other, I'm going to merge as-is rather than spend an hour or two tweaking this.

cwchong:
Hi, is there a way to hook into the handler or does clicking on the buttons emit any events which we can catch
There seems to be an issue if caption text are provided during insertion from file admin, as this automatically wraps the image in a ```div``` with a defined width
Toggling the image to 'Original' within the editor after will appear not to do anything since it is now wrapped by a div with smaller defined width

Optimally the handler should also toggle the wrapping div (classed "captionImage") to insert/remove the width css definition

brynwhyman:
Hey @cwchong that sounds like it could be a new bug. I'd recommend [reporting a new bug](https://docs.silverstripe.org/en/4/contributing/issues_and_bugs/#reporting-bugs).

Issue #903, opened 2019-01-18
Multiple entries appear within the "Used on" tab for a single file placed 


newleeland:
## Description
When a file is placed within a block within the WYSWYG, on the **Used on** tab the file appears to be used 4 times (Page, ElementalArea, Content block, File link) for a single placement. 

Example:
![image.png](https://images.zenhubusercontent.com/574b650dbf93871ccd9bc88f/346e3bde-bc2a-469d-bad5-cea9467a46f7)

Ideally, there should only be one entry if the image is used once. We should guide them to the place in which they can edit the placement or remove the file. 
e.g.
```
 #     Used on          Details
[#]   [Page name]       Placed within [Block name]
```

## Step to reproduce
Navigate to a file (which is being used on a page/block) in the Files area
Navigate to the **Used on** tab
The information displayed is confusing/misleading

# Pull Requests
- [x] https://github.com/silverstripe/silverstripe-assets/pull/344
- [x] https://github.com/dnadesign/silverstripe-elemental/pull/817

robbieaverill:
Reproduced on CWP 2.3

bergice:
Rewritten to use the `hide_in_campaign` config instead.

brynwhyman:
We're pausing work on this issue until we investigate implementing a new API that can take care of filtering data objects by marking some as internal and not worth 'displaying', v.s reusing the `$hide_in_campaigns` class for something that isn't really related to campaigns. 

That work is currently bundled with https://github.com/silverstripe/silverstripe-campaign-admin/issues/128 but might be better captured in https://github.com/silverstripe/silverstripe-cms/issues/2454

brynwhyman:
Contrary to my previous comment, we're interested in having another look at the existing PRs (reusing the $hide_in_campaigns class) and possibly moving forward with that approach, as opposed to a re-factor. This is a high impact issue so getting something into sites as early as possible would be great.

Feedback on this is welcome...

emteknetnz:
https://github.com/dnadesign/silverstripe-elemental/pull/817 should just be merged, has already been approved (I just don't want to self-merge on the dnadesign account)

https://github.com/silverstripe/silverstripe-admin/pull/1083 I've removed the newly added `<tr>` click functionality since that's what all the discussions was around

PR #1020, opened 2019-10-08
Do not display records with `hide_in_campaigns` config set to `true` in the `Used On` table of files

bergice:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/903


robbieaverill:
Tests are red, any chance of green?

emteknetnz:
We want to keep this functionality
- FileLinks should not show on the UsedOnTable https://github.com/silverstripeltd/product-issues/issues/222#issuecomment-663253169
- This is the correct way to remove them
- As Max said, cannot release this in a minor, would need to be a major release anyway

PR #1027, opened 2019-10-23
NEW Add tip to Alt Text field in Insert Media dialog

### LINKED ISSUE #983: Better defaults for file alt text
# Overview
We're combining a couple of issues relating to improving the default handling of alt text for files:
* [Alt text defaults to the image's title if not set on CMS](https://github.com/silverstripe/silverstripe-framework/issues/9099) from @Gazzayeatman
* [When inserting images to content, alt text should autofill
](https://github.com/silverstripe/silverstripe-asset-admin/issues/840) (similar to)

We need to ensure that the user is guided to make the right decisions here and is not unknowingly adding alt text that will not be beneficial to screen readers.

# Two changes:

#### 1. If alt text is empty the `alt` attribute should be empty

At the moment an image's alt text defaults to the image's title if there is no alt text set. There are some cases where the user does no want alt text at all. See: https://www.w3.org/WAI/tutorials/images/decorative/

Note, this isn't reflected on the admin UI, but can be seen in the image HTML.

#### 2. Guide user to add alt text for files

There should be an improvement in prompting users that alt text is probably required for a file (while keeping the use case from the above change in mind).

Suggestions:
- There is clear messaging advising users to add alt text for the benefit of screen readers
  - Noting that in CWP 1.x there was a prompt when the alt text is left empty: https://github.com/silverstripe/cwp/issues/149
- Do not auto-populate the alt text field to avoid setting alt text like "image 1"
- Note: this should only be implemented for the WYSIWYG, not for the upload field

#### ~(Stretch goal) 3. Reorder the fields presented for placing a file~
Tracking in: https://github.com/silverstripe/silverstripe-asset-admin/issues/991

### PRs

- [x] https://github.com/silverstripe/silverstripe-admin/pull/964
- [x] https://github.com/silverstripe/silverstripe-framework/pull/9281
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/1027

Cheddam:
Applies tips to the Alt Text and Title Tooltip fields in the Insert Media dialog, or falls back to descriptions if tips aren't available.

Resolves #983.

### DIFF
--- a/code/Forms/ImageFormFactory.php
+++ b/code/Forms/ImageFormFactory.php
@@ -12,6 +12,8 @@ use SilverStripe\Forms\NumericField;
 use SilverStripe\Forms\OptionsetField;
 use SilverStripe\Forms\Tab;
 use SilverStripe\Forms\TextField;
+use SilverStripe\Forms\Tip;
+use SilverStripe\Forms\TippableFieldInterface;
 
 class ImageFormFactory extends FileFormFactory
 {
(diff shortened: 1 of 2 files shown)


Cheddam:
@clarkepaul This update matches the wording in the designs you provided - can you confirm you're happy with it?

clarkepaul:
Yep I think so unless anything stands out as not coming across well. Feel free to provide a peer review of the text @Cheddam

code/Forms/ImageFormFactory.php
```diff
@@ -77,20 +79,42 @@ protected function getFormFieldAttributesTab($record, $context = [])
...
-                ))
+            $altTextField = TextField::create(
+                'AltText',
+                _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.AltText', 'Alternative text (alt)')
+            )
+        );
+
+        $altTextDescription = _t(
+            'SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.AltTextDescription',
+            implode(
```
maxime-rainville: Isn't it a tiny bit convoluted to use implode to add a space between two strings?

maxime-rainville: The content of the `AltTextDescription` message has changed substantially. We probably need to use a different translation key as any existing translation won't match the new content.

brynwhyman:
>I assumed that when you leave the alttext textfield blank, your image should have an empty alt on the front end. The Image Shorcode parser seem to be too clever for its own good and filling that in from the image metadata.

Yeah, @maxime-rainville is correct here. That's what I intended to convey in the issue with: "**1. If alt text is empty the alt attribute should be empty**"

Cheddam:
@maxime-rainville @brynwhyman That behaviour will be corrected when [this PR](https://github.com/silverstripe/silverstripe-assets/pull/338) is merged (I thought we'd gotten this across the line a while back, but the tests failed. Working on getting them green now.)

I've addressed the feedback related to this PR, so if you're happy with the state of it please give it a green tick and merge ✅

dnsl48:
Doesn't seem to be covering the AC 1

peek-asset-admin-1027 webm

Cheddam:
@dnsl48 As mentioned in a comment above, AC 1 is covered by [a separate PR](https://github.com/silverstripe/silverstripe-assets/pull/338) which also needs merging.

Issue #1035, opened 2019-12-10
Remove behat test that connects to youtube


emteknetnz:
tests/behat/features/insert-an-image.feature:40

Will make a connection to youtube.  Youtube has started responding with a 429 status code which means rate limited

Disable this test and switch to manually testing this functionality instead

emteknetnz:
Pull request https://github.com/silverstripe/silverstripe-asset-admin/pull/1036

brynwhyman:
Just a side note - interesting to see that production sites are getting limit rates too. The failing test is kind of helpful after all? https://github.com/silverstripe/silverstripe-framework/issues/9073#issuecomment-564732178

robbieaverill:
Merged

PR #1036, opened 2019-12-12
Remove insert youtube media behat test

emteknetnz:
This removes a behat test that is sporadically failing on Travis because youtube is returning a 429 status code which means rate limiting

Instead of running this as an automated test, we can run this instead as a manual test

Fix for https://github.com/silverstripe/silverstripe-asset-admin/issues/1035

### DIFF
--- a/tests/behat/features/insert-an-image.feature
+++ b/tests/behat/features/insert-an-image.feature
@@ -36,19 +36,6 @@ Feature: Insert an image into a page
       # Required to avoid "unsaved changed" browser dialog
       And I press the "Save" button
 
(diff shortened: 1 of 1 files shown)


robbieaverill:
>Instead of running this as an automated test, we can run this instead as a manual test

I agree with @maxime-rainville - you should add this to your HipTest suite, otherwise it will be forgotten.

maxime-rainville:
Just added a test specifically for YouTube https://studio.cucumber.io/projects/103281/test-plan/folders/693809/scenarios/3992494

Issue #813, opened 2018-07-24
Having to "Edit Original File" is unintuitive compared to SS3 inline editing


feejin:
When viewing image info the details (Title, Filename etc) appear to be form fields but are not editable, they can only be edited by clicking Edit Original File which opens in a new tab. 

It's more noticeable with the FocusPoint module when you want to just make a quick change as soon as you upload the image. Makes for quite a jarring user experience and I have to explain this to non-technical clients.

tl;dr make asset details editable without having to go via Edit Original File.

# User story
As a CMS user, I want to be able to stay in the current modal when editing a file, so that I can make quick updates to the file without loosing context of the area related to the file.

# Acceptance Criteria
- [x] 'Editing' a file keeps the user in the same modal
- [x] There is a button that initiates the file in a new editable state.
- [x] When editing the file, you can save/publish. 
- [x] When editing the file, the other actions are also available (archiving, unpublishing, replacing the file).
- [ ] A prompt will warn the user that edits will be applied globally.
- [x] There is a clear flow of inserting the image after editing. 
- [x] Editing a file in the modal might change its draft/live state. When you're done editing these changes should be reflected in the rest of the modal.

- [ ] This improvement is for the WYSIWYG modal only, and the uploadfield modal should be handled in a separate issue.

# Stretch Goal
* [x] Implement this feature in the Upload Field Modal or raise a separate issue for it.

# Suggested designs
* https://invis.io/WCMZMHHXRJV#/381953634_Files_Tiny_Insert_Files_new_File_start
* https://invis.io/WCMZMHHXRJV#/381953633_Files_Tiny_Insert_Files_existing_File_start

# Pull Requests
* [x] https://github.com/silverstripe/silverstripe-admin/pull/968
* [x] https://github.com/silverstripe/webpack-config/pull/39
* [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1038
* [ ] https://github.com/jonom/silverstripe-focuspoint/pull/78 (tweaks focus point module to work with the new image placement form)
* [x] https://github.com/silverstripe/silverstripe-framework/pull/9425
* [x] https://github.com/silverstripe/cwp-recipe-kitchen-sink/pull/63

maxime-rainville:
That's kind of related to https://github.com/silverstripe/silverstripe-asset-admin/issues/756 and https://github.com/silverstripe/silverstripe-asset-admin/issues/785.

[...]

PR #1038, opened 2019-12-18
Allow user to edit file information directly in the modal

maxime-rainville:
This PR adds the ability to edit the detail of a file within the file insertion modal.

# Other improvement
* Move `getFormSchema` out of AssetAdmin
* Replace convoluted logic to get breadcrumb data into redux store
* Tweak the header spacing

# Parent issue
* #813


client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -25,62 +24,7 @@ import configShape from 'lib/configShape';
 import { injectGraphql } from 'lib/Injector';
 import BulkDeleteConfirmation from '../BulkDeleteConfirmation/BulkDeleteConfirmation';
 import * as confirmDeletionActions from 'state/confirmDeletion/ConfirmDeletionActions';
-
-function getFormSchema({ config, viewAction, folderId, fileId, type }) {
```
maxime-rainville: `getFormSchema` is required in other places now. I've moved it to its own file.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -25,62 +24,7 @@ import configShape from 'lib/configShape';
...
-}
-
-
-/**
- * Check if either of the two objects differ
- *
- * @param {Object} left
- * @param {Object} right
- */
-function compare(left, right) {
```
maxime-rainville: `compare` was only use by some dodgy willReceiveProps. It's no longer used anymore.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -138,63 +72,6 @@ class AssetAdmin extends Component {
     return 0;
   }
 
-  /**
-   * Assign breadcrumbs from selected folder
-   *
-   * @param {Object} props
-   */
-  setBreadcrumbs(props) {
```
maxime-rainville: Most of this logic was moved to a dedicated stateless Asset Breadcrumb component.

client/src/containers/AssetAdmin/Breadcrumb.js
```diff
@@ -0,0 +1,82 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import i18n from 'i18n';
+import { Component as PlainBreadcrumb } from 'components/Breadcrumb/Breadcrumb';
+import { hasFilters } from 'components/Search/Search';
+
+/**
+ * Display a breadcrumb based on the current folder, bypassing the stateful Breadcrumb component.
+ */
+const Breadcrumb = ({ folder, query, getUrl, onBrowse, onFolderIcon }) => {
```
maxime-rainville: The old approach was taking the information we were getting back from our GraphQL request, building some breadcromb data and calling redux actions to store them in the redux store.

This was creating two source of truth and adding useless complexity. Instead, I'm using the plain Breadcrumb component (as opposed to the one that's wired into the redux store) and passing it the breadcrumb data directly.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -351,11 +228,8 @@ class AssetAdmin extends Component {
 
   /**
    * Handler for when the folder icon is clicked (to edit hte folder)
```
maxime-rainville:
```suggestion
   * Handler for when the folder icon is clicked (to edit the folder)
```

maxime-rainville:
Things left to do to get this PR over the line:
* [ ] Add unit test for `getFormSchema`
* [ ] Add unit tests to Editor to cover
  * [ ] handleClose
  * [ ] editorHeader
  * [ ] createFn
* [ ] Add unit tests to EditorHeader
* [ ] Update Modal store to cover new state
* [ ] Write some behat test
* [ ] UX feedback
  * [ ] Make placement an header rather than a tab
  * [ ] Add a warning about editing files globally in the modal

client/src/containers/Editor/Editor.js
```diff
@@ -33,6 +43,27 @@ class Editor extends Component {
     };
   }
 
+  /**
+   * Build the form schema URL to pass to the Form Builder Loader
+   * @returns {string}
+   */
+  getFormSchemaUrl() {
```
maxime-rainville: This used to be done in the rendering method, but I needed it in two places, so I moved it to it's own method. I've switch to using `qs` and `url` rather than directly manipulating string.

client/src/containers/Editor/tests/EditorHeader-story.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
```
maxime-rainville: Not sure we should keep this file. It's helpful to tweak the styling of the Editor Header, but it's kind of clunky.

client/src/entwine/TinyMCE_sslink-file.js
```diff
@@ -63,14 +64,21 @@ jQuery.entwine('ss', ($) => {
    */
   $(`.js-injector-boot #${modalId}`).entwine({
     renderModal(isOpen) {
-      const handleHide = () => this.close();
+      const { dispatch } = Injector.reducer.store;
```
maxime-rainville: This is a tiny bit of mad science. We're getting the redux store in entwine so we can push some data into to control the modal.

client/src/state/modal/ModalActionTypes.js
```diff
@@ -1,3 +1,7 @@
 export default {
-  DEFINE_IMAGE_SIZE_PRESETS: 'DEFINE_IMAGE_SIZE_PRESETS'
+  DEFINE_IMAGE_SIZE_PRESETS: 'DEFINE_IMAGE_SIZE_PRESETS',
```
maxime-rainville: I'm starting to think that it doesn't make a lot of sense to have the image presets along with all the form schema stuff together. Maybe we need to move it to a different reducer.

client/src/containers/Editor/tests/EditorHeader-story.js
emteknetnz: what are the other options for where it could live?

client/src/entwine/TinyMCE_sslink-file.js
emteknetnz: could you put a comment into the code explaining the mad science :)

client/src/state/modal/ModalActionTypes.js
emteknetnz: So how would file naming work, would you split this file into two, as well as ModalActions.js? i.e.
- ModalFormActionType
- ModalImageSizeActionType
- ModalImageSizeActions
- ModalImageSizeActions
?

client/src/containers/AssetAdmin/stateRouter.js
```diff
@@ -111,6 +112,10 @@ class AssetAdminStateRouter extends Component {
       throw new Error(`Invalid action provided: ${action}`);
     }
 
+    if (this.state.fileId !== fileId) {
+      this.props.actions.resetFormStack();
```
maxime-rainville: This reset the Editor form in modal

client/src/entwine/TinyMCE_sslink-file.js
maxime-rainville: Explain this redux store mad science

client/src/state/modal/ModalReducer.js
```diff
@@ -1,20 +1,72 @@
...
+ */
+
+/**
+ * Initial base state
+ * @type { ModalState }
  */
 export const initialState = {
-  imageSizePresets: []
+  imageSizePresets: [],
+  formSchemaStack: [],
```
maxime-rainville: Explain the stack mad logic.

tests/behat/features/insert-an-image.feature
```diff
@@ -36,6 +34,29 @@ Feature: Insert an image into a page
       # Required to avoid "unsaved changed" browser dialog
       And I press the "Save" button
 
+  @foobar
```
maxime-rainville:
```suggestion
```

client/src/containers/Editor/Editor.js
```diff
@@ -197,17 +283,22 @@ class Editor extends Component {
     const campaignTitle = i18n._t('Admin.ADD_TO_CAMPAIGN', 'Add to campaign');
     const Loading = this.props.loadingComponent;
 
-    return (<div className={editorClasses.join(' ')}>
+    // FormBuilderLoader doesn't do anything with the file props, but passing it along forces
+    // React to re-render the EditorHeader component when file gets populated by GraphQL data
+    const file = this.props.file;
```
maxime-rainville: This fixes an issue with the Insert Link form. Where firing a GraphQL request to fetch the file data and a request for the FormSchema of the Insert link form.

Most of the the time, the GraphQL data comes back first and the selected file data gets pass down to the Editor in time for the `EditorHeader` to be rendered correctly. However on some occasion the FormSchema comes back first and the `EditorHeader` gets rendered without the necessary file data.

Passing `file` to `FormBuilderLoader` will cause a re-render when the GraphQL file data is filled in later on, which will eventually cause a re-render of the `EditorHeader`.

client/src/containers/Editor/tests/EditorHeader-story.js
maxime-rainville: Just summarising the chat we had yesterday. The reason why this file is "clunky" is because it wrap the EditorHeader into a lot of fluff HTML that is normally generated by the parent Editor component. It's useful for doing basic testing and styling on the EditorHeader component, but not necessarily something we should keep around long term. 

For example, if we update the Editor component, it won't be reflected here, because the HTML is manually written here.

As a side node, this illustrate a weakness of the Editor component. It mixes state management and presentational concerns which make it difficult to test the state management side of it and to create a story for it in the pattern lib.

The conclusion was to delete this file.

client/src/state/modal/ModalActionTypes.js
maxime-rainville: On second thought, I'm tempted to say keep it like this. Moving it to it's own slice of state is not obviously better and there's an argument to say that all this state data is only relevant when AssetAdmin is displayed in a Modal.

emteknetnz: OK, so that's a definite keep it this way?  Means I can resolve conversation and we can focus on other bits

client/src/containers/Editor/Editor.js
emteknetnz: it's useful to say why a particular piece of code is done the way it is, though most people in the future won't see github comments - would you be able to summarise and add as a code comment?

maxime-rainville: That's what the two line above are meant to explain more succinctly. Do you think this is not clear enough?
```
    // FormBuilderLoader doesn't do anything with the file props, but passing it along forces
    // React to re-render the EditorHeader component when file gets populated by GraphQL data
```

client/src/entwine/TinyMCE_sslink-file.js
maxime-rainville: I've added a comment about the redux bit.

client/src/state/modal/ModalActionTypes.js
maxime-rainville: Yeah good to resolve. (FYI Sometime I'll leave some explanation about a decision I've made for which I feel ambivalent.)

maxime-rainville:
GGRGRGRGR!!!!

This seems to have broken the UploadField!!!

maxime-rainville:
Fixed the UploadField
Added some Behat test to cover it
Rebased with `1` branch

emteknetnz:
image

I found the text 'Details' on this button to be pretty confusing.  I'm thinking we should change it to 'Original image' or something similar.  Might be worth checking with PUX

emteknetnz:
This relates to the jonom/focuspoint module which was listed on the original issue

'Details' screen = editing original image
'Placement' screen = editing instance i.e. I just clicked the image in wysiwyg > edit image button

When I set the coordinates in the 'Details' screen, then when I click 'save/publish' it works as expected and the coordinates are saved

Then when I see the focus point in the 'Placement' screen, it now has the same coordinates as in the one in the 'Details' screen.   When I change the coordinates and then click 'update file', the coordinates are then reverted to the ones in the 'Details' screen

Possibly we just need to stop the focus point module from appearing in the 'Placement' screen

emteknetnz:
The word 'Details' is doubly confusing when clicking 'Insert from Files' > 'Upload file' (though to be fair it was probably already like this

When I upload an image I just uploaded, when I hover over it, it says 'Details'

When I click the image, I go to the 'Placement' screen.  There's now also a new button that says 'Details' in the top right.  Which is extremely confusing.  Previously this was a 'Details' tab

emteknetnz:
This is sort of an existing issue, though the new interface makes it seem broken

When I click 'Insert from Files' > 'Upload file', and I upload a new image

In both the new and old interface, I'm in the 'Placement' panel and there's a little orange 'Draft' in the top next to the filesize

In the old interface, the natural thing to do was click 'Insert image' which would take me back to the page edit screen.  When I then 'Publish' the page it's on.  The newly uploaded image would still be 'draft' though it didn't really matter because I could still see it from ?stage=Live

In the new interface, after I upload the image, when I goto to the 'Placement' screen and click the 'Details' button, and I click the 'Publish' button (which was not available on the 'Details' tab of the old interface) then I'm redirected back to the 'Placement' panel, but the orange 'draft' text is still there.

So it seems like it failed to publish.  However, if I refresh my browser, and navigate back to where I was, I can see that the file was actually published.

maxime-rainville:
# Wording of "Details" button
@silverstripeux Do we want to rethink the wording of this word.

# jonom/focuspoint module
Didn't think of that. I'll need to have a look at how that module inject their field into the forms.

Technically, that's not a supported module and we're not changing any APIs in a non backward compatible way ... I wouldn't consider changing the Fields in a Form to be an API change.

It is a very popular module however, so I'll see what I can do to make sure it still works. Arguably, the problem here is that we don't have a sensible way for third parties to modify these File/Image form.

# Sticky Draft tag
I think the problem is that I'm stashing the fields, so I might be accidentally stashing the publication status of the file. I'll have a look.

clarkepaul:
@maxime-rainville Had a quick chat to @emteknetnz about some of the things discussed above. We agreed that the wording on the hover state of files needs to change from "Details" to "View" (along with icon change to an eye) because when placing images it no longer takes you to the details of the file. This change would be visible in both File Admin and in the modal when selecting a file to place.
Pasted_Image_30_01_20__3_03_PM

As for changing the text in the button from "Details" to something else, I'm pretty sure Jared went through user testing around the office to get that wording with different designs. It used to be "Edit original" and I think people who were not used to the CMS didn't really understand the concept of an "original" file. I'm happy to stick with "Details" as the label because it essentially takes you to the Details tab of the file.

clarkepaul:
@maxime-rainville notes for the uploadfield UI, sorry move to a better place if not wanted here.

Uploadfield UX for discussion:
1. Bug - File details (name) isn’t updating when it is edited, saved and inserted into a page.
Pasted_Image_30_01_20__4_40_PM
3. When editing the details of a file... can we hide all of the checkboxes on other files, you only need to select one file... and can we hide/remove the bulk actions when editing a file (bottom left corner) as they aren't needed, all those actions are available on the edit panel for a single file.
Pasted_Image_30_01_20__4_34_PM

I haven't yet seen the multi-file experience into the uploadfield but I don't see that the bulk actions are needed there either, but the checkboxes are required to select multiple files.

With those changes, I think we have progressed things enough to release and provide enough benefit to users.

emteknetnz:
Re Sticky Draft tag - related issue https://github.com/silverstripe/silverstripe-asset-admin/issues/960

maxime-rainville:
I've fixed the following two points:
* Wording of "Details" button
* Sticky Draft tag

This leaves the following:
* [ ] jonom/focuspoint module support
* [ ] Reflecting changes back to Upload field after the file has been edited
* [ ] Disabling bulk action and checkboxes when editing a file
* [ ] Retain "File selected" counter and Insert Button when editing file in multi-select upload field.

maxime-rainville:
Got the UI fix done. "Reflecting changes back to Upload field after the file has been edited" didn't seem to be worth fixing.

I've rebased and squash the PR. It's ready for re-review.

maxime-rainville:
FYI The build won't pass until to related PRs are merged.

tests/php/Forms/FileFormBuilderTest.php
```diff
@@ -46,6 +46,15 @@ public function setUp()
 
     public function tearDown()
     {
+        // Unpublishing these files is necessary because when this test is run through the recipe-cms phpunit suite,
```
emteknetnz, line 49: :-/

I get this approach though it's not resolving what seems like a tech-debt situation

Looking at SearchFormTest.php
```
 * Because this manipulates the test database in severe ways, I've renamed the test to force it to run last...
 */
class ZZZSearchFormTest extends FunctionalTest
...
    public static function setUpBeforeClass()
    {
        // HACK Postgres doesn't refresh TSearch indexes when the schema changes after CREATE TABLE
        // MySQL will need a different table type
        static::$tempDB->kill();
        Config::modify();
        FulltextSearchable::enable();
        static::$tempDB->build();
        static::resetDBSchema(true);
        parent::setUpBeforeClass();
    }
```

Seems quite wrong that a unit tests can affect other unit tests.  Would it be possible instead to somehow make what's going on in ZZZSearchFormTest less nasty?

maxime-rainville: Not really. `FulltextSearchable::enable()` is not designed to be disabled after it's been enabled. I had a go at getting this working last time I came across this issue and didn't come to an obvious way of fixing it.

And FulltextSearchable is kind of a legacy feature we would probably dump if we had the option right now, so I can't see the logic on spending a lot of time debugging it's current implementation.

My instinct would be to disable this test if it's run as part of the `recipe-cms` suite to avoid these kind of errors going forward.

emteknetnz: >> My instinct would be to disable this test if it's run as part of the recipe-cms suite to avoid these kind of errors going forward.

Yup agree there.  I've spend some time trying to replicate this issue locally and I failed to do so.  It's a strange edge-case and probably not worth investing time in fixing properly.

@maxime-rainville do you know of a method to specifically disable this test in a recipe-cms context?

client/src/containers/AssetAdmin/Breadcrumb.js
```diff
@@ -0,0 +1,90 @@
+/* global alert, confirm */
+import React from 'react';
+import PropTypes from 'prop-types';
+import i18n from 'i18n';
+import { Component as PlainBreadcrumb } from 'components/Breadcrumb/Breadcrumb';
```
emteknetnz, line 5: So this means we would now have two different things called 'Breadcrumb'?

That's going to cause confusion down the road

We should rename the AssetAdmin Breadcrumb to something else like FolderBreadcrumb

client/src/containers/Editor/Editor.js
emteknetnz: your github comment says it much better :-) could just copy paste that in

client/src/containers/Editor/tests/EditorHeader-story.js
emteknetnz: we should now delete this file?

code/Forms/FileFormFactory.php
```diff
@@ -58,15 +49,22 @@ protected function getFormFieldTabs($record, $context = [])
         // All non-admin forms are typically readonly
```
emteknetnz, line 49: Can remove this comment

code/Forms/FileFormFactory.php
```diff
@@ -58,15 +49,22 @@ protected function getFormFieldTabs($record, $context = [])
...
+            default:
+                $tabs->push($this->getFormFieldDetailsTab($record, $context));
+                $tabs->push($this->getFormFieldSecurityTab($record, $context));
+                $tabs->push($this->getFormFieldUsageTab($record, $context));
+
+                if ($this->config()->get('show_history')) {
+                    $tabs->push($this->getFormFieldHistoryTab($record, $context));
+                }
+
+                $tabs->setReadonly($type === static::TYPE_SELECT);
```
emteknetnz, line 67: read better then old way i.e.
```
            case static::TYPE_SELECT:
                $tabs->setReadonly(true);
```

client/src/containers/AssetAdmin/Breadcrumb.js
maxime-rainville: This one is not exposed out of asset-admin, so other modules won't be able to access it. It's only going to be used in the AssetAdmin component.

I see this as having two classes with the same name but under different namespaces.

emteknetnz: Yeah it's still confusing though.  I spent about 10 minutes thinking that you'd moved the Breadcrumb Component out of the admin module and put it here :-)

client/src/containers/Editor/tests/EditorHeader-story.js
maxime-rainville: Done.

client/src/containers/Editor/Editor.js
maxime-rainville: Done.

code/Forms/FileFormFactory.php
maxime-rainville: Done.

client/src/containers/AssetAdmin/Breadcrumb.js
maxime-rainville: Done.

.travis.yml
```diff
@@ -59,7 +59,7 @@ before_script:
 # Install composer
   - composer validate
   - if [[ $DB == PGSQL ]]; then composer require --no-update silverstripe/postgresql:2.x-dev --prefer-dist; fi
-  - composer require silverstripe/recipe-testing:^1 silverstripe/recipe-cms 4.x-dev --no-update --prefer-dist
+  - composer require silverstripe/recipe-testing:^1 silverstripe/recipe-cms:4.x-dev silverstripe/frameworktest:dev-master --no-update --prefer-dist
```
emteknetnz, line 62: As discussed on slack:
- including frameworktest as dev-dependency in composer WILL NOT require frameworktest unless it's the in the project root.
- means it won't do anything on your local if you're doing development from the vendor folder (as most of us are)
- in the case of asset-admin being the root project, which it will in travis, it is functionally the same, though it's clearer being in travis.yml because it shows the intent better 

Seems like the bigger question though is why is frameworktest a requirement for behat tests in the first place?  I've always seen frameworktest as a quick and dirty way to create lots of records for manual testing.  We're including dev-master here because we don't bother versioning frameworktest

Rather then rely on framework test for behat fixture creation, seems like it would be much cleaner to create them the old fashioned way i.e. yml

maxime-rainville: There's a precedent for using frameworktests ... it gets installed by the CWP kitchen sink.

I can't just rely on creating fixtures with YML because I need an DataObject with an `has_many` relationship to Image or File. Off the top of my head, I don't think we have a recipe that has that particular set up out of the box

emteknetnz: You can probably guess what I'm going to say next :-)

Just create a TestOnly DataObject in asset-admin and use that.  Much cleaner and safer than using frameworktest that has no versioning and could change at any point.

emteknetnz:
Fairly minor UX issue, happy for this to be created as a new issue resolved sometime later on post-merge.

When I edit a page
And I use TinyMCE
And I click the insert an image button
And I upload a new image
And I click the "Details" button in the top right corner
And I change the "Title" of the image
And I click the "Save" button
Then I expect to see the green button to have the text "Insert image"
But actually it has the text "Update image"

emteknetnz:
@maxime-rainville 
One issue we should resolve before merging - the datetime fields on the image are now editable.  They should be read-only

image

On the main branch, they are read-only in the details tab

image

clarkepaul:
> Then I expect to see the green button to have the text "Insert image". But actually it has the text "Update image"

@emteknetnz I can't see the actual experience but from what I recall... as the image is already inserted, you are just updating it. If you were to remove it and re-insert it you get the "insert" button label again.

emteknetnz:
> I can't see the actual experience but from what I recall... as the image is already inserted, you are just updating it. If you were to remove it and re-insert it you get the "insert" button label again.

That's correct for already inserted images.  However you can still get this 'Update image' (or 'Update file' for pdfs) for "uninserted" Images/Files.  It still functions correctly as it will insert the file, it's just it shows the wrong text

maxime-rainville:
I've got a fix for the two fields that should be readonly.

@clarkepaul Would you have a problem if updated the modal so that the button always says "Insert" even if you are updating an image that has previously been inserted? The logic for this part is kind of clunky (which is why it's broken in the first place). I could probably make it work at the cost of adding an extra layer of clunkiness.

If we're OK with the button always saying "Insert", I can actually remove some clunky code.

maxime-rainville:
Just had a Zoom catch up with Paul to discuss how to handle the Insert/Update button issue. The 3 options discussed were:
1. Have the button to always display "Insert" no matter the context.
2. Me spending a couple hours to fix the bug with the button by adding some extra hackish logic on top of the existing hackish logic.
3. Live with the bug.

Paul says he's happy to live with the bug since it's not on the primary path most users will take and it's kind of minor anyway.

clarkepaul:
@maxime-rainville has just walked me through the UX button label bug. I'm okay with leaving the bug as-is for now if we can raise it as a separate issue that would be good.
My reasoning is that for the two most common user paths of:
- inserting an image, the language "insert" makes the most sense
- editing a file when it has already been inserted, the language "update" makes the most sense

From my experience, cms users are more concerned about the presentation of a file initially, editing the file details is a secondary thought (although the main benefits of this issue). Returning without making any edits at all would be subset of people. For the remaining people, I think they would complete the intended action of inserting the file with the button label being "Update" without too many concerns.

Obviously I'd prefer it fixed but I don't see this as a blocker to getting the issue across the line.

sachajudd:
Nice work everyone!! 🎉

purplespider:
This is nice! But may I suggest changing the "**Details**" button to say "**Edit**" instead? To make it more obvious how to edit the fields? 

I'd expect a "Details" button to provide me with more, erm, details, but this button makes the existing fields editable.

Screenshot 2020-06-01 at 16 35 27@2x

clarkepaul:
Much discussion and back and forth on this one. 
- The fields in the current placement view are already editable, you might think you need to click _Edit_ to adjust the current fields in view already.
- With a label _Edit original_ we had comments from users like "What's the original file? why would I want to edit a different file?" The concept of original made it feel like its a copy rather than the same file.

The button takes you to the file _Details_ tab, so having a button with the same name made some sense here. _Edit details_ was also considered but it tested fine without the word edit. This was the terminology which was already being used navigating between the _Placement_ tab and _Details_ tab.

This one doesn't the best experience you could ever want but hopefully it's an improvement from what was there.

purplespider:
Ah, thanks Paul, after reading your comments, I've realised this button appears in two different screens:

---

### 1. Above the Placement tab, when dealing with a WYSIWYG image:

Screenshot 2020-06-02 at 09 42 07@2x

In this context, everything you said makes sense and changing it to just say "Edit" would be unclear.

---

### 2. Above a read-only Details tab, when dealing with an image in an UploadField:

Screenshot 2020-06-02 at 09 43 37m@2x

It is in this context, where the **Details** button is currently confusing. As you are already viewing the Details tab.

---

I presume there is a reason why the fields just aren't immediately editable in the latter view. So, perhaps the best solution is to change the button to say **Edit Details** so that it's action is clear in both contexts?

Issue #1026, opened 2019-10-23
Appending file version id to image url breaks Protected Asset previews with SilverStripe-S3


obj63mc:
Silverstripe version 4.3+

Currently when an asset is rendered via the cms it appends ?vid=X to the end of the image url.

When you utilize the silverstripe-s3 module this append version to the signed url causes the image to not render.

Exmaple of line that causes the issue - https://github.com/silverstripe/silverstripe-asset-admin/blob/e0f919d3c6a6b0415ae06c9e91da5bf37da99672/client/src/components/PreviewImageField/PreviewImageField.js#L265

This is related to issue https://github.com/silverstripe/silverstripe-s3/issues/33
This also appears to affect UploadFields in the CMS as well (really anywhere assets are rendered that are 'protected' assets).  Looks like https://github.com/silverstripe/silverstripe-asset-admin/commit/52ae679d1d246b5e935efd8e449eb9a95f2c9703 introduced this issue.

# Pull requests
* ~https://github.com/silverstripe/silverstripe-asset-admin/pull/1039~
* [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1040
* [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1119
* [ ] https://github.com/silverstripe/silverstripe-s3/pull/43

obj63mc:
One possible quick solution for this would be to check if a querystring exists on the url, if it does don't append the version id.

maxime-rainville:
This problem is sort of related to https://github.com/silverstripe/silverstripe-versioned-admin/issues/134 in that if we were using a sensible library to generate our URLs we would avoid this kind of errors.

maxime-rainville:
https://github.com/silverstripe/silverstripe-asset-admin/pull/1040 should fix it. It uses a smarter logic to build that URL that should be OK with pre-existing query param.

emteknetnz:
The linked pull-request has been merged, I'll close this issue now

maxime-rainville:
Looks like making the logic to generate URL better doesn't fix the S3 issue because adding the `vid` invalidates the S3 signature.

https://github.com/silverstripe/silverstripe-asset-admin/pull/1040#issuecomment-660741105

maxime-rainville:
Just looking at the possible solution for this. Of the top of my head, we could try this:
[...]

PR #1039, opened 2019-12-20
BUG Properly parse preview image url before busting the cache

maxime-rainville:
We were manually updating the url for the preview image. This is causing problems for people using the S3 module because the S3 preview URL already has query parameter.

This PR use the `url` and `qs` libs to properly parsed and rebuild the url.

This PR builds on the work introduce by https://github.com/silverstripe/silverstripe-versioned-admin/issues/134

# Parent issue
* https://github.com/silverstripe/silverstripe-asset-admin/issues/1026


maxime-rainville:
Actually, I'll backport this to 1.4

PR #1040, opened 2019-12-20
BUG Properly parse preview image url before busting the cache

maxime-rainville:
We were manually updating the url for the preview image. This is causing problems for people using the S3 module because the S3 preview URL already has query parameter.

This PR use the `url` and `qs` libs to properly parsed and rebuild the url.

This PR builds on the work introduce by https://github.com/silverstripe/silverstripe-versioned-admin/issues/134

# Parent issue
* https://github.com/silverstripe/silverstripe-asset-admin/issues/1026


maxime-rainville:
NB: When compiled on 4.4 or 4.5, the url lib will be shipped with admin. In 4.6, the url lib is bundled with admin.

apuckey:
Bump :)

maxime-rainville:
I'm waiting a for a peer-review. I just asked some colleagues to take a look.

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -233,12 +235,25 @@ class PreviewImageField extends Component {
...
     return null;
   }
 
+  /**
+   * Append a vid parameter to the URL to bust the cache
+   * @param {string} url
+   * @param {string} vid
+   * @return string
+   */
+  cacheBustUrl(url, vid) {
```
emteknetnz, line 250: `vid` is not a very descriptive variable name.  Can we change this to versionID?

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -233,12 +235,25 @@ class PreviewImageField extends Component {
...
+    const parsedURL = urlLib.parse(url);
+    const parsedQs = qs.parse(parsedURL.query);
```
emteknetnz, line 252: `parsedQS` to match `parsedURL` above it

or maybe change to `parsedQS` and `parsedUrl` :-)

maxime-rainville:
Rebased and fixed.

emteknetnz:
@maxime-rainville Travis is not happy

This is an old version of asset-admin, so probably cannot just copy paste the latest `.travis.yml` from `1.6` since we removed php 5.6 from that

maxime-rainville:
I've retargeted to 1.5. We're not planning on doing another 1.4 release anyway.

emteknetnz:
OK tested, looks good

I tested by temporarily changing

https://github.com/silverstripe/silverstripe-asset-admin/blob/1/code/Forms/PreviewImageField.php#L50
to
```
'preview' => $record->PreviewLink() . '?abc=123',
```

The preview image src correctly rendered as 
`<mydomain>/assets/7e8b31efe2/sample-tui__FitMaxWzkzMCwzMzZd.jpg?abc=123&vid=2`

I compared this to another site pre-patch and it rendered as
`<mydomain>/assets/548acca037/sample-morepork__FitMaxWzkzMCwzMzZd.jpg?abc=123?vid=2`

Will merge now

emteknetnz:
Actually @maxime-rainville before I merge this, because this originally targeted 1.4, though not it targets 1.5, could I get you to confirm the following:
- you rebuilt asset-admin dist files after rebased to 1.5 asset-admin (instead of 1.4)
- in the corresponding admin module you had 1.5 checked when you ran `yarn build` (instead of 1.4)

maxime-rainville:
@emteknetnz  Yes, I rebuilt it. There's an entry in the travis build that validates this, by rebuilding the distribution files. If any of the distribution file changes, the build fails.

apuckey:
Hi,
Testing this this morning and it still breaks a s3 pre-signed url. Looks like adding anything to the url after its signed breaks it completely, however the url is different everytime its generated to does not need to be cachebusted. Possible to add an exception if it detects this is a pre-signed url?

perhaps by looking to see if the url querystring contains `X-Amz-Signature` then jsut dont modify it at all?

There could also be another usecase for google cloud storage but im not familiar with how that works.

maxime-rainville:
@apuckey I've reopen the parent issue.

PR #1046, opened 2020-01-21
Added missing translations

ovogel-wigital:


robbieaverill:
Thanks for the contribution! Our non-English translation files are all managed via Transifex, which is moderated by a community. These translations are periodically pulled in from Transifex to the various SilverStripe modules when releases happen. 

If you wouldn't mind, could you please make these changes in Transifex? https://www.transifex.com/silverstripe/silverstripe-asset-admin

Thanks again.

PR #1052, opened 2020-01-30
implode in _t() breaks i18nTextCollectorTask

lpostiglione:
implode in _t() breaks i18nTextCollectorTask

### DIFF
--- a/code/Forms/ImageFormFactory.php
+++ b/code/Forms/ImageFormFactory.php
@@ -87,10 +87,8 @@ class ImageFormFactory extends FileFormFactory
 
         $altTextDescription = _t(
             'SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.AltTextTip',
-            implode([
-                'Description for visitors who are unable to view the image (using screenreaders or ',
-                'image blockers). Recommended for images which provide unique context to the content.',
-            ])
+            'Description for visitors who are unable to view the image (using screenreaders or ' .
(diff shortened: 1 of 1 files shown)


Cheddam:
Hey @lpostiglione, thanks for raising this PR! Since it's a fix with no new functionality, it makes sense to target it to the `4.5` branch rather than the `4` branch, so that it can be included in the next patch release (`4.5.1`). Would you be able to retarget it?

brynwhyman:
@Cheddam I think you meant to say:
> it makes sense to target it to the ~4~ 1.5 branch rather than the ~4~ 1 branch, so that it can be included in the next patch release ~(4.5.1)~ (1.5.1). Would you be able to retarget it?

lpostiglione:
@Cheddam @brynwhyman 

I am no GitHub expert but I feel like what I did is not exactly a good idea?

Maybe you could tell me how to retarget this correctly?

Cheddam:
@lpostiglione Since you raised this directly via GitHub, it might be easiest to raise a new PR by making the change directly to the `1.5` branch. Otherwise, if you're confident enough to use the command line, you can clone your fork of the repository and rebase your branch onto `1.5` by hand using the following commands:

```
git clone git@github.com:lpostiglione/silverstripe-asset-admin.git
cd silverstripe-asset-admin
git checkout patch-1
git rebase --onto=1.5 HEAD~1
git push --force-with-lease
```

lpostiglione:
@Cheddam Thank you very much, should be ready now.

PR #1055, opened 2020-02-05
Do not render ImageSizePresentList react component for remote files

emteknetnz:
Fix for https://github.com/silverstripe/silverstripe-cms/issues/2498


client/src/components/ProportionConstraintField/ProportionConstraintField.js
```diff
@@ -148,6 +148,8 @@ class ProportionConstraintField extends Component {
       current: { width: currentWidth },
       imageSizePresets } = this.props;
 
+    const isRemoteFile = this.props.data.hasOwnProperty('isRemoteFile') && this.props.data.isRemoteFile;
```
unclecheese: This can probably be expressed as `const { isRemoteFile } = this.props.data`

emteknetnz: it'll be undefined sometimes, does it matter? (old browsers maybe?)

unclecheese: Pretty sure Babel handles all those checks. I've never seen this check done before, but I can't say 100%.

emteknetnz: Have moved to the const above `data: { originalWidth, isRemoteFile },`.  Have tested in IE11, works fine without the hasOwnProperty check

PR #1058, opened 2020-02-12
Stop an image's "Title text (tooltip)" being set to the filename

purplespider:
When any image is inserted into the WYSIWYG, it sets the "Title text (tooltip)" field to the image's Title, which is usually just it's filename. This isn't desired or useful in most cases, and usually results in ugly tooltips such as "IMG12345" or "image34" appearing when site visitors hover over an image.

This PR stops the "Title text" being set automatically, to ensure that any tooltips are added on purpose.

Resolves: https://github.com/silverstripe/silverstripe-assets/issues/129

### DIFF
--- a/code/Forms/ImageFormFactory.php
+++ b/code/Forms/ImageFormFactory.php
@@ -98,7 +98,7 @@ class ImageFormFactory extends FileFormFactory
             $titleField = TextField::create(
                 'TitleTooltip',
                 _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.TitleTooltip', 'Title text (tooltip)')
-            )->setValue($record->Title)
+            )
         );
 
         $titleDescription = _t(


emteknetnz:
@purplespider Thanks for the pull request

I just tested this to see what behaviour is being modified - it's triggered on hover after about 500 milliseconds

The filename of the uploaded file is "sample-kokako.jpg"

image

It does seem fairly pointless and kind of annoying.  I'll run this past our UX team to see if they want it gone or if they want to keep it

emteknetnz:
@sachajudd @clarkepaul do we want to keep that hover tooltip on images or get rid of it?

sachajudd:
I've just added the related issue to your description @purplespider. Cheers for this, looks good 👍

emteknetnz:
@purplespider Thanks for your contribution, I've merged this in

PR #1059, opened 2020-02-17
BUG Move the query resolution after the DataListQuery has been altered

maxime-rainville:


maxime-rainville:
Moving the PR to a different repo.

PR #1061, opened 2020-02-18
BUG Tweak UsedOnTableTest to dynamically switch protocol

maxime-rainville:
This test occasionally fails because it expect the protocol to always be HTTP. I've added a is_https check to swap protocol.

### DIFF
--- a/tests/php/Forms/UsedOnTableTest.php
+++ b/tests/php/Forms/UsedOnTableTest.php
@@ -4,6 +4,7 @@ namespace SilverStripe\AssetAdmin\Tests\Forms;
 
 use SilverStripe\Admin\Forms\UsedOnTable;
 use SilverStripe\Assets\File;
+use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Dev\SapphireTest;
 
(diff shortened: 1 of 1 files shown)


maxime-rainville:
Test didn't exist in 1.4.

tests/php/Forms/UsedOnTableTest.php
```diff
@@ -22,7 +23,17 @@ public function testExcludes()
...
-        $expected = '{"usage":[{"id":0,"title":"My Page","type":"Page","state":"Draft","link":"http:\/\/localhost\/admin\/pages\/edit\/show\/1"}]}';
+        $protocol = Director::is_https() ? 'https' : 'http';
+
+        $expected = json_encode([
+            "usage"=> [[
+                "id" => 0,
+                "title" => "My Page",
+                "type" => "Page",
+                "state" => "Draft",
+                "link" => sprintf('%s://localhost/admin/pages/edit/show/1', $protocol)
```
emteknetnz: link had escaped slashes in previous version, they're not escaped here.  is that intentional?

maxime-rainville: The previous string was raw json. Because, I switch to using an array and JSON encoding it, I'm letting json_encoding do the escaping.

Issue #1030, opened 2019-11-06
Bug: Multi-File Upload - Upload Cancelled


jchenevey:
Affects SilverStripe 4.4.4.

Full composer list:

```
benmanu/silverstripe-simple-styleguide         2.0.5        Provides a kitchen sink template useful for theme development.
colymba/gridfield-bulk-editing-tools           3.0.0        SilverStripe GridField component to upload images/files and edit records ...
innis-maggiore/silverstripe-content-extensions 4.0          Content Extensions
innis-maggiore/silverstripe-smtp-tester        4.0          Silverstripe SMTP Tester allows developers to test their site's SMTP conf...
innis-maggiore/silverstripe-social-meta-tags   4.0.2        Adds social media meta tags to Silverstripe sites.
phpunit/phpunit                                5.7.27       The PHP Unit Testing framework.
silverstripe-themes/simple                     3.2.0        The SilverStripe simple theme (default SilverStripe 3 theme)
silverstripe/blog                              3.4.0        A fresh take on blogging in Silverstripe set out to tackle the issue of a...
silverstripe/comment-notifications             2.0.1        Provides email notifications for visitor comments
silverstripe/comments                          3.2.1        This module provides commenting functionality for Pages and other DataObj...
silverstripe/recaptcha                         3.0.0-alpha1 Provides a form field which allows form to validate for non-bot submissio...
silverstripe/recipe-cms                        4.4.4        SilverStripe recipe for fully featured page and asset content editing
silverstripe/recipe-plugin                     1.3.0        Helper plugin to install SilverStripe recipes
silverstripe/spamprotection                    3.0.1        Spam protection module for SilverStripe.
silverstripe/userforms                         5.4.1        UserForms enables CMS users to create dynamic forms via a drag and drop i...
silverstripe/vendor-plugin                     1.4.0        Allows vendor modules to expose directories to the webroot
silverstripe/versioned                         1.4.3        SilverStripe Versioned component
silverstripe/widgets                           2.0.2        Widgets are small pieces of functionality such as showing the latest Comm...
stripe/stripe-php                              v6.43.1      Stripe PHP Library
suin/php-rss-writer                            1.6.0        Yet another simple RSS writer library for PHP 5.4 or later.
symbiote/silverstripe-gridfieldextensions      3.2.1        A collection of useful grid field components
undefinedoffset/sortablegridfield              2.0.8        Adds drag and drop functionality to SilverStripe 4.x's GridField
[...]

PR #1063, opened 2020-02-20
Fix last file upload showing as errored when uploading multiple files.

bergice:
A race condition keeps the last file in the upload queue as still active when destroying the dropzone. This fix removes the files before unmounting to prevent errors showing when unmounting.

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/1030


emteknetnz:
@bergice how come you're targetting `1` instead of `1.5`? seems like this is a non-breaking change and we could release it as a patch version?

bergice:
> Good job. That must have been a tough one to figure out. I've tested it in Firefox, Chrome, IE11 and Edge.
> 
> As we've discussed on Slack the other day, it's not really worth adding a behat test for this exact scenario.
> 
> @bergice Can you confirmed that you've added the multi-upload scenario to our Cucumber/HipTest test suite?

Yep, it was already there.

PR #1073, opened 2020-04-14
BUG: Legacy max upload size setting removal

mfendeksilverstripe:
* `allowed_max_file_size` setting removed as it's unused
* documentation updated for `max_upload_size` setting

## Related issues

https://github.com/silverstripe/silverstripe-asset-admin/issues/1074

### DIFF
--- a/code/Controller/AssetAdmin.php
+++ b/code/Controller/AssetAdmin.php
@@ -87,13 +87,6 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
 
     /**
      * @config
-     * @see Upload->allowedMaxFileSize
-     * @var int
-     */
-    private static $allowed_max_file_size;
-
-    /**
-     * @config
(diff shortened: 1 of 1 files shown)


dhensby:
What's the actual impact of these changes? Are we removing or changing anything that has an effect or is this code clean up?

Either way, I'm not sure this should go in a patch release (1.5.x) It should either be 1.x or 2.0...

mfendeksilverstripe:
@dhensby As far as I can see the `allowed_max_file_size` is not used anywhere and can be safely removed so there should be no functional impact. A patch release should be safe. The code documentation update simply makes the use of `max_upload_size` less confusing.

NPM test is failing but I don't think this is related to this change set.

```
The command "if [[ $NPM_TEST ]]; then git diff-files --quiet -w --relative=client; fi" exited with 1.
```

Issue #1003, opened 2019-09-17
Files and folder are not aligned


newleeland:
This is a regression. Extra CSS margins are being added. This make the Files area look broken for the user.

This is occuring in the bambusa demo as well as local SS4.4 environments

![Screen Shot 2019-09-17 at 3.58.47 PM.png](https://images.zenhubusercontent.com/596d365561b4ca3b0a74bd86/b7654559-4bbc-49b4-a918-319841d9d6ff)

clarkepaul:
There is markup and classes which aren't needed, so applies duplicate CSS spacing. 
There are duplicate ID's which will show as an issue as well, although it still seems to work from my quick test although this wasn't thoroughly tested.
<img width="401" alt="Pasted_Image_13_05_20__10_33_AM" src="https://user-images.githubusercontent.com/555033/81752406-7ba4e280-9505-11ea-8b1c-d1b47cc9e745.png">

Fix for the visual aspects, folders require both draggable and dropable functionality but don't need both classes to have the margin. Since all folders are draggable we can remove the margin from dropable items. Ideally, the two classes should probably appear on the same element then we wouldn't have this issue in the first place. Either way remove the dropable items from having the margin...
.gallary-item__draggable
~~.gallary-item__dropable~~ { margin... }

robbieaverill:
Fixed in #1089, thanks @sachajudd

PR #1089, opened 2020-05-19
FIX Remove direct descendant selector to apply correct margins

sachajudd:
Since `gallery-item__draggable` was not a direct descendant of `gallery-item__droppable` (`selectableItem-#` was) the removal of the margin was not being applied. I've removed the direct descendant selector so this now functions and fixes the extra margin seen on the frontend.

Fixes: https://github.com/silverstripe/silverstripe-asset-admin/issues/1003


robbieaverill:
I restarted the builds and they're green now - all good

PR #1090, opened 2020-05-26
FIX Asset revision timestamps are no longer underlined in asset admin history tabs

### LINKED ISSUE #1088: Version's relative time in History viewer appears as a link (but is not)
In the history tab, the relative time a version was created looks like a link (underline and cursor: pointer) but doesn't perform any action.

Screen Shot 2020-05-11 at 9 08 51 PM

I'd suggest changing the styling to remove the link-like look.

robbieaverill:
I also fixed the label/tag classes (should be badge) which broke when we were updating Bootstrap back in 4.2 ish.

image

Fixes #1088


emteknetnz:
Possibly dumb question, how do I see the (file?) history tab?  I've tried uploading and publishing an image, then replacing and publishing the image with another one.  I don't see a history tab in `/admin/assets/show/0/edit/1`

robbieaverill:
Yeah it took me a bit to work it out too, you have to enable it in config: https://github.com/silverstripe/silverstripe-asset-admin/blob/1/code/Forms/FileFormFactory.php#L30

emteknetnz:
Cool, replicated original issue and tested PR, nice work.

PR #1091, opened 2020-05-26
NEW Last modified date for assets in list view table are now formatted according to browser locale

### LINKED ISSUE #990: date displayed is a bit techy
image
Can we change this to something like:

```
Monday 12 June 13:45
```
?

robbieaverill:
image

Note: I'm in North America, so day/month is backwards.

There was a todo already to do this.

Resolves #990


bergice:
Codewise this is looking good. But I'm getting inconsistencies between the front-end and back-end locales:

image

robbieaverill:
Good point. I suppose we'd need to update the readonly DatetimeField so it uses the browser locale as well... thoughts @silverstripeux?

clarkepaul:
Sounds logical @robbieaverill

robbieaverill:
Cool. I've raised https://github.com/silverstripe/silverstripe-admin/issues/1044 for that. @bergice are you happy for that to be resolved separately to this PR?

Issue #1093, opened 2020-05-29
Pagination only work intermitentally


maxime-rainville:
Looks like our pagination for asset-admin is kind of hit and miss. Clicks on the pagination element only seem to register half the time.

Interestingly, everything work just swell if you use the keyboard.

https://youtu.be/dP8I5iSVNJE

Tested in 4.4, 4.5, 4.6.

# Pulls request
* [x] https://github.com/unclecheese/react-selectable/pull/64
* [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1095

maxime-rainville:
Looks like the problem is with [griddle-react](https://www.npmjs.com/package/griddle-react). We're on 0.8.2. Latest release is 1.13.1.

Doesn't look like we're doing anything wrong on our side. It's just that Griddle's `<GridPagination />` just doesn't seem to be calling our handlers.

I tried just installing 1.13.1 and got errors thrown in my face, so it doesn't look like it's a trivial upgrade. Even if we do manage to upgrade, it seems like quite a jump to do in a patch release.

maxime-rainville:
I've been investigating this a bit further. The problem is not actually `griddle-react` at all. 
It's with react-selectable. React selectable register any click event on Gallery as an attempt at selection which causes `ThumbnailView` to re-render.

So when you click on the pagination button, `react-selectable`  gets the event as well. This starts are race between `react-selectable` and the pagination button on who can handle the click event first. When `react-selectable` "wins" that race `ThumbnailView` gets re-rendered which prevents the Pagination button from calling its own handler.

maxime-rainville:
If this gets merged https://github.com/unclecheese/react-selectable/pull/64, I should be able to patch asset admin to cancel selection when the starting mouse over event is over the pagination.

Aaron suggested switching to https://react-selectable-fast.now.sh/. That's a good idea, but it's not a straight drop in replacement and it uses react hooks not available in 4.4/4.5. Should look at this at some other stage.

PR #1095, opened 2020-06-01
BUG Prevent react-selectable from interfering with pagination

maxime-rainville:
This hooks into a tweak API from https://github.com/unclecheese/react-selectable/pull/64 to prevent the selection from beginning when the mouse down occurs on the pagination.

# Parent issue
* https://github.com/silverstripe/silverstripe-asset-admin/issues/1093


bergice:
Tested and works fine.

Issue #1086, opened 2020-05-03
Insert/Update image button text


emteknetnz:
Minor UX issue that came about after merging the inline editing PR
https://github.com/silverstripe/silverstripe-asset-admin/pull/1038

When I edit a page
And I use TinyMCE
And I click the insert an image button
And I upload a new image
And I click the "Details" button in the top right corner
And I change the "Title" of the image
And I click the "Save" button
Then I expect to see the green button to have the text "Insert image"
But actually it has the text "Update image"

I believe it also affects 'Insert/Update file' when dealing with a PDF

We decided to leave this issue in place and merged anyway as it wasn't deemed as overly-important https://github.com/silverstripe/silverstripe-asset-admin/pull/1038#issuecomment-623192832

### PR
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1115

### Notes
* We've found that there's an issue with state management. The button text is being determined client-side so may be subject to having the state potentially cached and not always showing the correct button text for the context.
* When looking at this, we'll want to tidy up another area that @maxime-rainville has suggested (please add context here :) )

maxime-rainville:
The underlying problem is caused by this bit: https://github.com/silverstripe/silverstripe-asset-admin/blob/1/client/src/transforms/AssetAdmin/insertAssetModal.js

This logic gets run when fetch the FileInsertMedia form schema. It checks if there's any form value override for that form schema. When you're inserting a new image there's no overrides to provide, so we use the "insert" label. When you're updating an existing image, the short code info in the WYSIWYG gets pushed to the modal as a form value overrides.

The problem is that when you edit the original file, you're placement info gets recalled with form value overrides. Here's what happens in the background:
* User clicks the "insert image" button. No form value overrides are pushed to the modal.
* Modal gets displayed.
* User selects an image. We fetch the form schema. No override are apply.
  * The `insertAssetModal` transformation is applied. Because there's no overrides, we set the button label to "Insert".
* User decides to edit the original info of the image. Current placement info is stash to the redux store.
* User returns to insert media form.
  * We restore the placement info to the form with form value overrides
  * We refetch the insert media schema.
  * The `insertAssetModal` transformation is applied. Because there IS overrides this time, we set the button label to "Update".

maxime-rainville:
[...]

PR #1114, opened 2020-07-09
FIX Use correct button text if user is inserting or updating a file

emteknetnz:
**Update:**
This alternative PR is a bit more proper - https://github.com/silverstripe/silverstripe-asset-admin/issues/1086

Will probably close this shortly

---

Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/1086


emteknetnz:
I realise this is kind of unorthodox

I did initially attempt to read this TinyMCE_ssmedia.js and pass it to InsertMediaModal > AssetAdmin > Editor > though then it used the abstract FormBuilderLoader which XHR's back to PHP.

We could in theory pass this variable around like this and back to the PHP, get the button text set properly in PHP, and potentially throw out `insertAssetModal` altogether. though I don't know if it's used in contexts outside of inserting files via TinyMCE (file fields?)

It just seems much easier to just put in a handful of lines in a single JS file

PR #1115, opened 2020-07-09
FIX Show different action label if file is selected by user

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/1086


emteknetnz:
behat has this:
image

need to merge the framework PR which has the step defined and rerun travis

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -60,7 +60,12 @@ const filter = 'img[data-shortcode="image"]';
 
       ed.addCommand('ssmedia', () => {
         // See HtmlEditorField.js
-        jQuery(`#${ed.id}`).entwine('ss').openMediaDialog();
+        // work out if a file is selected within the tinymce iframe by checking
+        // the css class on the 'Insert from Files' button
+        const fileSelected = jQuery(ed.container)
```
maxime-rainville, line 65: You don't need this bit ... you just need to check the `ID` attribue from the results of `getOriginalAttributes` when creating your Modal.

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -185,9 +191,10 @@ jQuery.entwine('ss', ($) => {
...
      *
      * @param {boolean} isOpen
+     * @param {boolean} fileSelected
      * @private
      */
-    _renderModal(isOpen) {
+    _renderModal(isOpen, fileSelected = false) {
       const handleHide = () => this.close();
       const handleInsert = (...args) => this._handleInsert(...args);
       const { url, ...attrs } = this.getOriginalAttributes();
```
maxime-rainville, line 200: Just check if `attr.ID !== null` and that will tell you if `fileSelected` should be true or false.

tests/behat/features/insert-an-image.feature
```diff
@@ -49,14 +50,27 @@ Feature: Insert an image into a page
...
       And I should see "File One" in the ".editor__heading" element
     When I press the "Details" button
       Then I should see the "Form_fileEditForm" form
     When I click the ".editor-header__back-button" element
       Then I should see the "Form_fileInsertForm" form
-    When I press the "Update file" button
+    When I press the "Insert file" button
       Then the "Content" HTML field should contain "file1.jpg"
       And the "Content" HTML field should contain "My alt"
+      # Required so that we click the correct save button below
```
maxime-rainville, line 62: That's going a bit beyond the call of duty, but it's a nice touch.

client/src/entwine/TinyMCE_ssmedia.js
emteknetnz: yup nice, can actually do this all the way down in `AssetAdmin.js` which significantly reduces the amount files we need to pass `fileSelected` around in.  Also don't need the PR in the admin module. Have updated PR

client/src/entwine/TinyMCE_ssmedia.js
emteknetnz: have implemented this in `AssetAdmin.js` instead

emteknetnz:
@maxime-rainville I've implemented your suggestion to use the file.ID, simplifies this PR quite a bit

>     * Choose a different image ... the button still says update, the expectation is that it should go back to saying insert ... at least that's what it was doing before we added in-modal editing.

This behave is still the same after the refactor.  However I do think this new behaviour is correct. If you have selected an existing image, then you select a different image, as a user I would expect to see 'Update image' because you are updating the existing image with a different one.  'Insert image' does not make sense.

If anything it should say 'Replace image', though if we were to do that I'd say create a new issue since that is new functionality.  'Update' is closer to 'Replace' then 'Insert' is, so I say stick with 'Update'

maxime-rainville:
@silverstripeux Do you have any views on this dilemma:
- Author is editing some content in a WYSIWYG and the WYSIWYG contains a pre-existing image
- Author click the image and start editing the placement information of the image. The primary button in the modal says "Update image".
- Author picks a different image. Should the primary button now say "Update image" or "Insert image"?

emteknetnz:
The behat failure is quite nasty

What's looks to be happening is that the redux overrides are not getting applied because they have an id of

`admin/assets/schema/fileEditorLinkForm/3`

instead of the new form schema url which is

`admin/assets/schema/fileEditorLinkForm/3?fileSelected=true`

I don't *think* there was this issue when this PR was in its previous state, though I'm not sure. I'll need to revert to this to check though
https://github.com/silverstripe/silverstripe-asset-admin/commit/2601bac2e06f1d2bb5687037ec89e1841fa5bfe9

clarkepaul:
> Author picks a different image. Should the primary button now say "Update image" or "Insert image"?

I think it should say "Update image"...  short for "Update existing image with some new stuff"

clarkepaul:
The only time it would say "Insert" is if you are placing a file in a location that doesn't already contain a file, otherwise it would always be "Update". That makes the most sense to me but happy to hear others opinions on that.

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -191,6 +191,7 @@ jQuery.entwine('ss', ($) => {
...
+      const fileSelected = attrs.ID > 0;
```
maxime-rainville, line 194: Do we know for a fact that ID will always be present and will always be cast a number?

I think we should be a bit more picky here and have some type guards.

code/Forms/FileFormFactory.php
```diff
@@ -457,14 +458,18 @@ protected function getPopoverActions($record)
 
     /**
      * @param File $record
+     * @param bool $fileSelected
      * @return FormAction
      */
-    protected function getInsertAction($record)
```
maxime-rainville, line 462: That's an API change. If you're calling this method without the second parameter or if you've overwriten the method, you'll get an error.

emteknetnz: Updated, moved logic to parent function that calls `getInsertAction()` instead, it's the only place that calls this function.

Does meant that overriding the button text would be possibly annoying, though you can still do that via `_t()`

Other options considered:
- make it an optional argument, release as `1.7` (longer to release bugfix)
- use class field `private $fileSelected = false` (kind of ugly)

client/src/entwine/TinyMCE_ssmedia.js
emteknetnz: changed to `const fileSelected = attrs.hasOwnProperty('ID') && attrs.ID !== null;`

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -626,7 +626,14 @@ class AssetAdmin extends Component {
    * @returns {object}
    */
   renderEditor() {
-    const { sectionConfig: config, viewAction, type, fileId, dialog, requireLinkText } = this.props;
+    const { sectionConfig: config,
```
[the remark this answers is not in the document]
emteknetnz: Updated

Issue #1025, opened 2019-10-20
Recently uploaded files appear in parent folders


brynwhyman:
## Overview
There appears to be a bug where after uploading a file and subsequently navigating through different folders, the recently uploaded file will appear in each folder you navigate to.

If you're uploading a file to a specific folder, you want to be sure that it's been saved in the correct place. This bug is confusing and doesn't instil confidence that the file has been saved correctly.

## Steps to reproduce
1. Navigate to Files area
2. Create a new folder and navigate to it
3. Upload a new file
4. Identify that the file is displayed in the window
5. Bug: Navigate to the parent folder and see that the file appears here too. Navigate to other folders and see that the file persists in view.

### Expected result
5. Navigate to the parent folder (or any other) and you should not see the file that was recently uploaded.

## Acceptance criteria
* The issue is covered by some behat test
* The solution does not significantly degrade performance

## Versions affected
`silverstripe/asset-admin 1.4.x-dev 35f64e5`
`silverstripe/assets 1.4.x-dev dc8bebb`

# Pull request
* [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/1118

bergice:
This is still an issue on `1.4.x-dev` and also `4.5`.

Additionally, the newly created folder will disappear.

jules0x:
If you click on the file, in the place that you're not expecting to see it... it navigates your back to the place it was actually uploaded.

rafaeldsousa:
I can reproduce issues above mentioned in `1.4.4` SS `4.4.4`.

PR #1118, opened 2020-07-21
FIX Prevent uploaded files showing in parent folder

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/1025


emteknetnz:
@maxime-rainville Good spot.

I've updated the javascript so that the upload progress bar shows again.

I did try to test the upload progress bar in behat with the following line after adding a file to the dropzone

```
Then I should see a ".gallery-item__upload-progress" element
```

It worked locally, though not on travis.  I suspect there's timing issues involved.  I think the only way to solve this reliably would be to upload a large file say 20MB, though this issue is not big enough to warrant increase the filesize of the git repo IMO

emteknetnz:
@maxime-rainville I've slightly redone this solution in a clearner fashion by adding a `uploadedToFolderId` property when a file is added to the dropzone, which includes either drag and drop or via the upload button

I've added a bunch of jest tests as requested

PR #1121, opened 2020-07-24
BUG Make the file upload progressbar smooth

maxime-rainville:
I was looking at the progress bar that gets displayed when files are uploaded. Right now it's quite choppy. The files is uploaded in chunk and as each chunk gets uploaded, the progress bar jump directly to the next step ... which looks weird and broken.

I've added a simple CSS transition to the progress bar which looks nicer and creates the "illusion" of speed ... as opposed to actual speed I guess.

[Look at this before and after you tube clip](https://youtu.be/OT4g2AtavjU)

Here's where the progress bar was updated:
- Thumbnail view
- Table view
- Preview for replace file
- Upload field

^ @silverstripeux


client/src/containers/TableView/TableView.js
```diff
@@ -111,6 +111,7 @@ class TableView extends Component {
       gridClassName: 'gallery__main-view--table',
       rowMetadata: {
         bodyCssClassName: this.getRowMetadata,
+        key: 'key'
```
maxime-rainville, line 114: Griddle is use to generate the Table file view. Currently Griddle create its own internal key for each row because we don't give it one it can use. This create the problem that each row gets re-rendered every time because a new internal ID is generated on each render. This prevented the CSS transition from working because each row was rendered from scratch each time the upload progress was updated. 

By telling Griddle to use our own ID, we allow React to keep track of row and only re-render the bits we need which allow the CSS transition to kick in. It should also have some nice front-end performance side effect.

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -119,7 +119,7 @@ class ThumbnailView extends Component {
     const badge = badges.find((badgeItem) => badgeItem.id === item.id);
     let props = {
       sectionConfig,
-      key: `${item.id}__${item.queuedId}`,
```
maxime-rainville, line 122: This way of build the ID was kind of dumb. `queuedId` is undefined for pre-existing file and all upload have an ID of 0.

sachajudd:
So much smoother and cleaner! Awesome @maxime-rainville! The only thing I'm wondering about is the colour of the loading bar. It's currently using $brand-primary which is cyan. We could probably phase out this colour and use $brand-secondary or $mid-blue (badge-info uses this) instead. @clarkepaul let me know your thoughts.

Looks like we have some double up on colours in our variable sheets which I'm noting here for a future clean up:

_bootstrap-variables.scss
$blue:     #005a93; // custom
$mid-blue: #0071c4; // custom
$cyan:     #29abe2

_variables.scss
$brand-secondary: #005a93;
$brand-primary: #29abe2;

maxime-rainville:
On a side, there's a [Bootstrap progress bar UI](https://getbootstrap.com/docs/4.0/components/progress/) which arguably we should be using for this.

Here's the [matching React strap component](https://reactstrap.github.io/components/progress/).

clarkepaul:
Yeah that's a nice increment on what was there. I think it's worthy to merge in as a Bootstrap swap out could take ages.

client/src/containers/TableView/TableView.scss
```diff
@@ -198,6 +198,7 @@ $gallery-list-thumb-height: 68px;
 .gallery__progress-bar-progress {
   height: 100%;
   background: $brand-primary;
+  transition: width 1s ease;
```
Cheddam, line 201:
```suggestion
  transition: width $transition-speed-slow ease;
```

client/src/components/GalleryItem/GalleryItem.scss
```diff
@@ -338,6 +338,7 @@ $gallery-item-label-height: 40px;
 .gallery-item__progress-bar {
   height: 100%;
   background: $brand-primary;
+  transition: width .5s ease;
```
Cheddam, line 341:
```suggestion
  transition: width $transition-speed-mid ease;
```

client/src/components/PreviewImageField/PreviewImageField.scss
```diff
@@ -68,6 +68,7 @@
 .preview-image-field__progress-bar {
   height: 100%;
   background: $brand-primary;
+  transition: width .5s ease;
```
Cheddam, line 71:
```suggestion
  transition: width $transition-speed-mid ease;
```

client/src/components/UploadField/UploadFieldItem.scss
```diff
@@ -139,6 +139,7 @@ $uploadfield-item-label-height: 40px;
 .uploadfield-item__progress-bar {
   height: 100%;
   background: $brand-primary;
+  transition: width .5s ease;
```
Cheddam, line 142:
```suggestion
  transition: width $transition-speed-mid ease;
```

maxime-rainville:
@Cheddam Good suggestion. Didn't know we had variables for that. FYI the TableView transition was set to 1sec by accident. I've updated all of the transitions to be `$transition-speed-mid`

@silverstripeux Have we reach some conclusion on the colour to use. Are we happy with `$brand-primary`?

sachajudd:
You can use $mid-blue from _bootstrap-variables @maxime-rainville

maxime-rainville:
Tweaked the colour to "mid-blue" as requested.

image

sachajudd:
Spicy 🌶️

Issue #879, opened 2018-11-12
Replace file "save to continue" message has incorrect padding


robbieaverill:
Closely related to #878 

**Version: SS 4.3.0-rc1** (in CWP 2.2.0-rc1 context)

* Upload a file
* Select that file
* Choose "replace" from the more actions menu
* Select a new file

Observe incorrect top padding on the message in the preview window:
![image](https://user-images.githubusercontent.com/5170590/48345090-b1061100-e67f-11e8-84fe-91176d7eb1a6.png)

sachajudd:
**Note:** Looks like this is only happening if there is no thumbnail present. Works correctly if there is a thumbnail for an image or a default file/video/pdf thumbnail.

robbieaverill:
Yeah good point. Reproducible for me by uploading a PDF, opening its details, then replacing it with another PDF.

![image](https://user-images.githubusercontent.com/5170590/82848258-9ed97580-9ea7-11ea-9f6d-ed8d8aa7e97b.png)

robbieaverill:
Related (but not the same): https://github.com/silverstripe/silverstripe-asset-admin/issues/878

emteknetnz:
Linked PR was merged

Issue #775, opened 2018-04-24
Preview broken when replacing file with a text file


robbieaverill:
**SS 4.1.x-dev**

* Go to asset admin
* Upload a file
* Select the file
* Replace the file with a text file

Expected: document image shows for preview (like when you open it initially)

Actual:

* broken image shows "preview"
* the buttons are jammed together to fit into the alert

![image](https://user-images.githubusercontent.com/5170590/39162069-ce2d6708-47c7-11e8-8084-d1b593f154f7.png)

### Broken image

DOM changes required:

```diff
- <img alt="preview" src="framework/client/dist/images/app_icons/generic_92.png" class="editor__thumbnail">
+ <img alt="preview" src="resources/vendor/silverstripe/framework/client/images/app_icons/generic_92.png" class="editor__thumbnail">
```

Doesn't contain the correct resources location.

When you insert the correct image path manually it looks like this:

![image](https://user-images.githubusercontent.com/5170590/39162272-d60cd4bc-47c8-11e8-9f61-30cff37a1bee.png)

robbieaverill:
This has changed since initially raised. It now looks like this:

![image](https://user-images.githubusercontent.com/5170590/82848471-6f773880-9ea8-11ea-9552-69127e3078fa.png)

The top padding bug is tracked here: https://github.com/silverstripe/silverstripe-asset-admin/issues/879

Closing.

sachajudd:
I'm still seeing this when replacing a document with another document 
![image](https://user-images.githubusercontent.com/24258161/89135619-df79fd00-d582-11ea-9981-1aece62d7d0b.png)
You can see it just sticking out from under the success message, when removing the green bg it looks like this:
![image](https://user-images.githubusercontent.com/24258161/89135633-01737f80-d583-11ea-870d-fd3a652b9c08.png)

PR #1125, opened 2020-07-29
FIX Padding on non-image replace file message

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/879

After fix on IE11:

image


sachajudd:
Related: I've reopened https://github.com/silverstripe/silverstripe-asset-admin/issues/775 regarding the broken preview when replacing a file with certain extensions.

Issue #1044, opened 2020-01-14
Bulk files operation overlap on IE11


maxime-rainville:
When selecting files in asset-admin in IE11, the list of file operations is mangled. All the possible operation appear to overlap.

![image](https://user-images.githubusercontent.com/1168676/72311063-60a59b80-36e8-11ea-8381-438dfb2a98e5.png)

# Pull requests
* [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1127

maxime-rainville:
Confirmed on SS4.4 and 4.6.x.

GuySartorelli:
This also affects the "Insert" button when selecting an image from the "Choose existing" option from an upload field. It seems to be an issue with IE11's implementation of flex. I have found that changing
```css
.bulk-actions > * {
  flex: 0;
}
```
to
```css
.bulk-actions > * {
  flex: none;
}
```
resolves the issue, and I haven't yet seen any negative side effects on any browser.

maxime-rainville:
@GuySartorelli 's solution seems to work pretty well. I've got a PR for it. https://github.com/silverstripe/silverstripe-asset-admin/pull/1127

emteknetnz:
The linked PR has been merged and merged-up, closing now

PR #1127, opened 2020-08-03
BUG Correctly display bulk actions in IE11

maxime-rainville:
Removing this CSS rules prevents bulk actions from overlapping in IE11 and doesn't seem to have any negative affects in any other browser.

Parent issue
* https://github.com/silverstripe/silverstripe-asset-admin/issues/1044


emteknetnz:
Tested in IE11, Edge, Firefox, Chrome

PR #1128, opened 2020-08-06
Graphql 4 compatability

unclecheese:


chillu:
Starting a todo list to keep track of what's left :)

* [ ] Fix `DataList` strings in `readFiles` response (e.g. `category: "SilverStripe\ORM\DataList"`)
* [ ] Change field names to upper camelcase. Or do we consider this an API breakage?

_config/graphql.yml
```diff
@@ -1,31 +1,13 @@
 ---
 Name: assetadmingraphql
 ---
-SilverStripe\GraphQL\Manager:
```
sminnee, line 4: Should the old schema config be retained if people are continuing to use graphql 3?

client/src/state/files/buildPublicationMutation.js
```diff
@@ -5,17 +5,17 @@ const buildPublicationMutation = (mutationName) => {
   const operationName = mutationName.charAt(0).toUpperCase() + mutationName.slice(1);
   const mutation = gql`
   mutation ${operationName}($IDs:[ID]!, $Force:Boolean, $Quiet:Boolean) {
-    ${mutationName}(IDs: $IDs, Force: $Force, Quiet: $Quiet) {
```
sminnee, line 8: Whyyyyyyyyyyyyyyyyy....!??!??!?!!

Why create busy-work for people?

sminnee: More specifically, I question whether the benefit of changing case is greater than the cost of upgrading all the uses. As well as requiring new configuration in the module, it also breaks the API exposed.

sminnee: So anyone who uses graphql, considering an upgrade to v4, needs to consider not only the effort of upgrading their site, but they also need to get all of the consumers of their graphql API to change their code as well.

unclecheese: I'm not sure that's the case. A few points:

1) This endpoint is only exposed to `admin/graphql`. AFAIK this is not something a developer would be plugging into in their app layer. I suppose it's possible, but if your app is using `admin/graphql` and not the generic `/graphql` endpoint that's provided out of the box, you probably have bigger issues than case changes. There's a 99% chance this creates busy work for exactly zero people.

2) There are actually tangible gains to be made through the case changes. Previously, the implementation of features like filtering, sorting, and pagination were done with bespoke code in asset-admin. Those are now all core APIs of graphql that can be used like decorators on any query. They do, however, like the rest of the graphql module, assume lowerCamelCase field names unless you specify otherwise. While it adds some work on our part to update the queries (or maintain two iterations), it also pays off a massive technical debt that was holding this module back since its 1.0 release. We can now do more asset-admin in this module and less graphql due to more reusable code and standardisation.

3) The high level goal is to make Silverstripe a competitive solution for decoupled/headless setups, and the UpperCamelCase standard is an anti-pattern in graphql APIs, particularly because they often map to javascript properties, which are also lowerCamelCase, and this is just kind of a bad look.

So yes, there's a cost, mostly paid by us, *maybe* paid by the odd thirdparty dev, but more than enough gains to justify it IMO.

sminnee: Yeah I think @chillu backs you up on this, essentially because GraphQL 1-3 should have really been called 0.1 - 0.3 and this new major release is the first "actually" stable one, and the uptake of custom GraphQL set-ups is still very limited.

That being the case we'd want to avoid doing changes of this nature in a silverstripe/graphql 5, though.

unclecheese: Have migrated the old query to `_legacy/buildPublicationMutation`

_config/graphql.yml
chillu: Seems to be done through `_config/graphql-legacy.yml` now

unclecheese:
Closing as PR is moving to a local branch.

unclecheese:
Moved to: https://github.com/silverstripe/silverstripe-asset-admin/pull/1154

Issue #1109, opened 2020-07-02
Interacting with breadcrumb reverts list view display to grid


maxime-rainville:
# Steps to reproduce
- Navigate to a subfolder
- Switch to the list view
- Use the breadcrumb to navigate to a different folder

Expected results: You get taken to that folder and files are still displayed in the list view
Actual results: You get taken to that folder, but files are displayed in a grid

https://youtu.be/bBn3yFZfCUY

# Notes
The back button work fine.

# PR's

- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1129

emteknetnz:
Linked pull request has been merged

PR #1129, opened 2020-08-06
FIX breadcrumbs not using `view=list` when clicked

bergice:
Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/1109


emteknetnz:
@bergice broken behat test.  Also need to rebuild dist file - the failed npm test I think means that you had the wrong version of admin installed when you built asset-admin

bergice:
@emteknetnz I've rebuilt the dist. It's passing now.

emteknetnz:
Good job

Issue #1108, opened 2020-07-02
Folders appear on second page when sorting list view


maxime-rainville:
# Steps to reproduce
* Navigate to a folder with sub folders and 50+ files
* Switch to list view
* Notice that folders appear at the top of the list
* Sort the files by a different order

Expected results: Folders displayed first
Actual results: Folders are not displayed first. However if you navigate to the second page they appear at the top.

https://youtu.be/GY-fLxTCGBM

# Notes 
The grid view works as expected and if you have less than 50 entries the folder appear as expected in the list view

### PRs
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1222

emteknetnz:
I had 2 folders where I changed their names after uploading several dozen files, this meant the folders LastEdited was greater than all the filters

Use the query `/admin/assets/?view=table&sort=lastEdited%2Casc&page=2`
- sort by LastEdited ASC
- page 2

The graphql query on page 2 returned 20 child nodes for the files, of which 19 and 20 were the folders

So it seems the issue is that the graphql query is not returning the folders 'in front' of all the regular files, instead it's treating folders like files, and then presumably react is what's putting the folders at the top of the results

emteknetnz:
Possibly we could do this in the graphql query, though it seems as though we cannot specify multiple sortBy's + multiple directions?

https://github.com/silverstripe/silverstripe-asset-admin/blob/1/client/src/state/files/readFilesQuery.js#L44
```js
        sortBy: (sortField && sortDir)
          ? [{ field: sortField, direction: sortDir.toUpperCase() }]
          : undefined,
      },
```

https://github.com/silverstripe/silverstripe-asset-admin/blob/1/client/src/state/files/readFilesQuery.js#L93
```js
const query = {
  apolloConfig,
  templateName: READ,
  pluralName: 'Files',
  pagination: false,
  params: {
    limit: 'Int!',
    offset: 'Int!',
    rootFilter: 'FileFilterInput',
    childrenFilter: 'FileFilterInput',
    sortBy: '[ChildrenSortInputType]',
  },
```

michalkleiner:
Mind that there could be other subclasses of File, not just Folder, so that could mess up things more if sorting _just_ by the class, _just_ alphabetically.

emteknetnz:
Yeah I don't know how else we'd do this without using ClassName though, we're restricted by this 

`Does not support in-memory sorting for composite values (getters).`

[...]

PR #1130, opened 2020-08-12
FIX Show Folders before Files in paginated results

emteknetnz:
Fix for https://github.com/silverstripe/silverstripe-asset-admin/issues/1108

Some discussion on why it's done on the database column `ClassName` instead of the derived value `type` https://github.com/silverstripe/silverstripe-asset-admin/issues/1108#issuecomment-672547010

**Update: Since `Image extends File`, i.e. "Image" is greater than "Folder" means that sorting by LastEdited, Title, etc, is now broken =(**

Is there anyway to get graphql to do this?
`ORDER BY (CASE WHEN ClassName = 'SilverStripe\\Assets\\Folder' THEN 0 ELSE 1 END), LastEdited ASC`


emteknetnz:
Superseded by https://github.com/silverstripe/silverstripe-asset-admin/pull/1134

PR #1133, opened 2020-08-14
WIP Allow generation of thumbnail for non-image file

maxime-rainville:
Companion PR for https://github.com/silverstripe/silverstripe-assets/pull/411


ChloeSartorelli:
This relies on a PR that did not get merged. It's likely we'll do something similar in the future, but we won't use this PR.

PR #1134, opened 2020-08-17
BUG Always sort files with folders first

maxime-rainville:
This is my go at fixing https://github.com/silverstripe/silverstripe-asset-admin/issues/1108

It's somewhat dodgy in that it relies on extending the native connection with our own ReadFileConnection so we can give it a different applySort method.

### DIFF
--- a/client/src/state/gallery/GalleryReducer.js
+++ b/client/src/state/gallery/GalleryReducer.js
@@ -18,7 +18,7 @@ const initialState = {
    */
   sorters: [
     {
-      field: 'title',
+      field: 'name',
       direction: 'asc',
       label: i18n._t('AssetAdmin.FILTER_TITLE_ASC', 'title a-z'),
     },
--- a/code/GraphQL/ReadFileConnection.php
+++ b/code/GraphQL/ReadFileConnection.php
@@ -1,0 +1,50 @@
+<?php
+namespace SilverStripe\AssetAdmin\GraphQL;
+
+use SilverStripe\Assets\Folder;
+use SilverStripe\GraphQL\Pagination\Connection;
+use SilverStripe\ORM\DataQuery;
+use SilverStripe\ORM\DB;
+
+class ReadFileConnection extends Connection
+{
+
+    protected function applySort($list, $sortBy)
+    {
+        $list = parent::applySort($list, $sortBy);
+
+
+        $list = $list->alterDataQuery(static function (DataQuery $dataQuery) {
+            $query = $dataQuery->query();
+            $existingOrderBys = [];
+            foreach ($query->getOrderBy() as $field => $direction) {
+                if (strpos($field, '.') === false) {
+                    // some fields may be surrogates added by extending augmentSQL
+                    // we have to preserve those expressions rather than auto-generated names
+                    // that SQLSelect::addOrderBy leaves for them (e.g. _SortColumn0)
+                    $field = $query->expressionForField(trim($field, '"')) ?: $field;
+                }
+
+                $existingOrderBys[$field] = $direction;
+            }
+
+            // Folders should always go first
+            $dataQuery->sort(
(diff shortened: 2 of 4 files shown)


client/src/state/gallery/GalleryReducer.js
```diff
@@ -18,7 +18,7 @@ const initialState = {
    */
   sorters: [
     {
-      field: 'title',
+      field: 'name',
```
maxime-rainville, line 21: This part arguably doesn't have anything to do with with the main fix. It's just that while working on the main issue I realised that we sort on "name" by default. However the when you explicitly set the sort order, we switch to sorting by "title".

code/GraphQL/ReadFileQueryCreator.php
```diff
@@ -23,7 +23,7 @@ public function attributes()
 
     public function createConnection()
     {
-        return Connection::create('readFiles')
+        return ReadFileConnection::create('readFiles')
```
emteknetnz, line 26: So, if I understand this correctly, what we're doing is essentially making the default sort of any reading files query
1. Folder
2. Not-folder

It does seems kind of wrong given that graphql is supposed to be sort of abstract?   The "correct" way to do this would be a query something along the lines of

`files(orderBy: {field: NON_EXISTENT_DB_COLUMN, direction: DESC})`

Don't get me wrong, I think your solution is great and very pragmatic given the context, this is the asset-admin module after all.

code/GraphQL/ReadFileConnection.php
```diff
@@ -0,0 +1,50 @@
...
+                }
+
+                $existingOrderBys[$field] = $direction;
+            }
+
+            // Folders should always go first
+            $dataQuery->sort(
+                sprintf(
+                    '(CASE WHEN "ClassName"=%s THEN 1 ELSE 0 END)',
+                    DB::get_conn()->quoteString(Folder::class)
```
emteknetnz, line 35: In theory this should be Folder + subclasses of Folder, though that's probably a real edge case there so probably not worth putting in

code/GraphQL/ReadFileQueryCreator.php
maxime-rainville: Yeah, the problem is the current GraphQL API doesn't give us the flexibility to write really weird sort condition. In this context, creating our own asset-admin connection to implement our really weird sort condition sounds like a reasonable compromise.

code/GraphQL/ReadFileConnection.php
maxime-rainville: Fine with changing it. FYI that bit of logic got copied over from https://github.com/silverstripe/silverstripe-asset-admin/blob/a35d1a3cb7b02d488eff3f36613de0ef004478a0/code/GraphQL/FolderTypeCreator.php#L171-L204

We can probably remove it from `FolderTypeCreator` if we're going to code it directly in the connection. There's a comment in `FolderTypeCreator` that seems to imply that we don't always want the folder first, which sounds kind silly to me.

Might be worth validating with PUX if there's a scenario where we don't want the Folders always first.

code/GraphQL/ReadFileQueryCreator.php
emteknetnz: Will this limitation change with graphql4 ?

maxime-rainville: Yeah ... don't think this will work in graphql v4. We might be able to write a custom sort plugin for it. 

Maybe a simpler fix could be to update the UI so that it always sends along a 'sort by classname' instruction before any other sorting order. That would make some sense since this is mostly a UI concern.

emteknetnz:
Closed in favour of https://github.com/silverstripe/silverstripe-asset-admin/pull/1222

Issue #915, opened 2019-02-15
Error message from server contains extraneous markup 


gmlnchv:
### Overview

Error messages that come straight from the server are not readable to the CMS user.

One example is PDF files exceeding the specified maximum file size (for the server) seem to have some extraneous markup surrounding the error message.

![screen shot 2019-02-15 at 2 37 57 pm](https://user-images.githubusercontent.com/1403779/52828893-85d40780-312f-11e9-990e-8641c4b65fa0.png)

### Recreation steps
1. Use a project hosted on Silverstripe Cloud (e.g. demo.silverstripe.org)
2. Upload a 13mb image to asset-admin 
3. Observe error

### Acceptance criteria
- Where the error message/ response comes straight from the server, exclude any HTML tags, ensuring the messaging is readable
- Look at the common status codes and create a specific error message, or default to a generic error message
- If any new translation strings are introduced, ensure they're pushed to Transifex

### Notes
- [Similar work has recently been done in the silverstripe/mfa module ](https://github.com/silverstripe/silverstripe-mfa/pull/401/files)

robbieaverill:
Oh, that's probably coming from your web server (apache/nginx)! Is this reproducible on a vanilla installation by posting an enormous file at it? Could you please share some information about the versions you're running as well as any custom config, web server config etc? If you're running nginx especially

gmlnchv:
@robbieaverill you are right - I could only reproduce this on SilverStripe Platform. Fresh SS4 installation doesn't seem to have this :) Should probably close and raise with the Platform instead?

robbieaverill:
@gmlnchv I think you can leave the issue open, but maybe raise a helpdesk issue and reference this GitHub issue in it

chillu:
The frontend logic should be looking at the status code *as well as* the content type of the response. If it's `text/html`, then don't use it. If it's `application/json`, then try to parse it for error messages. If it's `text/plain`, display it. In all cases, the output should be sanitised.

gmlnchv:
> @gmlnchv I think you can leave the issue open, but maybe raise a helpdesk issue and reference this GitHub issue in it

@robbieaverill Still worthwhile raising with the helpdesk or can be fixed here you think?

robbieaverill:
@gmlnchv I would suggest raising a ticket and reference this issue

robbieaverill:
[...]

Issue #895, opened 2018-12-13
Error message when adding large files are not readable


robbieaverill:
Originally from https://github.com/silverstripe/silverstripe-elemental-fileblock/issues/9

### Reproduce

1. Install `silverstripe/elemental-fileblock` on `2.x-dev` and  `dnadesign/silverstripe-elemental` on        `4.x-dev`
2. Create a page that has the ElementalPageExtension enabled
3. Add a FileBlock via the Add Block button
4. Click on the newly created File Block to expand it
5. Use the `Browse` link to add a file that is 2MB+
6. Observe an truncated error message as shown below

![image](https://user-images.githubusercontent.com/14869519/47880099-340cb780-de87-11e8-9669-5add90850b0e.png)

7. Hover over the error message and see that the tooltip does not show the entire message either

clarkepaul:
I'm getting the same issue but also the error message isn't readable either. SS4.3 using large image.
![image](https://user-images.githubusercontent.com/555033/53048520-cfd73780-34f9-11e9-86e5-e4d9e057b9ee.png)

ScopeyNZ:
Yeah it seems to just dump the response of the validation request verbatim. In your case it's obviously receiving some HTML page that it's dumping the source of 😞 .

kinglozzer:
I suspect that https://github.com/silverstripe/silverstripe-asset-admin/issues/925 is the reason that @clarkepaul is seeing HTML instead of a friendly error message

robbieaverill:
Note that the escaped HTML part of this is tracked here: https://github.com/silverstripe/silverstripe-asset-admin/issues/915

This issue should just address the fact that a message like "Filesize is too large" isn't able to be rendered. I believe someone made a PR that would show the full message as a title/hover tooltip, which is an improvement and might be enough to close this issue.

brynwhyman:
Noting that this issue has been updated to include some new ACs. @silverstripeux noted that they want to have another look at the designs to check everything is in order

sachajudd:
Updated designs can be found here: [Silverstripe CMS Design System](https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/asset/componentContainers/5f5fffe36493f7b97c262d08/tab/design?mode=preview)

Few notes:
* Updated the padding on the basics component which should apply to all uploadfields
* Added an additional extended error uploadfield that grows in width depending on how long the error message is. This is to cater for longer messages or possible translations.
[...]

PR #1138, opened 2020-09-15
FIX Strip html tags from file upload error responses

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/915

**Notes**

- The max-width 50% possibly solves https://github.com/silverstripe/silverstripe-asset-admin/issues/895 (though it was unintentional)
- The unused messages for "file cannot be found" and "extension is not allowed" is for the [DSM](https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/asset/componentContainers/5f5fffe36493f7b97c262d08/tab/design?mode=preview) which does contain messages for "file size limit exceeded" and "something went wrong, please try again"

**Update transifix**
- [ ] Have pushed new translations to transifix (en.json has been updated in this PR, but not en.js)

**New behaviour:**

File Manager Gallery Server Error 413

image

File Manager Gallery Server Error - Non-413

image

Upload Field Error - Server Error 413

image

Upload Field Error - Server Error Non-413
image

Non Server Error file size too big messaging remains as is

image

image

Error can be simulated in File manager gallery by updating AssetAdmin.php

```php
    public function apiCreateFile(HTTPRequest $request)
    {
        $this->httpError(413);
```

Error can be simulated in upload field by updating UploadField.php

```php
    public function upload(HTTPRequest $request)
    {
        $this->httpError(413);
``` 

Add an upload field to Page.php

```php
    class Page extends SiteTree
    {
        public function getCMSFields()
        {
            $uploadField = new UploadField('ABC');
            $fields->insertBefore('Title', $uploadField);
            return $fields;
        }
```


Cheddam (changes requested):
@emteknetnz The goal was to produce readable error messages for users, which I'd argue simply stripping the HTML tags does not achieve. With the example in the original issue, an author would now see:

> 413 Request Entity Too Large

The author _might_ infer that this means their file was too big, but it's not exactly a clear description, and other errors may not fair as well. I think you'd be better off creating a map of HTTP error codes to author-readable messages and using those if the response is a string. If developers want to push through a custom error message, they need to make sure they return a valid JSON response with the `errors` key populated.

cc/ @silverstripe-ux - maybe you can collaborate on the error messaging?

sachajudd:
Linking to https://github.com/silverstripe/silverstripe-asset-admin/issues/895#issuecomment-692438512 as I've just added some designs with four different error messages that we'd like to add in an instance that an extension is not allowed or a file is too big etc. We can come up with a few more if need be but these should allow you to cover the most common errors.

client/src/lib/getStatusCodeMessage.js
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Convert an http status code to a message
+ *
+ * @param { int } statusCode
+ */
+export default function getStatusCodeMessage(statusCode) {
+  if (statusCode === 413) {
+    return i18n._t('AssetAdmin.ERROR_FILE_SIZE', 'File size limit exceeded');
+  }
```
Cheddam, line 11: I realise this will have no immediately impact besides an infinitesimal drop in performance, but for the sake of making the next developer's job easier, could you turn this into a Code => Message map?

Cheddam (commented):
Looking good, one small recommended change.

client/src/lib/getStatusCodeMessage.js
emteknetnz: I'm sure the next developer can refactor that themselves =)

Cheddam (approved):

dhensby:
Nice work!

PR #1139, opened 2020-09-30
ENH Exclude FileLink's using a different extension hook

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1098

Blocked by https://github.com/silverstripe/silverstripe-admin/pull/1139

### DIFF
--- a/code/Extensions/UsedOnTableExtension.php
+++ b/code/Extensions/UsedOnTableExtension.php
@@ -2,18 +2,38 @@
 
 namespace SilverStripe\AssetAdmin\Extensions;
 
+use SilverStripe\Assets\File;
 use SilverStripe\Assets\Shortcodes\FileLink;
 use SilverStripe\Core\Extension;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataObject;
+use SilverStripe\Security\Member;
 
 /**
- * Hides File Links on the "Used On" tab when viewing files
+ * Hides several types of DataObjects on the "Used On" tab when viewing files
  */
 class UsedOnTableExtension extends Extension
 {
+    /**
+     * @var array $excludedClasses
+     */
+    public function updateUsageExcludedClasses(array &$excludedClasses)
+    {
+        $excludedClasses[] = FileLink::class;
+        $excludedClasses[] = File::class;
+        $excludedClasses[] = Member::class;
+    }
+
+    /**
+     * Legacy function kept for semver, replaced with updateUsageExcludedClasses above
+     *
+     * @var ArrayList $usage
+     * @var DataObject $record
+     * @see UsedOnTable::updateUsage
+     * @deprecated 4.7.0 Use self::updateUsageExcludedClasses instead
+     */
     public function updateUsage(ArrayList &$usage, DataObject &$record)
     {
-        $usage = $usage->exclude('ClassName', FileLink::class);
+        // noop
     }
 }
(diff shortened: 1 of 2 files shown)


tests/php/Forms/UsedOnTableTest.php
```diff
@@ -27,11 +27,11 @@ public function testExcludes()
 
         $expected = json_encode([
             "usage"=> [[
-                "id" => 0,
+                "id" => 1,
                 "title" => "My Page",
                 "type" => "Page",
-                "state" => "Draft",
-                "link" => sprintf('%s://localhost/admin/pages/edit/show/1', $protocol)
+                "link" => sprintf('%s://localhost/admin/pages/edit/show/1', $protocol),
```
maxime-rainville, line 33: Can we avoid explicitly referencing localhost here? Otherwise this test will only pass when your test environment is on localhost.

code/Extensions/UsedOnTableExtension.php
```diff
@@ -2,18 +2,35 @@
...
 
+use SilverStripe\Assets\File;
 use SilverStripe\Assets\Shortcodes\FileLink;
 use SilverStripe\Core\Extension;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataObject;
+use SilverStripe\Security\Member;
 
-/**
- * Hides File Links on the "Used On" tab when viewing files
```
maxime-rainville, line 11: Why remove the comment?

maxime-rainville (changes requested):

tests/php/Forms/UsedOnTableTest.php
emteknetnz: I don't know.   'localhost' was part of an existing test that was works fine as is.  I don't think spending time here is worthwhile.

code/Forms/FileFormFactory.php
```diff
@@ -118,6 +118,7 @@ protected function getFormFieldUsageTab($record, $context = [])
             _t(__CLASS__ . '.USAGE', 'Used on'),
             $usedOnField
         );
+        $tab->addExtraClass('tab-used-on-table');
```
emteknetnz, line 121: This should be removed, it was for the old setInterval method of tab detection.

code/Extensions/UsedOnTableExtension.php
emteknetnz: Added back a comment

tests/php/Forms/UsedOnTableTest.php
maxime-rainville: This fix doesn't take a lot of time to apply and it's going to save people like me who run their test suite under a different BASE_URL a lot of aggravation.

```suggestion
                "link" => Director::absoluteURL('admin/pages/edit/show/1'),
```

maxime-rainville (changes requested):
Can we update `ReadFileUsageQueryCreator` so it uses the new `DataObject::findAllRelatedData()` method instead?

tests/php/Forms/UsedOnTableTest.php
emteknetnz: Updated

code/Forms/FileFormFactory.php
emteknetnz: Updated

emteknetnz:
>> Can we update ReadFileUsageQueryCreator so it uses the new DataObject::findAllRelatedData() method instead?

Is that for the count for when you're about to delete a file?  That's been split to its own card https://github.com/silverstripe/silverstripe-asset-admin/issues/1146

maxime-rainville (approved):

maxime-rainville:
Dependent PR has been merged. Just pushed a fake commit to force a new build.

maxime-rainville:
Build is not green, but we need this merge for the beta. We're confident we can get it green post beta.

PR #1143, opened 2020-10-07
MNT Fix insert image behat test

emteknetnz:


emteknetnz:
Wrong approach, should fix code rather than test

PR #1140, opened 2020-10-06
MNT PHP8 support

dnsl48:
A part of [PHP8 support initiative](https://github.com/silverstripe/silverstripe-framework/issues/9704)

PR #1144, opened 2020-10-08
Fix behat tests

dnsl48:
Fixes broken behat tests.
Requires https://github.com/silverstripe/silverstripe-asset-admin/pull/1140 to be merged first.

### DIFF
--- a/tests/behat/features/insert-an-image.feature
+++ b/tests/behat/features/insert-an-image.feature
@@ -19,7 +19,8 @@ Feature: Insert an image into a page
     Then I should see the "Form_fileInsertForm" form
       And I should not see an ".gallery-item--selectable" element
       And I should not see an ".bulk-actions" element
-    When I press the "Insert" button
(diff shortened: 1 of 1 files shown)


emteknetnz:
IDK.  I went down this exact route here https://github.com/silverstripe/silverstripe-asset-admin/pull/1143 before self-closing it because it's fixing tests rather than code

I cannot replicate the travis behaviour locally, so I get why we'd want to take this approach

Did you manage to get any insight into why this behaviour is happening in travis after the javascript dependency changes?

dnsl48:
> Did you manage to get any insight into why this behaviour is happening in travis after the javascript dependency changes?

Not sure what dependency changes you mean, but I've been able to reproduce that issue locally by following the travis installation process.

dnsl48:
resolved by https://github.com/silverstripe/silverstripe-framework/pull/9732

Issue #896, opened 2018-12-19
Alignment of insert media toolbar actions


newleeland:
![192_168_3_157_8888_demo-template_admin_assets_-2](https://user-images.githubusercontent.com/8085295/50190607-c3661f80-038e-11e9-93ab-779e0b0e06a8.png)

Small CSS alignment issues and update sort button to be more modern and mobile friendly 
(icon: update to this icon)
![27721211-083cc1e6-5db3-11e7-9a8c-e93c12a079ba](https://user-images.githubusercontent.com/8085295/50190650-f4465480-038e-11e9-83b6-3f06b2b83605.png)

Although on its own its very low impact, it is set to Impact medium, as over a long term small tasks like this, easy build up and start deteriorating the experience and can damage overall perception of product if UX debt is not managed.

newleeland:
![](https://github.trello.services/images/mini-trello-icon.png) [(2) Review insert file UI, raise any existing issues not captured yet](https://trello.com/c/XeRRjd9i/365-2-review-insert-file-ui-raise-any-existing-issues-not-captured-yet)

emteknetnz:
If we align the buttons then there's an issue with the hover state being "cut off"

![image](https://user-images.githubusercontent.com/4809037/95727862-43cde100-0cd7-11eb-9eb8-2e46cd4f6b12.png)

![image](https://user-images.githubusercontent.com/4809037/95727951-5fd18280-0cd7-11eb-8f7a-cd50bd24a936.png)

We could still align it and have the hover state will "go over", though this may look worse

@sachajudd how would you like to proceed with this?

emteknetnz:
Enhanced in https://github.com/silverstripe/silverstripe-asset-admin/pull/1145

PR #1145, opened 2020-10-12
FIX Button alignment tweaks

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/896


emteknetnz:
If we align the buttons then there's an issue with the hover state being "cut off"

image

image

We could still align it and have the hover state will "go over", though this may look worse

image

However we actually already do this for the search button on the "Pages" section in the CMS

image

This is how it looks at present:

image

image

@sachajudd how would you like to proceed with this?

emteknetnz:
The new icon looks, which is already used on small breakpoints, does look a little odd to me - would you like to keep this @sachajudd?

image

sachajudd:
> The new icon looks, which is already used on small breakpoints, does look a little odd to me - would you like to keep this @sachajudd?
> 
> image

In Jared's design he is recommending completely dropping the text and replacing to use the icon. It's a well known icon, you don't need the text beside it. 

In your other hover issues you're having I'm unsure why your hover is being cut off and isn't moving with the whole toolbar? We will need to catch up so I can see it visually. Set a meeting in my calendar and you can show me a demo.

emteknetnz:
@sachajudd have updated the sort icon

image

sachajudd (approved):
I've approved this from a UX perspective. Someone will need to pull down and test the UI with the code though. Nice one Steve 🙂

bergice (changes requested):
Tests need to be fixed.

emteknetnz:
@bergice Tests have been fixed

Issue #1136, opened 2020-09-10
"(Edit) Details" button only visible if grid shows the page of selected file


sb-relaxt-at:
Given a folder with so many images that the don't fit on one page of the asset admin overlay and an UploadField.

1. Select an image from the second page in the UploadField and insert
2. Click on the eye symbol for the just added file in the UploadField in order to edit it
3. The overlay will show page one (without the prior selected image)
4. You won't be able to edit the file using the "Details" button as this is only visible if the page with the file is selected

I would expect the overlay to automatically show the relevant page with the selected image. And the "Details" button should be visible independent of the page selected. 

### PRs
- [ ] ~~https://github.com/silverstripe/silverstripe-asset-admin/pull/1221~~
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1225

PortableSteve:
I've spent the last day trying to figure out why the "Details" button was showing for some images and not others - I can confirm it seems to be based on whether the image is on the visible page in the image browser on the left. In fact, if you switch between pages, the "Details" button is displayed/removed as you move on/off the relevant page.

jules0x:
I'm also experiencing this. 
I am able to replicate as per sb-relaxt-at and PortableSteve comments above.
This is a high user impact, so I'm unable to upgrade clients to this version/recipes using this version 😢

Leapfrognz:
I have had to send an email to several clients highlighting this issue, so In my opinion this is a major.

brynwhyman:
FYI, our team are planning to get started on a fix for this in the next 2-3 weeks.

bergice:
The details button depends on the file being passed to the `EditorHeader` component. The file is not found when viewing a page without the selected file because the file is looked up from the list of files on the current page (see `AssetAdmin.findFile`).

The rest of the `Editor` component renders fine on a different page, but I think that's either because it's rendered using `FormBuilder`/`redux-form` or because we override the `createFn` for the `EditorHeader` component (see `Editor.createFn`).

maxime-rainville:
Just summarising the discussion we had on Slack.

Part of the problem is on this line: https://github.com/silverstripe/silverstripe-asset-admin/blob/f96d0f84a2d150297e9e570128566fefe2644e76/client/src/containers/Editor/Editor.js#L218-L223

The purpose of this snippet is to decide what action should happen when you click the "Details" button. If no action is provided, then the button doesn't show at all.

[...]

PR #1153, opened 2020-11-02
Fix details button not showing on different pages

bergice:
Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/1136


maxime-rainville (changes requested):
That's not going to work. There's some scenarios where `file` will is undefined and it will crash:
- In the modal navigate to a folder with 50+ children
- Start editing a file or a folder in the editor
- Navigate to a different page

https://youtu.be/q5cae8h4UmQ

maxime-rainville (changes requested):
That's still not going to work. The insert modal allows you to edit folders and it needs to know it's a folder to suppress the "Details" button. This PR introduces the following bug.

https://youtu.be/cr4KG93QgN0
- In the "Insert Image" modal, navigate to a directory with 50+ folders/files
- Edit a folder.
- Use the pagination to access more files/folder.

Expected result: the editor is still present allowing you to edit the folder and an the "Details" button doesn't show.
Actual result: The "Details" button start showing

PR #1154, opened 2020-11-02
NEW: GraphQL 4 compatibility 

unclecheese:
Migrated from: https://github.com/silverstripe/silverstripe-asset-admin/pull/1128


chillu (changes requested):

.travis.yml
```diff
@@ -76,7 +93,7 @@ before_script:
   - if [[ $NPM_TEST ]]; then rm -rf client/dist && nvm install && nvm use && npm install -g yarn && yarn install --network-concurrency 1 && (cd vendor/silverstripe/admin && yarn install --network-concurrency 1) && yarn run build; fi
 
 script:
-  - if [[ $PHPUNIT_TEST ]]; then vendor/bin/phpunit tests/php; fi
+  - if [[ $PHPUNIT_TEST ]]; then vendor/bin/phpunit tests/php '' flush=1; fi
```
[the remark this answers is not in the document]
unclecheese: It was the only thing that got the tests working at the time. (table names were exceeding 64 chars because `table_name` was not in the config manifest).

I'll remove and see if it's still a factor.

_graphql/mutations.yml
```diff
@@ -0,0 +1,39 @@
+'createFile(file: FileInput)':
+  type: FileInterface
+  description: Creates files and images
+'createFolder(folder: FolderInput)':
+  type: FileInterface
+  description: Creates files and images
+'deleteFiles(IDs: [ID]!)':
```
chillu, line 7: This should be lowercase `IDs`?

_graphql/queries.yml
```diff
@@ -0,0 +1,3 @@
+'readFileUsage(IDs: [ID]!)': '[FileUsage]'
```
chillu, line 1: This should be lowercase `IDs`?

_graphql/mutations.yml
```diff
@@ -0,0 +1,39 @@
...
+  type: '[ID]'
+  description:
+'moveFiles(folderId: ID!, fileIds: [ID])':
+  type: FileInterface
+publishFiles:
+  type: '[PublicationResult]'
+  description: Publishes a list of files
+  args:
+    IDs:
```
chillu, line 16: This should be lowercase `IDs`?

_graphql/mutations.yml
```diff
@@ -0,0 +1,39 @@
...
+      description: If true, proceed with the mutation, regardless of notices
+      defaultValue: false
+    quiet:
+      type: Boolean
+      description: If true, suppress warnings
+unpublishFiles:
+  type: '[PublicationResult]'
+  description: Unpublishes a list of files
+  args:
+    IDs:
```
chillu, line 29: This should be lowercase `IDs`?

code/GraphQL/Schema/Builder.php
```diff
@@ -0,0 +1,46 @@
...
+        }, File::config()->get('app_categories'));
+
+        // Sanitise GraphQL Enum aliases (some contain slashes)
+        foreach ($categoryValues as $key => $v) {
+            unset($categoryValues[$key]);
+            $newKey = strtoupper(preg_replace('/[^[[:alnum:]]]*/', '', $key));
+            $categoryValues[$newKey] = $v;
+        }
+
+//        $appCategories = array_keys(File::config()->get('app_categories'));
```
chillu, line 31: Why is this commented out?

composer.json
```diff
@@ -5,8 +5,8 @@
     "license": "BSD-3-Clause",
     "require": {
         "silverstripe/framework": "^4.5",
-        "silverstripe/admin": "^1.7",
-        "silverstripe/graphql": "^3",
+        "silverstripe/admin": "1.x-dev",
```
chillu, line 8: This technically allows installing `silverstripe/admin` lower than `1.7` which was consciously bumped via https://github.com/silverstripe/silverstripe-asset-admin/commit/b17f5fda42fbc740fdee01a7cdb8627c1f04a2f6. Wouldn't a `^1.7` constraint also allow `1.x-dev`?

chillu (changes requested):

composer.json
unclecheese: Lots of frustration getting composer to be happy, and this may be an artefact. I'll revert and see what happens.

_graphql/mutations.yml
unclecheese: Because asset-admin never used scaffolding, and everything was a custom query/mutation, the graphql 4 implementation we did the same. Ideally, we should have migrated it to models, but I don't see a whole lot of value in that migration.

So because it's custom to custom, casing doesn't matter. You can be consistent across both.

chillu: Talked with Aaron, we decided to fix this anyway.

chillu (approved):

Issue #1120, opened 2020-07-24
Upload button broken after moving files


maxime-rainville:
If you move files to a different folder and navigate to that folder, the upload button doesn't work any more. Interestingly, the drag and drop for your file manager works just fine.

# Step to reproduce
- Access a folder with some pre-existing file and some pre-existing test folder
- Select the files
- Move the file to the test folder by dragging your selection over the folder.
- Navigate to the test folder
- Click the "Upload" button

Expected results: A file dialog allow pop up and allows you to select files to upload
Actual results: Nothing happens. This JS error is displayed: 
```
Uncaught TypeError: Cannot read property 'click' of null
    at HTMLButtonElement.click (bundle.js?m=1595570320:1)
click @ bundle.js?m=1595570320:1
```

[Youtube recording of the bug](https://youtu.be/YDdOcYEu138)

Reproduced in Chrome and Firefox. Didn't test in other browsers.

tom890180:
Firefox spews a variable:

```
Uncaught TypeError: e.hiddenFileInput is null
    click http://localhost/(project)/resources/vendor/silverstripe/asset-admin/client/dist/js/bundle.js?m=1600147545:1
    addEventListener http://localhost/(project)/resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1581903006:1
    value http://localhost/(project)/resources/vendor/silverstripe/asset-admin/client/dist/js/bundle.js?m=1600147545:1
    value http://localhost/(project)/resources/vendor/silverstripe/asset-admin/client/dist/js/bundle.js?m=1600147545:1
    value http://localhost/(project)/resources/vendor/silverstripe/asset-admin/client/dist/js/bundle.js?m=1600147545:1
    value http://localhost/(project)/resources/vendor/silverstripe/asset-admin/client/dist/js/bundle.js?m=1600147545:1
    value http://localhost/(project)/resources/vendor/silverstripe/asset-admin/client/dist/js/bundle.js?m=1600147545:1
    pn http://localhost/(project)/resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1581903006:1
    Dn http://localhost/(project)/resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1581903006:1
...
```

It seems to be failing on this line:
https://github.com/silverstripe/silverstripe-asset-admin/blob/1/client/src/containers/Editor/Editor.js#L161

[...]

PR #1165, opened 2020-11-23
FIX Allow file upload after drag and drop files move

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/1120


client/src/components/AssetDropzone/AssetDropzone.js
```diff
@@ -63,9 +63,6 @@ class AssetDropzone extends Component {
           this.props.options
         );
       }
-    } else {
-      // remove dropzone listeners (so it potentially doesn't interrupt other listeners)
-      this.dropzone.destroy();
```
emteknetnz, line 68: This doesn't actually remove any listeners, it just removes some things, notably the dropzone.hiddenFileInput which is what's causing the javascript error noted in the issue

maxime-rainville:
Can't replicate the original bug any more. If it's been fixed by some other means then maybe we don't need this PR.

maxime-rainville (approved):
Steve says he can still replicate the issue.

I tested the fix in IE11, Chrome and Firefox and I couldn't find any adverse affect. So I'll merge as is.

PR #1172, opened 2021-01-14
Update error messages to match DSM designs

bergice:
Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/895


maxime-rainville:
@clarkepaul Do you want to comment on this? While it doesn't fully implement the desired look and feel from the DSM, it takes it closer to it and addresses the immediate problem of the error message being truncated when too long.

# Status quo
image

# With PR
image

maxime-rainville (changes requested):
Error messages still get chopped on mobile. The DSM calls for the message to be dispalyed over multiple lines.

image

PR #1173, opened 2021-01-27
NEW: Compatibility with GraphQL 4 changes

unclecheese:
## Summary

Use new `UserContextProvider` API, and new sort resolver. Previous sort resolver was relying on something too model-specific, and asset-admin does not use models.

Depends on https://github.com/silverstripe/silverstripe-graphql/pull/352


unclecheese (commented):
I don't think this belongs in asset-admin. It's a concern of `admin` in general, because more modules than just asset-admin will rely on apollo and use interfaces and unions.

PR #1176, opened 2021-02-09
FIX Remove duplicate url dependency in package.json

scott1702:
Duplicate dependency in package.json, both entries point to the same version


emteknetnz:
Failing travis test is unrelated

emteknetnz (approved):

Issue #1146, opened 2020-10-15
File delete modal should use same data as new used on table


emteknetnz:
When deleting a file, the number of place(s) a file is used should use the same data source as the [updated used on table](https://github.com/silverstripe/silverstripe-admin/issues/1098)

![image](https://user-images.githubusercontent.com/4809037/96059569-c4b0f680-0eea-11eb-871f-522e5201672d.png)

## ACs
- [x] The new endpoint created as part of the used on table work is referenced
- [x] Performance impacts are considered. If there's significant delay, consider reviewing the modal wording with PUX
- [x] Modal wording is updated to report the number of files within the folder (a more performant query)
- [x] Implementation in PR is for GraphQL3 - create follow up issue to check in GraphQL4 update: it works fine in GraphQL4 so no need

## Notes
- There's a couple of scenarios to consider with this modal. This will need some broad testing when: deleting a folder, v.s. deleting a file, etc.
- Consider the performance impacts of deleting large folders and consider options for reducing this.

## PRs
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1178

emteknetnz:
The file count on the modal when you attempt to delete a folder is currently even more broken then when you attempt to delete a file

Setup:
- folder 'abc' which contains a 2x images, one of a fantail image, another of a tui image

- fantail sample image used 5x:
- 3x via page content
- 2x via has_one on a Page.has_one relations
- file usage table correct reports 5x usage
- file delete modal (use on "owns" endpoint) incorrectly reports 3x

- tui sample image used 3x
- 2x via page content
- 1x via has_one on Page.has_on
- file delete modal incorreclty reports 2x

However when I attempt to delete the parent 'abc' folder, the modal shows the following:

![image](https://user-images.githubusercontent.com/4809037/107583572-fa2edc00-6c5f-11eb-821d-e91d4c85f757.png)

So it seems the language is pretty misleading. What it's actually reporting is that "2 of the files contained in this folder are used AT LEAST ONCE (according to the 'owns' method)

brynwhyman:
Nice catch on the existing bug on how the modal is counting file usage @emteknetnz. For context these new GraphQL query endpoints [were created this relatively recently](https://github.com/silverstripe/silverstripe-asset-admin/issues/931).

[...]

PR #1178, opened 2021-02-18
NEW Update data source used for file delete modal

emteknetnz:
Issue silverstripe/silverstripe-asset-admin#1146

[Video demo](https://www.youtube.com/watch?v=dQPwc8QFvIQ)

[Behat failure](https://travis-ci.com/github/silverstripe/silverstripe-asset-admin/jobs/486370504#L1968) is an existing failure


emteknetnz:
@clarkepaul Could you review the video demo I linked above and ensure this behaves as you expect it to.

Let me know if you'd like a further demo.  Otherwise just let me know that you're happy with how things as they are as a comment on this pull-request.  Cheers.

clarkepaul:
Hey, minor change to simplify the text a bit. We were using the word _folder_ instead of _file_ .

The same change for all three scenarios... let me know if that makes sense to you, if so then I'm all good with it :)
delete_folder_and_file
Pasted_Image_2_03_21__4_47_PM

delete_single_file

emteknetnz:
Thanks for the review, it now looks like this:

image

I'll move this to code review now

Are you happy to sign off the design review part of this?

client/lang/src/en.json
```diff
@@ -9,6 +9,7 @@
     "AssetAdmin.BULK_ACTIONS_DELETE_FOLDER": "These folders contain files which are currently in use, you must move or delete their contents before you can delete the folder.",
     "AssetAdmin.BULK_ACTIONS_DELETE_MULTI_CONFIRM": "There are %s files currently in use, are you sure you want to delete these files?",
     "AssetAdmin.BULK_ACTIONS_DELETE_SINGLE_CONFIRM": "This file is currently in use in %s places, are you sure you want to delete it?",
+    "AssetAdmin.BULK_ACTIONS_DELETE_ITEMS_CONFIRM": "Warning: You're about to delete %s file(s) which may be used in your site's content. Please carefully check the file usage to ensure they are removed from the content areas prior to deleting them, otherwise they will appear as broken links.",
```
maxime-rainville, line 12: The message is displayed in an obnoxious modal. Is it not clear enough it's a warning?
```suggestion
    "AssetAdmin.BULK_ACTIONS_DELETE_ITEMS_CONFIRM": "You're about to delete %s file(s) which may be used in your site's content. Carefully check the file usage to ensure they are removed from the content areas prior to deleting them, otherwise they will appear as broken links.",
```

maxime-rainville (commented):
This is just a partial review.

Main thing I'm wondering is why not update the GraphQL service to return a single number containing the total number of selected files for your current selection?

That way you could remove a bunch of frontend logic for adding all these numbers up. That was only needed because we needed a way to pick which message to show.

Also, the message looks kinda of weird when deleting empty folders.

emteknetnz:
> Main thing I'm wondering is why not update the GraphQL service to return a single number containing the total number of selected files for your current selection?

Reason being that
- we *might* at a future date want to use the data to overlay the number of files within folder on top of the folder icons in asset-admin.
- it made it easier to update the logic on the frontend since it was reusing the old way of doing things

This is possibly pointless and over engineered though.  Happy to change if you think it's worthwhile.

_legacy/GraphQL/DescendantFileCountType.php
```diff
@@ -0,0 +1,45 @@
...
+
+if (!class_exists(TypeCreator::class)) {
+    return;
+}
+
+/**
+ * Define the return type for ReadDescendantFileCountsQueryCreator
+ * Return as an array of object with an 'id' property and 'count' property.
+ *
+ * @deprecated 4.8..5.0 Use silverstripe/graphql:^4 functionality.
```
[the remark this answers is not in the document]
emteknetnz: I don't think it should get @internal, simply because it won't match all the other graphql3 stuff

Good point about the version number

client/lang/src/en.json
emteknetnz: @clarkepaul ?

clarkepaul: Was a bit torn on this one as the "waring:" text isn't best practice but the circumstances of not reading the text could be pretty damaging. It was only there because we are being a little lazy with the visuals communicating its importance e.g. no warning icon. I've been swayed towards removing it as Max has a good point, the whole modal acts as the warning. 

FYI its also best practice to avoid using the word "please" in UIs so that is good it was removed.

brynwhyman: As per @clarkepaul's comment above, this still needs actioning

clarkepaul:
My honest thought is that it doesn't need a confirm when its an empty folder, it should just delete.
It might be like that because the same pattern is used for a file? if thats the case then the header could be simplified to just "Confirm deletion" as less words and more to the point. Otherwise what @brynwhyman has mocked is okay too.

brynwhyman:
Okay, so priority order of changes (depending on effort)
a. Remove modal when deleting an empty folder
b. Change folder wording to: "Are you sure you want to delete this folder?" with a modal heading of: "Confirm deletion" ?
  *  modal heading of: "Confirm deletion" should apply to deleting files and or folders

client/src/containers/BulkDeleteConfirmation/BulkDeleteMessage.js
```diff
@@ -1,114 +1,43 @@
...
-const BulkDeleteMessage = ({ foldersInUse, filesInUse, itemCount }) => (
+const BulkDeleteMessage = ({ folderDescendantFileTotals, fileTotalItems }) => (
   <Fragment>
-    <p>{confirmationMessage(foldersInUse, filesInUse, itemCount)}</p>
-    {(foldersInUse.totalItems > 0 || filesInUse.totalItems > 0) && <p>{i18n._t(
-      'AssetAdmin.BULK_ACTIONS_DELETE_WARNING',
-      'Ensure files are removed from content areas prior to deleting them, otherwise they will appear as broken links.'
-    )}</p>}
+    <p>{confirmationMessage(folderDescendantFileTotals, fileTotalItems)}</p>
   </Fragment>
```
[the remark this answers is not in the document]
emteknetnz: `<Fragment>` is required because there are 2 `<p>` tags

client/lang/src/en.json
emteknetnz: This was previously actioned

_legacy/GraphQL/ReadDescendantFileCountsQueryCreator.php
```diff
@@ -0,0 +1,73 @@
...
+use SilverStripe\ORM\DataList;
+use SilverStripe\Versioned\Versioned;
+
+if (!class_exists(QueryCreator::class)) {
+    return;
+}
+/**
+ * GraphQL Query to retrieve the file count within a folder
+ *
+ * @deprecated 4.8..5.0 Use silverstripe/graphql:^4 functionality.
```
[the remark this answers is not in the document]
emteknetnz: Updated

_legacy/GraphQL/DescendantFileCountType.php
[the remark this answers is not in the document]
emteknetnz: Updated version number

emteknetnz:
Have made the following changes:
a) Modal heading is now always "Confirm deletion"
b) When deleting a single empty folder, the message is "Are you sure you want to delete this folder?"
c) When deleting more than one empty folder, the message is "Are you sure you want to delete these folders?"

emteknetnz:
Have confirmed this DOES work correctly with GraphQL4 as well as GraphQL3

clarkepaul:
I've had a quick squiz at the end user experience with the following scenarios (file delete, empty folder delete, multiple empty folders delete, folders delete with files in them) and happy with the end result in terms of language used :).

As a side not there are issues selecting files across multiple paginated pages but that is a separate issue.

code/GraphQL/Resolvers/AssetAdminResolver.php
```diff
@@ -217,6 +218,34 @@ public static function resolvePublicationResultUnion($value): string
         }
     }
 
+    public static function resolveReadDescendantFileCounts($object, array $args, $context, ResolveInfo $info): array
+    {
+        /** @var DataList|File[] $files */
+        /** @var File|AssetAdminFile $file */
+        if (!is_array($args['ids'] ?? 0)) {
```
maxime-rainville, line 225: Is the point of the `??` operator here to squash the "Undefined index" notice?

Seems like a somewhat abusive approach. I rather we did something a little more explicit like this.

```suggestion
        if (!isset($args['ids']) || !is_array($args['ids'])) {
```

code/GraphQL/Resolvers/AssetAdminResolver.php
```diff
@@ -217,6 +218,34 @@ public static function resolvePublicationResultUnion($value): string
         }
     }
 
+    public static function resolveReadDescendantFileCounts($object, array $args, $context, ResolveInfo $info): array
+    {
+        /** @var DataList|File[] $files */
```
maxime-rainville, line 223: Why not put this type definitions directly above where `$files` is used for the first time?

code/GraphQL/Resolvers/AssetAdminResolver.php
```diff
@@ -217,6 +218,34 @@ public static function resolvePublicationResultUnion($value): string
...
+        if (!is_array($args['ids'] ?? 0)) {
+            throw new \InvalidArgumentException('ids must be an array');
+        }
+        $ids = $args['ids'];
+
+        $files = Versioned::get_by_stage(File::class, Versioned::DRAFT)->byIDs($ids);
+        if ($files->count() < count($ids)) {
+            $class = File::class;
+            $missingIds = implode(', ', array_diff($ids, $files->column('ID')));
+            throw new \InvalidArgumentException("{$class} items {$missingIds} are not found");
```
maxime-rainville, line 234: Does it matter all that much if our user somehow sends us IDs that don't exists?

I'm thinking of a scenario where maybe someone is about to delete some files and another user ends up deleting them just beforehand. In that scenario we would throw a nasty error for something that doesn't sound all that consequential.

tests/php/GraphQL/Legacy/ReadDescendantFileCountsQueryCreatorTest.php
```diff
@@ -0,0 +1,103 @@
...
+
+    protected static $fixture_file = '../../fixtures.yml';
+
+    public function setUp()
+    {
+        parent::setUp();
+        if (class_exists(Schema::class)) {
+            $this->markTestSkipped('GraphQL 3 test ' . __CLASS__ . ' skipped');
+        }
+        TestAssetStore::activate('AssetAdminTest');
```
maxime-rainville, line 29: It's not clear to me you need physical files for this test to work because you're only interacting with DB records.

maxime-rainville (changes requested):
> > Main thing I'm wondering is why not update the GraphQL service to return a single number containing the total number of selected files for your current selection?
> 
> Reason being that
> - we *might* at a future date want to use the data to overlay the number of files within folder on top of the folder icons in asset-admin.
> - it made it easier to update the logic on the frontend since it was reusing the old way of doing things
> 
> This is possibly pointless and over engineered though.  Happy to change if you think it's worthwhile.

My main beef is this could end up making this request substantially slower in some cases where you have thousand of files. Sounds like a big price to pay for something we might want to do.

emteknetnz:
>> My main beef is this could end up making this request substantially slower in some cases where you have thousand of files. Sounds like a big price to pay for something we might want to do.

I'm not sure I follow

(it's been so long since I did the backend for this I may be making some incorrect assumptions)

a)
Is the concern that now the query is readying nested files, there's a risk that it will end up slower because all the addition queries required?  The only other option would be to simply not attempt to read the number of files?  The previous implementation was straight out inaccurate, so it was worse than not showing a number

b)
Or is the convern that the formatting of the output was an array of some key/values rather than just a integer?  This will have virutally zero bearing on performance.  An array with tens of thousands of values will still download very fast.  Also I think we now require this to work out only folders with no files are selected which impacts the message we show the user.

c)
Something else

@maxime-rainville Let me know which of the above you are concerned about

maxime-rainville:
Sorry ... I had a brain freeze. Was trying to make sense of my previous comments. Ran a test with 10000+ folder and got a response within 2 sec on my local ... which I think is reasonable.

Tested this locally on a plain install with/without GraphQL v4 and CWP install with/without GraphQL v4. Everything worked fine.

Just need to address my other comments.

code/GraphQL/Resolvers/AssetAdminResolver.php
emteknetnz: It's so that it's consistent with how the rest of the system currently works e.g. https://github.com/silverstripe/silverstripe-asset-admin/blob/1/_legacy/GraphQL/DeleteFileMutationCreator.php#L54

code/GraphQL/Resolvers/AssetAdminResolver.php
emteknetnz: Updated

code/GraphQL/Resolvers/AssetAdminResolver.php
emteknetnz: Updated

_legacy/GraphQL/ReadDescendantFileCountsQueryCreator.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        /** @var DataList|File[] $files */
+        /** @var File|AssetAdminFile $file */
+        if (!is_array($args['ids'] ?? 0)) {
+            throw new \InvalidArgumentException('ids must be an array');
+        }
+        $ids = $args['ids'];
+
+        $files = Versioned::get_by_stage(File::class, Versioned::DRAFT)->byIDs($ids);
+        if ($files->count() < count($ids)) {
```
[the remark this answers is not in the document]
emteknetnz: It's so that it's consistent with how the rest of the system currently works e.g. https://github.com/silverstripe/silverstripe-asset-admin/blob/1/_legacy/GraphQL/DeleteFileMutationCreator.php#L54

_legacy/GraphQL/ReadDescendantFileCountsQueryCreator.php
```diff
@@ -0,0 +1,73 @@
...
+                'type' => Type::nonNull(Type::listOf(Type::id())),
+            ],
+        ];
+    }
+
+    public function resolve($object, array $args, $context, ResolveInfo $info)
+    {
+        /** @var DataList|File[] $files */
+        /** @var File|AssetAdminFile $file */
+        if (!is_array($args['ids'] ?? 0)) {
```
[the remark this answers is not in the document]
emteknetnz: Updated

tests/php/GraphQL/Legacy/ReadDescendantFileCountsQueryCreatorTest.php
emteknetnz: Removed

_legacy/GraphQL/ReadDescendantFileCountsQueryCreator.php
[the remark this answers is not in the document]
maxime-rainville: Might be nice to add a `@throws` comment to make this explicit.

code/GraphQL/Resolvers/AssetAdminResolver.php
maxime-rainville: Might be nice to add a `@throws` comment to make this explicit.

maxime-rainville (approved):
Looks good ... left some optional suggestions. Do with them as you please.

maxime-rainville:
@clarkepaul Do you have any outstanding concerns with this one?

clarkepaul:
@maxime-rainville was only looking over @emteknetnz 's shoulder for the review but couldn't see any issues, so I'm happy with it.

PR #1182, opened 2021-03-04
#437: Ensure folder names are shown when a user zooms in/out

adrhumphreys:
This is to resolve: https://github.com/silverstripe/silverstripe-assets/issues/437

Opted to use a relative width rather than a hardcoded one so that it should work at any zoom level


emteknetnz:
We've merged this pull-request instead https://github.com/silverstripe/silverstripe-asset-admin/pull/1185

Issue #1147, opened 2020-10-15
Zooming in/out doesn't move folder titles properly


b-rhodes:
When in the files section of the CMS, if you zoom the web page out (cmd + `-` or ctrl + `-`) the text behaves strangely.
This results in the text being below the folders, and if there are multiple rows of folders, they can overlap the text so you can't see the text properly.

This is only an issue for folders, I could not replicate the issue with images and files.

This issue is present in SilverStripe framework 4.6.0 (asset-admin 1.6.0), as well as in SilverStripe framework 4.5.1 (asset-admin 1.5.1). I haven't checked any other versions.

Screenshots:
![image](https://user-images.githubusercontent.com/25066797/96070907-98eb3c00-0efd-11eb-95c0-4c7905f58ebf.png)
![image](https://user-images.githubusercontent.com/25066797/96071037-d8b22380-0efd-11eb-8299-c5c9961a0a3b.png)

# Pull requests
- [ ] #1185
OR 
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/1182

lerni:
I cannot reproduce this locally (asset-admin 1.6.1) nor on demo https://demo.silverstripe.org/admin/assets/ also not with different browsers FF/Chrome/IE11. Does it behave different on demo.silverstripe.org?

lerni:
I'm on Mac Chrome 86.0.4240.80 and also checked some browserstack-variants - removed some font-settings, hdpi and not, different languages... but but still cannot reproduce with the same steps :(
![Bildschirmfoto 2020-10-15 um 11 13 28](https://user-images.githubusercontent.com/472115/96103240-c1f8d600-0ed7-11eb-8bf7-d26c2a457468.png)

michalkleiner:
I can replicate this in Chrome 86.0.4240.80 on macOS Catalina 10.15.3

Zooming in is fine, zooming out (less than 100%) breaks it. @lerni are you zooming **out**?

[
![assets zoomed out](https://user-images.githubusercontent.com/233342/96111876-4c742080-0f3e-11eb-952f-ec34fe88c66e.png)
](url)

lerni:
@michalkleiner & @b-rhodes with zooming-out I can break it :) Sorry for the hassle. Above 100% is fine but not below.

With `.gallery-item__title` `width: calc(100% - 45px);` instead of `width: 131px;` works.

lerni:
[...]

PR #1185, opened 2021-03-11
BUG Set folder gallery item display to flex to allow it to scale with browser zoom level

maxime-rainville:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/1147


maxime-rainville:
I tested this in Firefox, chrome and IE11.

emteknetnz (approved):
> I tested this in Firefox, chrome and IE11.

Me too :-)

PR #1201, opened 2021-05-12
Compatibility with new updateSchema signature

unclecheese:
Depends on https://github.com/silverstripe/silverstripe-graphql/pull/374


unclecheese:
Unnecessary API change. Have removed from the graphql PR.

PR #1206, opened 2021-06-02
Preserve comment nodes when editing in tinymce.

### LINKED ISSUE #1190: Saving page strips html comments
Coming from SS 3.x, our page editors were used to being able to insert HTML comments in some pages using the tinymce source code browser.

In 4.x (framework 4.7.3, admin 1.7.3, asset-admin 1.7.1 are our exact versions),  HTML comments are being stripped away on save despite our configuring Tiny MCE to accept them.

We traced this down to "TinyMCE_ssmedia.js", specifically in its SaveContent handler.
Here, some work is done to deal with image shortcodes, then the result of that is serialized back to HTML by walking the resulting set of nodes and concatenating the "outerHTML" of each.
Because comment nodes do not have an outerHTML, they are skipped and discarded from the result.

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1206

dratr:
To test this, add this code to your project's `_config.php` file:
```php
<?php

use SilverStripe\Forms\HTMLEditor\TinyMCEConfig;

TinyMCEConfig::get('cms')->setOption('valid_elements', TinyMCEConfig::get('cms')->getOption('valid_elements') . ',--[*]');
```
Then add an HTML comment to any WYSIWYG field that uses the default `cms` config.

## Issue
- https://github.com/silverstripe/silverstripe-asset-admin/issues/1190


dhensby:
Please address the linting issues

dhensby (changes requested):
Incorrect variable name used and other linting issues

ChloeSartorelli:
I've retargetted and rebased this PR to `1.13` to it can be released as a patch.
I've also changed the magic numbers into their corresponding consts (`8` is `COMMENT_NODE` and `3` is `TEXT_NODE` - see https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType).
Finally, I ran `yarn build` so this includes the actual distribution files required for the change to take effect.

emteknetnz (changes requested):
I'm not sure why we need this?

I tried locally with and without this PR using the updated _config.php and I was able to add comment tags both using TinyMCE normally as well as when viewing the HTML source using the editor. In both scenarios the HTML comments were retained

Am I doing something wrong?  Possibly since this PR was originally made the underlying TinyMCE was updated to support HTML comments?

Let us know if this PR is still needed, if not we'll close it.

ChloeSartorelli:
> In both scenarios the HTML comments were retained

Did you save the form and reload? And were you using CMS 4?

emteknetnz:
Yes and yes

ChloeSartorelli:
Strange - I had tested it before picking this up and saw it now working without this PR, but now when I test it without the PR it works fine. Gonna close this since it seems to no longer be necessary.

PR #1212, opened 2021-06-17
FIX Update overrides url when file is selected

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/2621

This code should have been added when this work was originally done https://github.com/silverstripe/silverstripe-asset-admin/pull/1115


client/src/containers/InsertMediaModal/InsertMediaModal.js
```diff
@@ -197,13 +197,21 @@ function mapStateToProps(state, ownProps) {
...
-  const requireTextFieldUrl = ownProps.requireLinkText ? '?requireLinkText=true' : '';
+  const arr = [];
+  if (ownProps.requireLinkText) {
+    arr.push('requireLinkText=true');
+  }
+  if (ownProps.fileSelected) {
+    arr.push('fileSelected=true');
+  }
+  const joined = arr.join('&');
+  const qs = arr.length ? `?${joined}` : '';
```
maxime-rainville, line 208: We have libraries to interact with URLs. I would rather we use those instead of using string concatenations to build URLs. e.g.:

https://github.com/silverstripe/silverstripe-asset-admin/blob/4a68b7ca6055699d4b79dc1fe13e9f48cba446ef/client/src/components/PreviewImageField/PreviewImageField.js#L257-L259

maxime-rainville (changes requested):
The fix appears to resolve the problem nicely. Since this is a high impact bug, I think it's worth back porting it to 4.7.

emteknetnz:
Closed in favour of 1.7 branch version of PR https://github.com/silverstripe/silverstripe-asset-admin/pull/1213

PR #1213, opened 2021-06-20
FIX Update overrides url when file is selected

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/pull/1212


maxime-rainville (approved):
I think something like this would be more elegant. But your way is OK.

```js
  const query = {};
  if (ownProps.requireLinkText) {
    query.requireLinkText = true;
  }
  if (ownProps.fileSelected) {
    query.fileSelected = true;
  }
  const pathname = `${schemaUrl}/${targetId}`;
  const newSchemaUrl = urlLib.format({pathname, query});

  // set schemaUrl for `fileSchemaModalHandler` to load the default form values properly
  // This schema URL is not actually passed down to the Editor, it's just use to set the
  // form schema overrides
  return {
    schemaUrl: newSchemaUrl,
    type: formSchema && formSchema.type
  };
```

Tested in Fiefox and IE11

Issue #1214, opened 2021-06-22
CMS field for managing image `loading` state


brynwhyman:
### Overview
With images moving towards being lazy-loaded by default, there will still be a place on a per image basis to opt-out of this setting. This issue focuses on creating a field in the CMS for content editors to decide to 'turn off' lazy-loading for a specific image.

A key example where this might be used is where the content editor knows that the image they are adding to a web page will be in view when the page loads ('above the fold') and so should load automatically.

Background: https://forum.silverstripe.org/t/browser-level-lazy-loading-from-silverstripe-cms/4218

### Acceptance Criteria
- [x] There is a CMS field in the image edit form that the user can use to turn lazy-loading on or off for the specified image
- [x] If the developer has [globally opted out of the feature](https://github.com/silverstripe/silverstripe-assets/issues/452), the CMS field is not shown
- [x] There is enough information presented to the user in the context of the field that they understand what the feature does
- [x] If the attribute is broken for the image (e.g. the user manually fudged the values through the WYSIWYG HTML edit field) default to the 'false' state
- [x] 4.9.0 changelog has been updated to reflect language used on the component
- [x] Behat test js added to support the field
- [x] Option 3 (Dropdown) is implemented
- [x] The field is at the bottom of the form (validate with Paul)
- [x] The "tooltip on label" is implemented globally for all fields and replaces the bulb icon (validate with Paul)

### Notes
- This should require behat tests

### PRs
- [x] https://github.com/silverstripe/silverstripe-behat-extension/pull/203
- [x] https://github.com/silverstripe/silverstripe-framework/pull/10029
- [x] https://github.com/silverstripe/silverstripe-admin/pull/1239
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1218

clarkepaul:
Based on team conversations about various approaches [I've assembled 4 main possible directions](https://invis.io/WCMZMHHXRJV#/454169956_Lazy_Loading) we could go in. Would be good to get feedback from anyone who has opinions.

brynwhyman:
Just for a bit more context, I've just had another look through some of the other implementations that we've been referencing. It doesn't appear that Wordpress or Typo3 actually provide a feature within the CMS UI to disable this on a per-image basis. See:
[...]

PR #1218, opened 2021-06-23
NEW Add image form input to opt out of lazy loading

emteknetnz:
Needs this to be merged and released for behat tests to pass https://github.com/silverstripe/silverstripe-behat-extension/pull/203
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1214


client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -354,17 +361,18 @@ jQuery.entwine('ss', ($) => {
...
     getAttributes() {
       const data = this.getData();
-
-      return {
+      const obj = {
         src: data.url,
         alt: data.AltText,
         width: data.Width,
         height: data.Height,
+        'data-eager-load': data.EagerLoad,
```
maxime-rainville, line 369: Most of the other attributes match the actual attribute that will be outputted in the `` tag. I think the attribute should be called loading and store a string so the shortcode is as close as possible to the actual rendered HTML.

emteknetnz: This is just a temporary thing for the checkbox.  Checkboxes only accept [0's and 1's ](https://github.com/silverstripe/silverstripe-framework/blob/4/src/Forms/CheckboxField.php#L15)

maxime-rainville: Maybe use a dropdown then. Otherwise, we'll have to refactor the entire logic once we decide what actual field we're going to use.

emteknetnz: Will keep as is - the AC is `A 'dummy' checkbox is created to enable/ disable lazy-loading`

maxime-rainville: The point was to have a field ... not that it would be a checkbox.

emteknetnz: It was decided during planning that we would use a checkbox as a temporary form control.  The designs currently have mixture of checkbox an non-checkbox options.  We may end up going with a checkbox option in which case it would make sense to keep this code more or less as is.  I think refactoring this now given we don't know what the final design state is would just be a potential waste of time.

maxime-rainville: I'm the one who went to bat to get this in ACs in I'm telling you I don't care if it's a checkbox or a dropdown.

If you absolutely want a checkbox in, then make sure it saves data in the right format.

emteknetnz: As previously mentioned - (Silverstripe) Checkboxes only accept [0's and 1's ](https://github.com/silverstripe/silverstripe-framework/blob/4/src/Forms/CheckboxField.php#L15)

So seems like we use the value 'eager' - at least that was my experience because I originally tried passing a string value before changing it to an int when I ran into this issue.

bergice: I agree we should be saving the data in the format of `loading="<string>"`. If we need to change this later and the data is saved using a different attribute and using a `bool` it will be a pain to change it.

I don't see the point in creating our own API's that handle this logic when we can just use the officially supported browser API that everyone will understand immediately out of the box.

If you just want to get this merged as is, then make sure we follow up on this later.
But to be fair, it looks like 90% of this PR is all about handling the data this way, so... Maybe we should just change it now?

bergice (changes requested):

emteknetnz:
The problem here is that we're blocked waiting to decide what we're doing with https://github.com/silverstripe/silverstripe-asset-admin/issues/1214

There's a few outcome from this that will affect this PR:
a) If we decide that we don't actually want this field after all, this we should just remove most of the code in this PR.
b) If we do decide that we want something, likely a drop down, then this PR should be updated to use a string value
c) If we decide that we want a checkbox (unlikely at this stage), then this code should stay as is

Since a decision hasn't been made at this stage, I'll move this issue to blocked

client/src/entwine/TinyMCE_ssmedia.js
maxime-rainville: ACs are not suicide packs and can be change after the fact. If no one is making the argument that it makes a difference if this use dropbox instead of a checkbox, let's just change the AC.

code/Forms/ImageFormFactory.php
```diff
@@ -59,23 +62,44 @@ protected function getFormFieldAttributesTab($record, $context = [])
...
+                'Loading',
+                _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.Loading', 'Loading'),
+                [
+                    '' => _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.LazyDefault', 'Lazy (default)'),
+                    'eager' => _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.Eager', 'Eager'),
+                ]
+            )
+                ->setTitleTip(new Tip($titleTipContent));
+            $dimensionsLoading->push($field);
+        }
```
maxime-rainville, line 103: Probably would be better if this bit of logic was in a private method for clarity.

code/Forms/ImageFormFactory.php
```diff
@@ -135,7 +159,7 @@ public function getForm(RequestHandler $controller = null, $name = FormFactory::
             /** @var FieldList $fields */
             $fields = $form->Fields();
 
-            $dimensions = $fields->fieldByName('Editor.Placement.Dimensions');
+            $dimensions = $fields->fieldByName('Editor.Placement.DimensionsLoading.Dimensions');
```
maxime-rainville, line 162: I suspect other people out there have attempted to customise this field and access it by its group name. Reorganising that group will force them to update their custom code.

I don't think that would be the end of the world and our semver commitment doesn't extend to field names. But we should be mindful about avoiding changing a field's name unless it can't be avoided.

If we do decide to update that field name, we should note it in the change logs.

code/Forms/ImageFormFactory.php
```diff
@@ -59,23 +62,44 @@ protected function getFormFieldAttributesTab($record, $context = [])
 
         $tab->insertAfter(
             'Alignment',
-            FieldGroup::create(
-                NumericField::create(
-                    'Width',
-                    _t('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin.ImageWidth', 'Width')
+            $dimensionsLoading = FieldGroup::create(
```
maxime-rainville, line 65: In a perfect world, a `FieldGroup` would be used to organise related fields together. Here, our groups are only used for layout purposes, which is not ideal.

maxime-rainville (changes requested):
I'm really not a big fan of having the "Loading" field on the same line as the dimension.

It looks weird on some resolution. That screen shot got taken on a 1999 pixel window
image

It's not clear to me there's any logical relation between the dimension and the lazy loading attribute.

It also leads to weird/hackish logic and CSS.

Unless someone wants to make a case there's substantial value in putting this field next to the dimensions, I would rather we keep this simple and just have a plain field on its own line.

tests/behat/features/insert-an-image.feature
```diff
@@ -36,14 +36,15 @@ Feature: Insert an image into a page
...
+  Scenario: I can edit an image in the file modal
     When I press the "Insert from Files" HTML field button
       And I select the file named "folder1" in the gallery
       And I click on the file named "file1" in the gallery
       Then I should see the "Form_fileInsertForm" form
       And I should see the "Insert file" button
+      Then I should see the "#Form_fileInsertForm_Loading option[value='']:checked" element
     When I fill in "Alternative text (alt)" with "My alt"
-      # Filling in width is purely to get the framework to scroll-up so that the details button is in view
+      Then I select "eager" from the "Form_fileInsertForm_Loading" field
```
maxime-rainville, line 47: Did you try this?

```suggestion
      Then I select "eager" from "Loading" input group
```

tests/behat/features/insert-an-image.feature
```diff
@@ -58,16 +59,21 @@ Feature: Insert an image into a page
       Then I should see the "Form_fileEditForm" form
     When I click on the ".editor-header__back-button" element
       Then I should see the "Form_fileInsertForm" form
+      And I should see the "#Form_fileInsertForm_Loading option[value='eager']:checked" element
```
maxime-rainville, line 62: Did you try something like this?
```suggestion
      And the "Loading" field should contain "eager"
```

emteknetnz:
> Unless someone wants to make a case there's substantial value in putting this field next to the dimensions, I would rather we keep this simple and just have a plain field on its own line.

^ @clarkepaul 

Built following designs here - https://projects.invisionapp.com/share/WCMZMHHXRJV#/screens/454169956_Lazy_Loading

clarkepaul:
> Unless someone wants to make a case there's substantial value in putting this field next to the dimensions, I would rather we keep this simple and just have a plain field on its own line.

@maxime-rainville this was designed with the help of feedback from users. If placed in its own line it either pushes other important fields out of view, or if placed at the bottom is out of view and would not be seen. The idea of having the loading field close to other presentational aspect of the file made sense to be in the vicinity of each other. With alignment, width and height being the main fields people thought it related to. I'm keen to leave it where it is.

code/Forms/ImageFormFactory.php
```diff
@@ -2,7 +2,10 @@
 
 namespace SilverStripe\AssetAdmin\Forms;
 
+use SilverStripe\Assets\Image;
 use SilverStripe\Control\RequestHandler;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Forms\CheckboxField;
```
maxime-rainville, line 8: Presumably this is not needed anymore?

```suggestion
```

maxime-rainville:
If we're confident this layout provides value, I'm fine keeping it like this.

If this situation happens again in some other context, we should come up with a better more systematic way of doing this.

Otherwise, the approach is adequate for a one-off edge case. I still would like to see the CSS improved and me need to mention the new group name in the changelog.

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,42 @@
     justify-content: center;
   }
 }
+
+#Form_fileInsertForm_DimensionsLoading_Holder {
```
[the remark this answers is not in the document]
emteknetnz: I agree this is ugly CSS that's hacking elements to behave in ways they were not intended to.  However I'm really not sure how to get around doing this unfortunately.

The core issues is that the `ProportionConstraintField`, which has its own FieldGroup, was originally intended to be on a row by itself.  Now it needs to share that row with this new Loading field.  

If the Loading field was on its own row there would be no need for any of this CSS

The only way I know of to split the row between the ProportionConstraintField FieldGroup and the Loading FormField was two wrap them both in a 'DimensionsLoading' FieldGroup, which is a bit of a misuse of a FormGroup

Do you know of a different way to horizontally split these?

code/Forms/ImageFormFactory.php
emteknetnz: Agreed, though I don't know of another method to do the horizontal layout

code/Forms/ImageFormFactory.php
emteknetnz: I don't think so, it's in one of those long function that has lots of similar looking code

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -92,7 +92,10 @@ const filter = 'img[data-shortcode="image"]';
               // don't save caption, since that's in the containing element
               title: el.attr('title'),
               alt: el.attr('alt'),
-            });
+              loading: el.data('loading')
+            };
+
+            const properties = sanitiseShortCodeProperties(obj);
```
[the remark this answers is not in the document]
emteknetnz: Updated

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -364,7 +368,9 @@ jQuery.entwine('ss', ($) => {
         class: data.Alignment,
         'data-id': data.ID,
         'data-shortcode': 'image',
+        'data-loading': data.Loading,
       };
+      return obj;
```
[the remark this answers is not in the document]
emteknetnz: Updated

code/Forms/ImageFormFactory.php
emteknetnz: Removed

tests/behat/features/insert-an-image.feature
emteknetnz: I tried it, it does not work

tests/behat/features/insert-an-image.feature
emteknetnz: Yup that works, there's another couple of css selectors in the file I changed too using field names

maxime-rainville (changes requested):

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,53 @@
...
+  #Form_fileInsertForm_Loading_Holder {
+    width: 33%;
+    margin-right: 0;
+
+    select {
+      height: 36px;
+    }
+
+    .tip.tip--title {
+      float: right;
```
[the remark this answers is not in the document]
emteknetnz: @clarkepaul  `float: right` was specifically for the 'Loading' field to match the design, should it be removed to be consistent?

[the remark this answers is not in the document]
clarkepaul: I think floating right on all of them would be good, I can't think of any labels where that wouldn't work. What I like about the right positioning is that the label of the field seems cleaner to read and the label flow into the field isn't obstructed, e.g. eye's stopping at the (?) icon.

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,53 @@
...
+    width: 100%;
+  }
+
+  #Form_fileInsertForm_Dimensions_Holder {
+    width: 67%;
+
+    label {
+      font-size: inherit;
+      color: inherit;
+    }
```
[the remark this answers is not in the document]
emteknetnz: Removed

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,53 @@
...
+
+    .field-group-component {
+      display: flex;
+      flex-wrap: wrap;
+
+      .field {
+        width: 50%;
```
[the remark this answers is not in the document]
emteknetnz: Removed

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,53 @@
...
+        margin-bottom: 0;
+
+        &:last-child  {
+          margin-right: $spacer / 2;
+        }
+      }
+
+      .image-size-preset-list {
+        width: 100%;
+      }
```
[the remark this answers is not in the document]
emteknetnz: Moved

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,53 @@
...
+      display: flex;
+      flex-wrap: wrap;
+
+      .field {
+        width: 50%;
+        margin-bottom: 0;
+
+        &:last-child  {
+          margin-right: $spacer / 2;
+        }
```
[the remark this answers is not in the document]
emteknetnz: Removed

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,53 @@
...
+
+      .image-size-preset-list {
+        width: 100%;
+      }
+    }
+  }
+
+  #Form_fileInsertForm_Loading_Holder {
+    width: 33%;
+    margin-right: 0;
```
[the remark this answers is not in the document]
emteknetnz: Removed

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,53 @@
...
+
+    select {
+      height: 36px;
```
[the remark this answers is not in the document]
emteknetnz: Have updated globally

client/src/containers/Editor/Editor.scss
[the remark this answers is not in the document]
emteknetnz: Have changed to using float: right for everything

clarkepaul:
@emteknetnz took me through the UI following fixes and looks all good to me :) :
 - Tabbing ands focus of icon
 - Active of icon (popover open)
 - Hover of icon
 - Icon position (right) in relation to label
 - Colour variant of icon

tests/behat/features/insert-an-image.feature
maxime-rainville: Does `Then I select "eager" from "Loading" input group` do something different than `Then I select "eager" from the "Loading" field`? 

If not, than let's use the "input group" version and ditch the behat PR.

client/src/containers/Editor/Editor.scss
```diff
@@ -185,3 +185,53 @@
...
+.editor .tab-pane .show {
+  display: flex;
+  flex-wrap: wrap;
+  margin-top: -10px;
+
+  & > div {
+    width: 100%;
+  }
+
+  #Form_fileInsertForm_Dimensions_Holder {
```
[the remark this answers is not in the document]
maxime-rainville: Raised a separate issue. https://github.com/silverstripe/silverstripe-admin/issues/1243

maxime-rainville (changes requested):
Just need to confirm that we really [need that extra Behat scenario](https://github.com/silverstripe/silverstripe-asset-admin/pull/1218#discussion_r681435628).

Otherwise, everything works great.

tests/behat/features/insert-an-image.feature
emteknetnz: `Then I select "eager" from "Loading" input group` doesn't work, so we need the keep the `Then I select "eager" from the "Loading" field` version

maxime-rainville (approved):
All is right with the world and the crooked places have been made straight.

PR #1221, opened 2021-07-02
FIX Save fetched files so details button works after pagination

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1136


maxime-rainville (changes requested):
That's super dodgy and won't work if you try accessing a file directly by its link.

emteknetnz:
>> That's super dodgy

Is it?  It's just caching a result

>> won't work if you try accessing a file directly by its link.

Not sure I follow.  Do you mean
a) accessing a file directly in the CMS like /admin/assets/show/13/edit/73? That still works fine.  
b) making a graphql request filtering on a file id - that would just use the regular graphql query and not the cache version?
c) something else

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -60,6 +64,14 @@ class AssetAdmin extends Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
+    // Save file data in case we need to use it later after paginating the gallery
+    prevProps.files.forEach(file => {
```
maxime-rainville, line 69: Using [`Array.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) would be better here 

Not that it matters.

maxime-rainville (changes requested):
There's already caching built into AssetAdmin. That's why you don't get a GraphQL request when you navigate to a folder you've already visited. If we wanted a way to recall results of a previous page, we should try to interact with that GraphQL caching directly, rather than subvert it in AssetAdmin.

The fundamental problem is not caching. The problem is that the the Editor form is missing bits of data that it expects to be passed to it by AssetAdmin:
- Let's say you have selected a file that is on the second page and you saved your DataObject
- You refresh you CMS screen (your `fetchedFiles` cache will be empty at that point)
- You click the view icon on the UploadField
- The AssetAdmin modal popup and load the first page of results ... that page won't contain your selected image because it's on the second page. It's not in the cache either because you never accessed the second page of results.

I tried running your PR locally and that's exactly the behaviour I saw.

There's [detailed explanation of the problem and how I think it should be approached](https://github.com/silverstripe/silverstripe-asset-admin/issues/1136#issuecomment-726516706) on the issue.

PR #1222, opened 2021-07-05
FIX Always sort by folders first

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1108

Original PR for ReadFileConnection of https://github.com/silverstripe/silverstripe-asset-admin/pull/1134
Original PR with behat tests https://github.com/silverstripe/silverstripe-asset-admin/pull/1130

For the GraphQL 3, I tried a few different approaches of getting the sort to be done in a non-connection class, though in the end, because applySort() is a protected method and we need to add a non-standard sort, it's hard to do much else short of modifying the graphql module (which I really don't want to do)

The GraphQL 4 version has no unit test, as there are currently no unit tests for resolvers in general for GraphQL 4 in asset-admin.  The new functionality is still covered by the updated behat test

### DIFF
--- a/_legacy/GraphQL/ReadFileConnection.php
+++ b/_legacy/GraphQL/ReadFileConnection.php
@@ -1,0 +1,36 @@
+<?php
+namespace SilverStripe\AssetAdmin\GraphQL;
+
+use SilverStripe\Assets\Folder;
+use SilverStripe\GraphQL\Pagination\Connection;
+use SilverStripe\ORM\DataQuery;
+use SilverStripe\ORM\DB;
+use SilverStripe\ORM\SS_List;
+
+if (!class_exists(Connection::class)) {
+    return;
+}
+/**
+ * Connection that sorts by folders first
+ *
+ * @deprecated 1.8 Use silverstripe/graphql:^4 functionality.
+ */
+class ReadFileConnection extends Connection
+{
+    /**
+     * Always sort by folders before files
+     *
+     * @param SS_List $list
+     * @param array $sortBy
+     * @return SS_List
+     */
+    protected function applySort($list, $sortBy)
+    {
+        $className = DB::get_conn()->quoteString(Folder::class);
+        $field = "(CASE WHEN \"ClassName\"={$className} THEN 1 ELSE 0 END)";
+        $sortableFields = $this->getSortableFields();
+        $this->setSortableFields(array_merge([$field => $field], $sortableFields));
+        $updatedSortBy = array_merge([['field' => $field, 'direction' => 'DESC']], $sortBy);
+        return parent::applySort($list, $updatedSortBy);
+    }
+}
--- a/code/GraphQL/Resolvers/FolderTypeResolver.php
+++ b/code/GraphQL/Resolvers/FolderTypeResolver.php
@@ -12,7 +12,6 @@ use SilverStripe\GraphQL\QueryHandler\UserContextProvider;
 use SilverStripe\GraphQL\Schema\DataObject\FieldAccessor;
 use SilverStripe\GraphQL\Schema\Schema;
 use SilverStripe\ORM\DataList;
-use SilverStripe\ORM\DataQuery;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\Sortable;
 use SilverStripe\Versioned\Versioned;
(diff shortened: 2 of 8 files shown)


code/GraphQL/Resolvers/FolderTypeResolver.php
```diff
@@ -56,43 +55,6 @@ public static function resolveFolderChildren(
         $filter['parentId'] = $object->ID;
         $list = FileFilter::filterList($list, $filter);
 
-
-        // Ensure that we're looking at a subset of relevant data.
-        if (!isset($args['sort'])) {
-            // only show folders first if no manual ordering is set
-
-            $list = $list->alterDataQuery(static function (DataQuery $dataQuery) {
```
emteknetnz, line 64: No longer required as the folder/file sorting now happens in sortChildren

tests/php/GraphQL/FolderTypeCreatorTest.php
```diff
@@ -28,26 +28,6 @@ protected function setUp()
         }
     }
 
-    public function testItSortsChildrenOnTypeByDefault()
```
emteknetnz, line 31: This test was failing because of the default sort code removed from FolderTypeResolver (GraphQL 4)

From a functional point of view, with this PR if you go to asset-admin folder now always display before files, implying that sortChildren is always run.

This unit test is not required because there isn't a real world scenario where we run the resolver without sortChildren

maxime-rainville (commented):
Looking at it now, I'm thinking this might not be the best way of approaching this problem.

We're hacking our data layer to achieve a purely UX consideration. This could lead to weird results. For example, if you're trying to get file data out of the system, it will force you to sort you result with folder firs even if that's not what you are after.

I'm thinking the proper fix would be to have some sort of dynamic `isFolder` property and then update this bit to always sort by that first: https://github.com/silverstripe/silverstripe-asset-admin/blob/692697cf90b1bd395155cb82877d806400ff927b/client/src/state/files/readFilesQuery.js#L44-L46

emteknetnz:
> For example, if you're trying to get file data out of the system

Keep in mind this fix is only for asset-admin which is a UX component?.  If you're trying to get file data out of the system, you'll probably be writing a separate rest/graphql query?

> I'm thinking the proper fix would be to have some sort of dynamic isFolder property and then update this bit to always sort by that first

We've already tried that [here](https://github.com/silverstripe/silverstripe-asset-admin/pull/1130#issue-466485227) issue is that
a) dynamic implies you need to query the entire database table to work this out, and then sort the results
b) you could use ClassName column as your sort, but `Since Image extends File, i.e. "Image" is greater than "Folder" means that sorting by LastEdited, Title, etc, is now broken`

_legacy/GraphQL/ReadFileConnection.php
```diff
@@ -0,0 +1,32 @@
...
+use SilverStripe\GraphQL\Pagination\Connection;
+use SilverStripe\ORM\DataQuery;
+use SilverStripe\ORM\DB;
+use SilverStripe\ORM\SS_List;
+
+if (!class_exists(Connection::class)) {
+    return;
+}
+
+class ReadFileConnection extends Connection
```
maxime-rainville, line 14: This needs to be @deprecated and/or @internal.

maxime-rainville (changes requested):
I managed to get my GraphQL 3.5 to sort by a fake `isFolder` column with this https://github.com/creative-commoners/silverstripe-asset-admin/commit/04ff53630fe6b09f6d8f9e50ad79929ac6f0092b

It replaces some dodgy bits of logic with some slightly less dodgy bits of logic and doesn't require a new `ReadFileConnection`. Not sure it's worth spending more time on this thought.

Otherwise, what you have here works and is acceptable. Just need to make sure we don't support the `ReadFileConnection` class.

_legacy/GraphQL/ReadFileConnection.php
emteknetnz: Done

maxime-rainville (approved):
Looks good. Tested on GraphQL 3.5/4 and on installer/cwp-sink.

Technically we are adding an API to a patch release, but since we're deprecating it right away and this is a high impact bug, I think it's worth it.

Issue #1081, opened 2020-04-23
Can't download/delete image if image isn't on the current page


blueo:
To reproduce:

1. have at least 2 pages of assets
2. Select an image from the first page so it shows in the editor pane on the right
3. notice here, clicking the popover menu and selecting 'Download File' will work
4. advance to the next page using the arrows at the bottom of the asset view
5. attempt to download the file again - nothing appears to happen and an exception is thrown in the console (below)

the exception is (with dev files):
```
TypeError: this.props.file is undefined 2 Editor.js:130:4
    downloadFile Editor.js:130
    handleAction Editor.js:50
    handleAction self-hosted:844
    handleAction FormBuilder.js:191
    handleAction self-hosted:844
    handleClick FormAction.js:156
    handleClick self-hosted:844
    React 17
    bind_applyFunctionN self-hosted:1042
    dispatchEvent self-hosted:1005
    unstable_runWithPriority scheduler.development.js:697
    React 4
    dispatchDiscreteEvent self-hosted:948
    (Async: EventListener.handleEvent)
    addEventListener index.js:190
    React 20
    initReactRouter BootRoutes.js:111
    start BootRoutes.js:55
    _callee$/< index.js:54
    load buildInjectorContainer.js:78
    _callee$/< index.js:67
    (Async: setTimeout handler)
    _callee$ index.js:67
    Babel 9
    (Async: EventHandlerNonNull)
    js index.js:69
    Webpack 5

```

it seems file object is removed from the Editor props when the item is no longer in the gallery view.

tested on Silverstripe 4.5.1 with silverstripe/asset-admin 1.5.1

# Pull request
- [x] https://github.com/silverstripe/silverstripe-asset-admin/issues/1225

maxime-rainville:
Managed to replicate.

Looks like the problem is that when you swap page, the file you are currently editing looses its graphql context. So the editor doesn't know what the file ID is anymore. Looks like other action are broken as well, like Delete. Interestingly, Save/Publish work just fine.

I'm thinking, one fix here is probably to close the editor when you change page. There's precedent for that: when you change folder it will close the editor, so it would arguably make sense to do the same thing when you change page.

blueo:
it might be slightly unexpected to have that close when navigating a page/searching. I think my preference would be that it stays open. Could the editor keep its state independent of the gallery?

maxime-rainville:
[...]

PR #1225, opened 2021-07-16
Allow the Editor to fetch file details if the parent doesn't pass…

maxime-rainville:
Ready for review.

A couple edge cases to consider:
- Editing a File and a Folder through the File section
- Editing a File and a Folder through the WYSIWYG
- Editing a File and a Folder through the UploadField

# Parent issue
- fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/1136
- fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/1081 (Fix by accident)


client/src/boot/registerQueries.js
```diff
@@ -11,6 +12,7 @@ const registerQueries = () => {
   Injector.query.registerFragment('FileInterfaceFields', fileInterface);
   Injector.query.registerFragment('FileFields', file);
   Injector.query.register('ReadFilesQuery', isLegacy ? readFilesQueryLegacy : readFilesQuery);
+  Injector.query.register('ReadOneFileQuery', isLegacy ? readFilesQueryLegacy : readOneFileQuery);
```
maxime-rainville, line 15: Just have the GraphQL 4 version working for now ... need to do the GraphQL  v3 equivalent as well.

client/src/containers/Editor/Editor.js
```diff
@@ -360,6 +360,19 @@ function mapStateToProps({ assetAdmin: { gallery, modal } }) {
 
 export { Editor as Component };
 
+
+function magic(Component) {
+  return (props) => {
+    if (props.file) {
+      return <Component {...props} />;
+    } else {
+      const newProps = {...props, fileId: props.targetId};
```
maxime-rainville, line 369: When file is NOT provided by the parent, we need to fire a GraphQL request to retrieve a single file as defined by `props.targetId`.

client/src/containers/Editor/Editor.js
```diff
@@ -360,6 +360,19 @@ function mapStateToProps({ assetAdmin: { gallery, modal } }) {
 
 export { Editor as Component };
 
+
+function magic(Component) {
+  return (props) => {
+    if (props.file) {
+      return <Component {...props} />;
```
maxime-rainville, line 367: When `file` is provided by the parent, we do nothing.

client/src/state/files/readOneFileQuery.js
```diff
@@ -0,0 +1,69 @@
+import { hasFilters } from 'components/Search/Search';
+import { graphqlTemplates } from 'lib/Injector';
+
+const apolloConfig = {
```
maxime-rainville, line 4: This file mostly copies `readFilesQuery`, but removes all the fluff.

We're still using the same graphQL endpoint to retrieve our single file. Maybe we should create an actual readOneFile query server side. Not sure if that's advisable.

client/src/state/files/readOneFileQuery.js
```diff
@@ -0,0 +1,69 @@
...
+      data: {
+        error,
+        refetch,
+        readFiles,
+        loading: networkLoading,
+      },
+      ownProps: { actions },
+    }
+  ) {
+    const file = readFiles ? readFiles[0] : null;
```
maxime-rainville, line 27: Need to figure out what we do ... if anything ... if the file doesn't exist.

client/src/state/files/readOneFileQuery.js
```diff
@@ -0,0 +1,69 @@
...
+      error.graphQLErrors.map((graphQLError) => graphQLError.message);
+    return {
+      loading: networkLoading,
+      file,
+      graphQLErrors: errors,
+      actions: Object.assign({}, actions, {
+        files: Object.assign({}, actions.files, {
+          readFiles: refetch,
+        }),
+      }),
```
maxime-rainville, line 39: This bit is probably not needed.

client/src/state/files/readOneFileQuery.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  },
+  fragments: [
+    'FileInterfaceFields',
+    'FileFields',
+  ],
+  fields: [
+    '...FileInterfaceFields',
+    '...FileFields',
+  ],
```
maxime-rainville, line 65: We don't need all of this, but it could have value if we ever encounter a scenario where we need to retrieve a single file.

Maybe we should just get the bits that allows us to decide if we are working with a folder and have this query tighly coupled to the specific use case we need it for.

client/src/containers/Editor/Editor.js
maxime-rainville: There's a scenario where AssetAdmin is not done loading the first page of file results, so it can't pass the `file` to the Editor. 

When it's done, it provides the `file`, which causes the Editor to reload. I'm thinking either:
- AssetAdmin communicates us its loading state and we just wait until it's done fetching its initial list before we start loading the editor or,
- We bail on having AssetAdmin passing us the file info and always rely on our own GraphQL request all the time.

I'm leaning towards the second options. That would increase our number of request slightly, but it reduces the complexity of the component and the tie-in with AssetAdmin.

maxime-rainville (commented):
Could use some initial feedback before I get too far down this rabbit hole.

This thing mostly works, but needs a bit of cleaning

client/src/containers/Editor/Editor.js
maxime-rainville: Had a chat with @emteknetnz. We decided to break the link between AssetAdmin and Editor and always fire a GraphQL request to check if targetID is a folder or a file.

client/src/state/files/readOneFileQuery.js
maxime-rainville: @emteknetnz and I have decided to tightly couple the query to the editor use case. I'll rewrite this accordingly.

client/src/containers/Editor/Editor.js
```diff
@@ -323,12 +324,11 @@ class Editor extends Component {
 Editor.propTypes = {
   file: fileShape,
   className: PropTypes.string,
-  targetId: PropTypes.number.isRequired,
+  fileId: PropTypes.number.isRequired,
```
maxime-rainville, line 327: I've renamed this to `fileId`. This avoids us having to do gymnastic to change the prop from `targetId` to `fileID` when we pass it to the graphQL query and it just makes more sense since the target will always be a file ID any way.

client/src/state/files/_legacy/readOneFileQuery.js
```diff
@@ -0,0 +1,52 @@
...
+    return {
+      variables: {
+        rootFilter: {
+          id: fileId
+        },
+      },
+    };
+  },
+  props({ data: { error, readFiles, loading } }) {
+    const file = (readFiles && readFiles.nodes[0]) ? readFiles.nodes[0] : null;
```
maxime-rainville, line 14: That's the only difference between v3.5 and 4. In 4, results are on the top level of `readFiles`

client/src/state/files/readOneFileQuery.js
maxime-rainville: Decided not to do anything. If you try to access a file that doesn't exists, the editor will be barked anyway.

emteknetnz (approved):
Good

PR #1230, opened 2021-08-11
FIX Update unit test to use TestSession for session

emteknetnz:
Issue https://github.com/silverstripe/recipe-cms/issues/47

### DIFF
--- a/tests/php/Controller/AssetAdminTest.php
+++ b/tests/php/Controller/AssetAdminTest.php
@@ -40,8 +40,8 @@ class AssetAdminTest extends FunctionalTest
         parent::setUp();
 
         TestAssetStore::activate('AssetAdminTest');
-        $memberID = $this->logInWithPermission('ADMIN');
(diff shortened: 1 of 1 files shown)


tests/php/Controller/AssetAdminTest.php
```diff
@@ -13,6 +13,7 @@
 use SilverStripe\Control\Director;
 use SilverStripe\Control\Session;
 use SilverStripe\Dev\FunctionalTest;
+use SilverStripe\Dev\TestSession;
```
maxime-rainville, line 16: `TestSession` doesn't appear to be used?
```suggestion
```

maxime-rainville (changes requested):

tests/php/Controller/AssetAdminTest.php
emteknetnz: Removed

PR #1243, opened 2021-11-23
v1.8: Add missing TypeCreator import to FolderTypeCreator

chrispenny:
The import is currently missing, which means the `class_exists()` check is always failing, and the class is never available.

Screen Shot 2021-11-24 at 11 50 21 AM


maxime-rainville (approved):
We've tweaked the build upstream, which partially solved tho problem here.

However, it introduce a PHP8 build which will always fail on the 1.8 branch.

Any way good enough for me.

On a side note, I think I figured out why the build was green in the first place. https://github.com/silverstripe/silverstripe-framework/issues/9500

PR #1244, opened 2021-11-24
MNT Targte specific version of installer

maxime-rainville:
We seem to be installing the wrong version of framework on Travis.

Blocks https://github.com/silverstripe/silverstripe-asset-admin/pull/1243

### DIFF
--- a/_legacy/GraphQL/FolderTypeCreator.php
+++ b/_legacy/GraphQL/FolderTypeCreator.php
@@ -19,9 +19,6 @@ if (!class_exists(TypeCreator::class)) {
     return;
 }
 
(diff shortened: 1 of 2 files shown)


_legacy/GraphQL/FolderTypeCreator.php
```diff
@@ -19,6 +19,9 @@
     return;
 }
 
+$reflector = new \ReflectionClass(TypeCreator::class);
```
michalkleiner, line 22: remove?

maxime-rainville:
The problem was with the travis shared config ... not the asset-admin travis config.

PR #1249, opened 2022-01-16
FIX Use regular string instead of heredoc so textcollector runs

emteknetnz:
When running `vendor/bin/sake dev/tasks/i18nTextCollectorTask merge=1` the following exception was being triggered

`E_USER_NOTICE: Missing localisation default for key SilverStripe\AssetAdmin\Controller\AssetAdmin.LoadingTitleTip`

Converting the heredoc syntax to a regular string means text collector no longer throws the exception


michalkleiner (approved):

michalkleiner:
Self-merge on green

PR #1252, opened 2022-02-19
ENH Allow developers to decide if resampling is needed

leochenftw:
Sometimes, you just don't need the CMS to resample the images (e.g. when using S3 module + lambda resizing)

### DIFF
--- a/code/Controller/AssetAdmin.php
+++ b/code/Controller/AssetAdmin.php
@@ -1207,7 +1207,7 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
     public function generateThumbnails(File $file, $thumbnailLinks = false)
     {
         $links = [];
-        if (!$file->getIsImage()) {
+        if (!$file->getIsImage() || $file->config()->resample_images === false) {
             return $links;
         }
         $generator = $this->getThumbnailGenerator();
--- a/code/Helper/ImageThumbnailHelper.php
+++ b/code/Helper/ImageThumbnailHelper.php
@@ -140,6 +140,10 @@ class ImageThumbnailHelper
     {
         $generated = [];
 
+        if ($file->config()->resample_images === false) {
+            return $generated;
+        }
+
         $store = Injector::inst()->get(AssetStore::class);
         $assetAdmin = AssetAdmin::singleton();
         $generator = $assetAdmin->getThumbnailGenerator();
--- a/code/Model/ThumbnailGenerator.php
+++ b/code/Model/ThumbnailGenerator.php
@@ -98,7 +98,7 @@ class ThumbnailGenerator
      */
     public function generateThumbnail(AssetContainer $file, $width, $height)
     {
-        if (!$file->getIsImage() || !$file->exists()) {
+        if (!$file->exists() || !$file->getIsImage() || $file->config()->resample_images === false) {
             return null;
         }
 
@@ -111,7 +111,7 @@ class ThumbnailGenerator
         $method = $this->config()->get('method');
         return $file->$method($width, $height);
     }
-    
+
     /**
      * Generate "src" property for this thumbnail.
      * This can be either a url or base64 encoded data


GuySartorelli:
It would be great to see a companion PR in silverstripe/silverstripe-assets to add this config variable with a default value and PHPdoc, and a unit test added to this PR.

Looks like a useful optimisation for the use case you pointed out. 👍

leochenftw:
thanks, @GuySartorelli for your feedback. 

Ideally, the default config value should be added under https://github.com/silverstripe/silverstripe-assets, but, it will look confusing if I submit a PR with `no_resample` config variable along over there without the context.

So maybe SS could add that officially?

GuySartorelli:
> Ideally, the default config value should be added under https://github.com/silverstripe/silverstripe-assets, but, it will look confusing if I submit a PR with `no_resample` config variable alone over there without the context.

If you mention in that PR that is is for this 9ne then they'll be treated together.

leochenftw:
corresponding PR made on SS Assets module, at https://github.com/silverstripe/silverstripe-assets/pull/479

michalkleiner:
I've commented on the admin PR. We need to change the name around to reflect the default and then update this PR to reflect that. Happy to merge both when the changes are addressed.

leochenftw:
thanks, @michalkleiner . I've made the changes as you suggested.

Cheers

code/Controller/AssetAdmin.php
```diff
@@ -1207,7 +1207,7 @@ public function getMinimalistObjectFromData(File $file, $thumbnailLinks = true)
     public function generateThumbnails(File $file, $thumbnailLinks = false)
     {
         $links = [];
-        if (!$file->getIsImage()) {
+        if (!$file->getIsImage() || !$file->config()->resample_images) {
```
michalkleiner, line 1210:
```suggestion
        if (!$file->getIsImage() || $file->config()->resample_images === false) {
```

michalkleiner: With this change it will only disable the resampling when it's strictly `false`, otherwise when the config is missing or true it will work as it was before

code/Controller/AssetAdmin.php
```diff
@@ -1181,7 +1181,7 @@ public function getMinimalistObjectFromData(File $file, $thumbnailLinks = true)
         }
 
         /** @var File $file */
-        if ($file->getIsImage()) {
+        if ($file->getIsImage() && $file->config()->resample_images) {
```
michalkleiner, line 1184: Do we need this change here?
If we prevent returning links to resampled images from the `generateThumbnails` method, we don't need to do anything here, right?

leochenftw: I think we do, because the code inside of the if block still tries to `Save dimensions` into the object that gets returned. Correct me if my didn't see the full picture of this function's purpose

michalkleiner: Yeah but it would take the original dimensions from the file, wouldn't it?
Is your issue actually about having the images resampling or the object having the width and height attributes which are two different things?

michalkleiner: A unit test would help here.

leochenftw: > 

I don't have issues with taking the original width and height - just thought the dimensions might not be necessary. I will take your point and make the change then

code/Helper/ImageThumbnailHelper.php
```diff
@@ -140,6 +140,10 @@ protected function generateThumbnails(File $file)
     {
         $generated = [];
 
+        if (!$file->config()->resample_images) {
```
GuySartorelli, line 143: In AssetAdmin you're explicitly checking for `false`, but here you're just checking loose falseness. Is there a reason to check for exactly false there but not here?

GuySartorelli: Same question applies for the check in ThumbnailGenerator.

leochenftw: thanks @GuySartorelli for pointing that out - I've updated the code

code/Model/ThumbnailGenerator.php
```diff
@@ -98,7 +98,7 @@ public function generateThumbnailLink(AssetContainer $file, $width, $height, $gr
      */
     public function generateThumbnail(AssetContainer $file, $width, $height)
     {
-        if (!$file->getIsImage() || !$file->exists()) {
+        if (!$file->getIsImage() || !$file->exists() || $file->config()->resample_images === false) {
```
michalkleiner, line 101:
```suggestion
        if (!$file->exists() || !$file->getIsImage() || $file->config()->resample_images === false) {
```

michalkleiner:
Looking good now @leochenftw. Can you please swap the order of the conditions on one of the touched rows and then squash your commits into one "ENH Allow developers to decide if resampling is needed". I'll be happy to merge it after that.

leochenftw:
thanks @michalkleiner - is it possible for you to do the squash upon merge? I recently returned the laptop that I was doing the dev work on, so it will be easier to do it that way

michalkleiner:
Thanks for helping with the squash, Dan, I need to look myself into how to push to fork's PR branches.

michalkleiner (approved):

dhensby:
@michalkleiner as long as the contributor has allowed maintainers to push to their PRs it's as simple as `git p [-f] https://github.com/their/repo HEAD:their-branch`

PR #1254, opened 2022-02-28
NEW Use embed/embed v4

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10243

Requires https://github.com/silverstripe/silverstripe-framework/pull/10244

### DIFF
--- a/code/Exceptions/InvalidRemoteUrlException.php
+++ b/code/Exceptions/InvalidRemoteUrlException.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace SilverStripe\AssetAdmin\Exceptions;
+
+use Exception;
+
+/**
+ * This exception is thrown by RemoteFileFormFactory when failing to fetch an
+ * embeddable remote resource because it's either an invalid URL or because
+ * the resource is invalid
+ */
+class InvalidRemoteUrlException extends Exception
+{
+}
--- a/code/Forms/RemoteFileFormFactory.php
+++ b/code/Forms/RemoteFileFormFactory.php
@@ -2,8 +2,8 @@
 
 namespace SilverStripe\AssetAdmin\Forms;
 
-use Embed\Exceptions\InvalidUrlException;
 use InvalidArgumentException;
+use SilverStripe\AssetAdmin\Exceptions\InvalidRemoteUrlException;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\Config\Configurable;
@@ -164,7 +164,7 @@ class RemoteFileFormFactory implements FormFactory
     /**
      * @param string $url
      * @return bool
-     * @throws InvalidUrlException
+     * @throws InvalidRemoteUrlException
      */
     protected function validateUrl($url)
     {
@@ -181,12 +181,12 @@ class RemoteFileFormFactory implements FormFactory
      *
      * @param Embeddable $embed
      * @return bool
-     * @throws InvalidUrlException
+     * @throws InvalidRemoteUrlException
      */
     protected function validateEmbed(Embeddable $embed)
     {
         if (!$embed->validate()) {
-            throw new InvalidUrlException(_t(
+            throw new InvalidRemoteUrlException(_t(
                 __CLASS__.'.ERROR_EMBED',
                 'There is currently no embeddable media available from this URL'
             ));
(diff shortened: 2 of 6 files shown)


code/Exceptions/InvalidUrlException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace SilverStripe\AssetAdmin\Exceptions;
+
+use Exception;
+
+class InvalidUrlException extends Exception
```
maxime-rainville, line 7: Why are we creating this exception?

code/Forms/RemoteFileFormFactory.php
```diff
@@ -2,8 +2,8 @@
 
 namespace SilverStripe\AssetAdmin\Forms;
 
-use Embed\Exceptions\InvalidUrlException;
 use InvalidArgumentException;
+use SilverStripe\AssetAdmin\Exceptions\InvalidUrlException;
```
maxime-rainville, line 6: Seems like we should be swapping this for `NetworkException` or `RequestException`?

code/Forms/RemoteFileFormFactory.php
```diff
@@ -2,8 +2,8 @@
```
maxime-rainville, line 2: Is this thing use when we hotlink to a file?

maxime-rainville (changes requested):

code/Forms/RemoteFileFormFactory.php
emteknetnz: Yes?

code/Forms/RemoteFileFormFactory.php
emteknetnz: The RequestException implementation in embed/embed doesn't actually allow for custom messages, despite it being int he constructor.  We have a bunch of existing _t() data that we should retain.

https://github.com/oscarotero/Embed/blob/master/src/Http/RequestException.php#L16

code/Exceptions/InvalidUrlException.php
emteknetnz: Commented in reply below

code/Forms/RemoteFileFormFactory.php
emteknetnz: Also the RequestException has 3 params, int $code and RequestInterface $request

We could make up a code, but we have no way to get $request from many contexts, eg

**RemoteFileFormFactory.php**

```php
    /**
     * Checks if the embed generated is not just a link
     *
     * @param Embeddable $embed
     * @return bool
     * @throws InvalidUrlException
     */
    protected function validateEmbed(Embeddable $embed)
    {
        if (!$embed->validate()) {
            throw new InvalidUrlException(_t(
                __CLASS__.'.ERROR_EMBED',
                'There is currently no embeddable media available from this URL'
            ));
        }
        return true;
    }
```

maxime-rainville: Cool ... that makes sense.

maxime-rainville (changes requested):
`InvalidUrlException` sound too generic for something that is very `oEmbed` specific. At the very least, there should be a comment on the class to explain its purpose.

I'm wondering if it wouldn't make more sense to have this exception in framework instead?

emteknetnz:
Have renamed to `InvalidRemoteUrlException`

Note there's something of an unavoidable (semver breaking?) API change in that the Exception thrown has changed.  It's unavoidable because the old Exception came from embed/embed v3, which no longer exists.

code/Exceptions/InvalidRemoteUrlException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace SilverStripe\AssetAdmin\Exceptions;
+
+use Exception;
+
+class InvalidRemoteUrlException extends Exception
```
maxime-rainville, line 7: Can we include a comment to explains what this is for?

maxime-rainville (changes requested):

code/Exceptions/InvalidRemoteUrlException.php
emteknetnz: Added

maxime-rainville (approved):

maxime-rainville:
Build is failing because it's still trying to install an older version of frameworktest.

I took the liberty to bump it up

maxime-rainville:
Actually, the framewortest version got fixed in https://github.com/silverstripe/silverstripe-asset-admin/pull/1259 already. I removed my extra commit, merged up the other fixed and rebased.

PR #1256, opened 2022-03-09
BUG Update asset-admin to use the correct allowed extension list for dropzone

maxime-rainville:
Asset-admin sends a list of allowed file extension to the front end for its drop zone settings. The problem is that this list is read directly from the `File::$allowed_extensions` rather than use the more robust `File::getAllowedExtension()`.

This means that if you use this kind of YML to [remove an extension from the allowed extension list](https://docs.silverstripe.org/en/4/developer_guides/files/allowed_file_types/#file-extensions-validation), the front end will still accept it:
```yml
SilverStripe\Assets\File:
  allowed_extensions:
    zip: false
```

The back end will still block the file, but you waste time uploading it and get a less useful error message.

### DIFF
--- a/tests/php/Controller/AssetAdminTest.php
+++ b/tests/php/Controller/AssetAdminTest.php
@@ -451,7 +451,6 @@ class AssetAdminTest extends FunctionalTest
 
             $config = $assetAdmin->getClientConfig();
             $acceptedFiles = $config['dropzoneOptions']['acceptedFiles'];
-            var_dump($acceptedFiles);
             $this->assertStringContainsString(
                 '.boom',
                 $acceptedFiles,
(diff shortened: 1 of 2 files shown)


tests/php/Controller/AssetAdminTest.php
```diff
@@ -441,4 +442,31 @@ public function testGetObjectFromDataFileWithFolder()
...
+
+    public function testGetClientConfigExtensions()
+    {
+        Config::withConfig(function () {
+            $assetAdmin = AssetAdmin::singleton();
+            Config::modify()->set(File::class, 'allowed_extensions', ['boom']);
+
+            $config = $assetAdmin->getClientConfig();
+            $acceptedFiles = $config['dropzoneOptions']['acceptedFiles'];
+            var_dump($acceptedFiles);
```
michalkleiner, line 454:
```suggestion
```

michalkleiner (approved):
Merge on green provided the var_dump is removed

PR #1261, opened 2022-03-22
Force react component to always update when field is rendered

RVXD:
Force react component to always update when we render the field, by making value.Files and data.files mismatch in the state.

### DIFF
--- a/code/Forms/UploadField.php
+++ b/code/Forms/UploadField.php
@@ -199,6 +199,9 @@ class UploadField extends FormField implements FileHandleField
         $state = parent::getSchemaStateDefaults();
         $state['data']['files'] = $this->getEncodedItems();
         $state['value'] = $this->Value() ?: ['Files' => []];
+        // force react component to always update when field is rendered,
+        // by making value.Files and data.files mismatch in the state
+        $state['value']['Files'][] = -1;
         return $state;
     }
 


dhensby:
The tests are failing because you've updated the attributes array to have a new value and the tests are not expecting it. The test assertions need to be updated

kinglozzer:
I’m not sure this is an appropriate fix to be honest, surely it’s just hiding the real problem? I understand the bug is frustrating (our clients ask us about this constantly too!) but I worry that this is just kicking the can down the road

michalkleiner:
If it was just adding a new value that doesn't influence anything else and just causes a refresh, I'd think about it, but from the failing tests it seems the added value is mixed with actual real values. I've expressed my view on Slack already, though in general, politely said, I agree with Loz that this is not the right fix. The original issue has a long discussion thread, likely for a reason. Happy to be proven wrong though.
At the moment I can't offer much time to deep dive into this, though I appreciate the pain of dealing with the stale state of the component.

GuySartorelli:
There is definitely a lot of discussion in both the original issue and at least one other PR about a bunch of interconnected pieces that need to be considered, as well as two other related issues that seem to have the same underlying cause.
I agree that this PR doesn't really address the underlying issue - it's fine as a workaround but it's not a fix.

That said, thank you for taking a stab at it @RVXD. Your efforts are appreciated, I just don't think that the proposed fix in this case is robust enough to merge in as it stands.

dhensby:
OK, that's 3 votes against this approach from core members/contributors so I'm going to close it for the time being

RVXD:
Thanks for looking into this. Clients complain about this often. For a lot of users the whole versioned files concept is complex and doesn't add any value.

PR #1264, opened 2022-04-04
ENH PHP 8.1 compatibility

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250

### DIFF
--- a/tests/behat/src/FixtureContext.php
+++ b/tests/behat/src/FixtureContext.php
@@ -164,8 +164,8 @@ class FixtureContext extends BaseFixtureContext
         // Get path
         $filesPath = $this->getFilesPath();
         if ($filesPath) {
-            $fullPath = rtrim(realpath($filesPath), DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$path;
-            if (is_file($fullPath)) {
+            $fullPath = rtrim(realpath($filesPath ?? '') ?? '', DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$path;
+            if (is_file($fullPath ?? '')) {
                 $path = $fullPath;
             }
         }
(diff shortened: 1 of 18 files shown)


tests/behat/src/FixtureContext.php
```diff
@@ -164,8 +164,8 @@ public function iAttachTheFileToDropzone($path, $name)
         // Get path
         $filesPath = $this->getFilesPath();
         if ($filesPath) {
-            $fullPath = rtrim(realpath($filesPath), DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$path;
-            if (is_file($fullPath)) {
+            $fullPath = rtrim((string) realpath((string) $filesPath), DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$path;
```
maxime-rainville, line 167: Is there any chance that `realpath` would not return a string here?

```suggestion
            $fullPath = rtrim(realpath((string) $filesPath), DIRECTORY_SEPARATOR).DIRECTORY_SEPARATOR.$path;
```

emteknetnz: `string|false`

https://www.php.net/manual/en/function.realpath.php

The code in the PR is automatically generated.  If a native function has a single return type there won't be a cast/ternary, if there's more than one type, such as in this instance, it will be cast/ternary

GuySartorelli (approved):
For posterity: See discussion in silverstripe/silverstripe-admin#1294 about the general approach. tl;dr:
> the purposes of this PR is to avoid passing null (specifically null) from built-in functions that as of php 8.1 will throw deprecation warnings if null is passed in. The approach is necessarily heavy-handed, and while there are many situations where it isn't needed, it would be prohibitively laborious to find all of those situations by hand.

Looks good to me

GuySartorelli:
There's a question on a couple of the other PRs about `file_put_contents()` and `file_get_contents()` so I'll hold off merging this until that's been answered.

Issue #1248, opened 2022-01-10
Error message not shown if replacing file with too big file


t3hn0:
Tested on SS 4.8.0 and on 4.9.0

If I try to upload file that's too big I get error message "Filesize is too large, maximum 25 MB allowed" which is ok.

But if I try to replace existing file with a file (which is too big) Assets Admin shows me success message "Upload successful, the file will be replaced when you Save." although I can see in browsers network tab JSON response was: "{"status":"error","errors":[{"type":"error","code":400,"value":"Filesize is too large, maximum 25 MB allowed"}]}".

Assets Admin shouldn't show success message if upload was not successful. It should display error.

### ACs
- [x] File too large error message shows when attempting to replace a file

### PR 
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1274

GuySartorelli:
PR merged. Closing.

PR #1273, opened 2022-05-30
FIX Show correct error message instead of successful message if file …

sabina-talipova:
### Description
Fixed an issue by adding "complete" event. After the file has been added, wait until the file status will be changed to "successful" or "error", only then update the content and show a message depending on what result will be received from the server.
New test cases were added.

### Parent issue

#1248


client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -302,7 +307,13 @@ class PreviewImageField extends Component {
...
         </div>
       );
-    } else if (progress === 100) {
+    } else if (errors) {
+      messageBox = (
+        <div className={`preview-image-field__message preview-image-field__message--${errors.type}`}>
+          {errors.value}
+        </div>
+      );
+    } else if (progress === 100 && !errors) {
```
GuySartorelli, line 316:
```suggestion
    } else if (progress === 100) {
```
We don't need to check again if there are errors since this is an `else if` after we already checked that.

tests/behat/src/FixtureContext.php
```diff
@@ -508,4 +509,31 @@ public function iScrollTheEditorDetailsPanelToTheTop()
         $script = "document.querySelector('.editor__details fieldset').scrollTo(0, 0);";
         $this->getMainContext()->getSession()->executeScript($script);
     }
+
+    /**
+     * Example: Given a maximum file size is 5k
+     *
+     * @Given /^a maximum file size is "([^"]+)"$/
```
GuySartorelli, line 516:
```suggestion
     * @Given /^the maximum file size is "([^"]+)"$/
```

tests/behat/features/replace-file.feature
```diff
@@ -45,3 +45,18 @@ Feature: Replace a file with a new file
     When I press the "Save" button
     Then the "Name" field should contain "document-v2.pdf"
       And I should not see a ".preview-image-field__message--success" element
+
+  @javascript
+  Scenario: I upload an image file exceeding maximum file size to replace a file
+    Given a maximum file size is "5k"
```
GuySartorelli, line 51:
```suggestion
    Given the maximum file size is "5k"
```

tests/behat/src/FixtureContext.php
```diff
@@ -508,4 +509,31 @@ public function iScrollTheEditorDetailsPanelToTheTop()
...
+     */
+    public function stepCreateMaximumFileSizeStep($size): void
+    {
+        $uploadValidatorClass = Upload_Validator::class;
+        $config = <<<YAML
+---
+name: fileallowedsize
+---
+
+$uploadValidatorClass:
```
GuySartorelli, line 527:
```suggestion
SilverStripe\Assets\Upload_Validator:
```
This isn't dynamic - it will never change. There's no need to inject it into the string like that.
You _might_ need to escape the backslashes.

tests/behat/src/FixtureContext.php
```diff
@@ -508,4 +509,31 @@ public function iScrollTheEditorDetailsPanelToTheTop()
...
+
+    /**
+     * Example: Given a maximum file size is 5k
+     *
+     * @Given /^a maximum file size is "([^"]+)"$/
+     * @param string $size Max file size
+     */
+    public function stepCreateMaximumFileSizeStep($size): void
+    {
+        $uploadValidatorClass = Upload_Validator::class;
```
GuySartorelli, line 521:
```suggestion
```

tests/behat/src/FixtureContext.php
```diff
@@ -7,6 +7,7 @@
 use Page;
 use PHPUnit\Framework\Assert;
 use SilverStripe\Assets\Image;
+use SilverStripe\Assets\Upload_Validator;
```
GuySartorelli, line 10:
```suggestion
```

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -219,6 +219,10 @@ class PreviewImageField extends Component {
    * @param {object} progress
    */
   handleUploadProgress(file, progress) {
+    if (file && file.status === 'uploading') {
+      setTimeout(() => this.handleUploadProgress(file, progress), 1000);
+      return;
+    }
```
GuySartorelli, line 225:
```suggestion
```
This part isn't needed - and in fact stops this function from performing its intended function. For very large files (where large files are allowed), this function provides the current upload progress to a progress bar.

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -302,7 +307,13 @@ class PreviewImageField extends Component {
           {message.value}
         </div>
       );
-    } else if (progress === 100) {
+    } else if (errors) {
```
GuySartorelli, line 310: In what situation is there a `message` value? Would it make sense for `if (errors)` to come first?

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -302,7 +307,13 @@ class PreviewImageField extends Component {
...
+      messageBox = (
+        <div className={`preview-image-field__message preview-image-field__message--${errors.type}`}>
+          {errors.value}
```
GuySartorelli, line 313:
```suggestion
    } else if (errors) {
    const errorMessage = errors.value ? errors.value : i18n._t('AssetAdmin.DROPZONE_RESPONSE_ERROR', 'Server responded with an error.');
      messageBox = (
        <div className={`preview-image-field__message preview-image-field__message--${errors.type}`}>
          {errorMessage}
```
Sometimes there is no error message in the response, for example if you upload a file that is within the size limits in yml config, but too large for php upload limits.

tests/behat/features/replace-file.feature
```diff
@@ -45,3 +45,18 @@ Feature: Replace a file with a new file
...
+      And I select the file named "folder1" in the gallery
+      And I click on the file named "file1" in the gallery
+    When I attach the file "file1.jpg" to dropzone "PreviewImage"
+    Then I should see a ".preview-image-field__message-button" element
+      And I should see a ".preview-image-field__message--success" element
+      And the "Name" field should contain "file1.jpg"
+    When I attach the file "file3.jpg" to dropzone "PreviewImage"
+      And I wait for 2 second
+    Then I should see a ".preview-image-field__message--error" element
+      And the "Name" field should contain "file1.jpg"
```
GuySartorelli, line 62: Check for the error message as well.

tests/behat/src/FixtureContext.php
```diff
@@ -508,4 +509,31 @@ public function iScrollTheEditorDetailsPanelToTheTop()
...
+        $uploadValidatorClass = Upload_Validator::class;
+        $config = <<<YAML
+---
+name: fileallowedsize
+---
+
+$uploadValidatorClass:
+    default_max_file_size:
+        '*': $size
+YAML;
```
GuySartorelli, line 530:
```suggestion
        ---
        name: fileallowedsize
        ---
        $uploadValidatorClass:
            default_max_file_size:
                '*': $size
        YAML;
```
Heredocs and nowdocs don't need to be aligned to the very left side anymore - we can keep them indented with the rest of the code.

GuySartorelli (changes requested):
Works well locally, a few things here to tidy up.

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -296,13 +309,19 @@ class PreviewImageField extends Component {
     ) : null;
     let messageBox = null;
 
-    if (message) {
+    if (errors || status === 'error') {
+      const errorMessage = errors && errors.value
+      ? errors.value
+      : i18n._t('AssetAdmin.DROPZONE_RESPONSE_ERROR', 'Server responded with an error.');
```
GuySartorelli, line 315:
```suggestion
        ? errors.value
        : i18n._t('AssetAdmin.DROPZONE_RESPONSE_ERROR', 'Server responded with an error.');
```
Indent these to make it easier to scan-read - they're part of the same const assignment as the line above.

GuySartorelli (changes requested):
There are a few small formatting changes to make plus some questions for things I don't quite understand.

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -296,13 +309,19 @@ class PreviewImageField extends Component {
...
+
       messageBox = (
-        <div className={`preview-image-field__message preview-image-field__message--${message.type}`}>
-          {message.value}
+        <div className={`preview-image-field__message preview-image-field__message--${errorType}`}>
+          {errorMessage}
         </div>
       );
-    } else if (progress === 100) {
+    } else if (progress === 100 && status === 'success') {
```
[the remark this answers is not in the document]
sabina-talipova: If there is progress less then 100, but more then 0, we process it there https://github.com/silverstripe/silverstripe-asset-admin/blob/b4dcd712f5d9bd0603f2d8e4e48a03d0a297e678/client/src/components/PreviewImageField/PreviewImageField.js#L305
and render it there https://github.com/silverstripe/silverstripe-asset-admin/blob/b4dcd712f5d9bd0603f2d8e4e48a03d0a297e678/client/src/components/PreviewImageField/PreviewImageField.js#L351

In the current code we are not processing any other situations. And how I can see in back end we don't have any other response than 4XX error responses.

I'm not sure that it's possible situation when status to be 'success' but progress to be something other than 100. Since  progress is computed by Dropzone.js and "success" status is the result of response from the server. It means if this situation is happining we could rise defect for Dropzone. 

Any way, we cannot provide the users with correct message in this situation. I'll add else condition for this case.

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -318,6 +337,12 @@ class PreviewImageField extends Component {
           )}
         </div>
       );
+    } else if (message) {
```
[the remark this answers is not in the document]
sabina-talipova: I suppose it's a redundant value. I think it was created for error messages initially. I couldn't find whence we send this value to Preview Image Field. I'll do more tests and if I couldn't find the code where we create this message I'll add new ticket to refactoring this part.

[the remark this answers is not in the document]
GuySartorelli: If you can't find it, please move this to be _before_ the success check (which is where it was previously), but _after_ the error check (error messages should take precedence imo)

client/src/components/PreviewImageField/PreviewImageField.js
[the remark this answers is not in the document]
sabina-talipova: I haven't add additional ELSE condition, since it works up when we just load preview already existing file.

GuySartorelli (changes requested):
Looks good.

Sorry that I only just noticed, but this should be targetting the `1.10` branch (since 1.11 isn't out yet, and this bug fix doesn't include any API changes)

sabina-talipova:
WRONG BRANCH

PR #1293, opened 2022-09-19
FIX Update plugins to work with tinymce 6

ChloeSartorelli:
Update plugins to with with tinymce6

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10369


ChloeSartorelli:
Note: The old behat selection logic stopped working - it's not really clear why, but the new logic is much more vanilla (doesn't rely on anything from the editor other than grabbing the internal document) so should be even more reliable than the old implementation,

ChloeSartorelli:
Needs https://github.com/silverstripe/silverstripe-framework/pull/10508 and https://github.com/silverstripe/silverstripe-admin/pull/1367 merged to go green, but tests are passing locally.

sabina-talipova (approved):
The branch will be merged after merging of [silverstripe-admin](https://github.com/silverstripe/silverstripe-admin/pull/1367) and [silverstripe-framework](https://github.com/silverstripe/silverstripe-framework/pull/10508) new changes.

PR #1299, opened 2022-10-07
graphQL: changed size attribute of File to Float

sirtoobii:
Resolves #889 

This bug needs to be addressed since its effect on the file admin is potentially devastating: In our case we migrated an SS3 installation to SS4 and they had files larger than the `2147483647` 32-bit integer size which basically rendered the whole file admin unusable.

As long there is no support for 64-bit integers, this could be feasible workaround.

An other approach would be to define a custom scalar type - is this supported by silverstripe's graphQL implementation?

### DIFF
--- a/_graphql/types/File.yml
+++ b/_graphql/types/File.yml
@@ -17,7 +17,7 @@ File:
     name: String
     filename: String
     extension: String
-    size: Int
+    size: Float
     url: String
     thumbnail: String
     smallThumbnail: String


ChloeSartorelli:
Hi there,
Thank you for submitting this pull request.

I don't think this is a good approach to resolving the issue - the type should not be a float, it should be an int. An appropriate solution would be to implement a larger int type (which you have also referred to in https://github.com/silverstripe/silverstripe-graphql/issues/501).

I'm going to close this PR but if you want to give a go at implementing a proper 64bit int type I'd be happy to look at a PR which solves it that way.

PR #1300, opened 2022-10-10
API Strongly-type action method signatures

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10466

### DIFF
--- a/code/Controller/AssetAdmin.php
+++ b/code/Controller/AssetAdmin.php
@@ -525,16 +525,14 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
 
     /**
      * Redirects 3.x style detail links to new 4.x style routing.
-     *
-     * @param HTTPRequest $request
      */
-    public function legacyRedirectForEditView($request)
+    public function legacyRedirectForEditView(HTTPRequest $request): HTTPResponse
     {
         $fileID = $request->param('FileID');
         /** @var File $file */
         $file = File::get()->byID($fileID);
         $link = $this->getFileEditLink($file) ?: $this->Link();
-        $this->redirect($link);
+        return $this->redirect($link);
     }
 
     /**
@@ -890,13 +888,8 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
 
     /**
      * Gets a JSON schema representing the current edit form.
-     *
-     * WARNING: Experimental API.
-     *
-     * @param HTTPRequest $request
-     * @return HTTPResponse
      */
-    public function schema($request)
+    public function schema(HTTPRequest $request): HTTPResponse
     {
         $formName = $request->param('FormName');
         if ($formName !== 'fileHistoryForm') {
@@ -950,12 +943,7 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
         return $form;
     }
 
-    /**
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
-     */
-    public function createfolder($data, $form)
+    public function createfolder(array $data, Form $form): HTTPResponse
     {
         $parentID = isset($data['ParentID']) ? intval($data['ParentID']) : 0;
         $data['Parent'] = null;
@@ -991,39 +979,23 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
         return $this->getSchemaResponse($schemaId, $createForm, null, $schemaData);
     }
 
-    /**
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
-     */
-    public function save($data, $form)
+    public function save(array $data, Form $form): HTTPResponse
     {
         return $this->saveOrPublish($data, $form, false);
     }
 
-    /**
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
-     */
-    public function publish($data, $form)
+    public function publish(array $data, Form $form): HTTPResponse
     {
         return $this->saveOrPublish($data, $form, true);
     }
 
     /**
-     * Update thisrecord
-     *
-     * @param array $data
-     * @param Form $form
-     * @param bool $doPublish
-     * @return HTTPResponse
+     * Update this record
      */
-    protected function saveOrPublish($data, $form, $doPublish = false)
+    protected function saveOrPublish(array $data, Form $form, bool $doPublish = false): HTTPResponse
     {
         if (!isset($data['ID']) || !is_numeric($data['ID'])) {
             $this->jsonError(400);
-            return null;
         }
 
         $id = (int) $data['ID'];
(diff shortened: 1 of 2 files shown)


code/Controller/AssetAdmin.php
```diff
@@ -525,10 +525,8 @@ public function apiHistory(HTTPRequest $request)
 
     /**
      * Redirects 3.x style detail links to new 4.x style routing.
-     *
-     * @param HTTPRequest $request
      */
-    public function legacyRedirectForEditView($request)
+    public function legacyRedirectForEditView(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 529: This method isn't actually returning the redirect

ChloeSartorelli: But also this method should probably be removed in cms5 anyway

code/Controller/AssetAdmin.php
```diff
@@ -1076,7 +1054,7 @@ protected function saveOrPublish($data, $form, $doPublish = false)
         return $this->getRecordUpdatedResponse($record, $form);
     }
 
-    public function unpublish($data, $form)
+    public function unpublish(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 1049: This method has some `return null;` statements

code/Controller/AssetAdmin.php
```diff
@@ -991,22 +979,12 @@ public function createfolder($data, $form)
...
 
-    /**
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
-     */
-    public function publish($data, $form)
+    public function publish(array $data, Form $form): HTTPResponse
     {
         return $this->saveOrPublish($data, $form, true);
```
ChloeSartorelli, line 989: Also need to make sure that method has the typehints

code/Controller/AssetAdmin.php
```diff
@@ -1076,7 +1054,7 @@ protected function saveOrPublish($data, $form, $doPublish = false)
         return $this->getRecordUpdatedResponse($record, $form);
```
ChloeSartorelli, line 1046: That method also needs the typehints

code/Controller/AssetAdmin.php
```diff
@@ -1237,12 +1215,8 @@ public function generateThumbnails(File $file, $thumbnailLinks = false)
 
     /**
      * Action handler for adding pages to a campaign
-     *
-     * @param array $data
-     * @param Form $form
-     * @return DBHTMLText|HTTPResponse
      */
-    public function addtocampaign($data, $form)
+    public function addtocampaign(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 1208: This method has some `return null;` statements

ChloeSartorelli (changes requested):

code/Controller/AssetAdmin.php
emteknetnz: Updated to return the redirect

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: The strong types in the method signature will type hint in IDE's

code/Controller/AssetAdmin.php
emteknetnz: `getRecordUpdatedResponse()` already has them in its phpdoc

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
ChloeSartorelli: That's my point. `saveOrPublish()` doesn't have strong types in the method signature. Sorry, I should have more clearly indicated what method "that method" is.

code/Controller/AssetAdmin.php
ChloeSartorelli: Right, but it could return anything. Which means this could return anything. Which means `publish()` and `save()` might throw errors.

code/Controller/AssetAdmin.php
```diff
@@ -1237,24 +1212,17 @@ public function generateThumbnails(File $file, $thumbnailLinks = false)
...
 
         $handler = AddToCampaignHandler::create($this, $record, 'addToCampaignForm');
         $results = $handler->addToCampaign($record, $data);
-        if (!isset($results)) {
-            return null;
-        }
 
         // Send extra "message" data with schema response
-        $extraData = ['message' => $results];
+        $extraData = ['message' => $results->getBody()];
```
ChloeSartorelli: For anyone looking at this in the future and wondering why this change was made, it reflects the change in https://github.com/silverstripe/silverstripe-campaign-admin/pull/244/files#diff-005db379d07dad4da0ae604f8201249c122b800056b951f9b64b05ef977b7fdcR336

code/Controller/AssetAdmin.php
emteknetnz: Have added strong typing to saveOrPublish() and removed the unnecessary `return null` underneath `$this->jsonError()` which throws exceptions

code/Controller/AssetAdmin.php
emteknetnz: For `getRecordUpdatedResponse()` I have:
- added strong typing to the `$record` param - it's already strong typed in the subsequent call to `getObjectFromData()`
- retaining dynamic typing with phpdoc for the `$form` param - it's also dynamic typed in the subsequent call to `getSchemaResponse()`

Note: `getRecordUpdatedResponse()` already had a `:HTTPResponse` strong return type

ChloeSartorelli (approved):
Don't merge until other PRs are ready to avoid breaking CI

code/Controller/AssetAdmin.php
```diff
@@ -1237,24 +1204,17 @@ public function generateThumbnails(File $file, $thumbnailLinks = false)
...
+    public function addtocampaign(array $data, Form $form): HTTPResponse
     {
         $id = $data['ID'];
         $record = File::get()->byID($id);
 
         $handler = AddToCampaignHandler::create($this, $record, 'addToCampaignForm');
         $results = $handler->addToCampaign($record, $data);
-        if (!isset($results)) {
-            return null;
-        }
```
ChloeSartorelli: I think this needs the same treatment as https://github.com/silverstripe/silverstripe-cms/pull/2788/files#r997551623 (`isset()` would have returned false for null, so it's effectively the same check as being done in cms)

ChloeSartorelli (changes requested):

code/Controller/AssetAdmin.php
emteknetnz: Updated

ChloeSartorelli (approved):

PR #1302, opened 2022-10-18
API Update deprecations

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10531

### DIFF
--- a/_legacy/GraphQL/CreateFileMutationCreator.php
+++ b/_legacy/GraphQL/CreateFileMutationCreator.php
@@ -1,6 +1,7 @@
 <?php
 namespace SilverStripe\AssetAdmin\GraphQL;
 
+use SilverStripe\Dev\Deprecation;
 use GraphQL\Type\Definition\ResolveInfo;
 use SilverStripe\Assets\File;
 use SilverStripe\Assets\Folder;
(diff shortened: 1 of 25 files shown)


ChloeSartorelli (changes requested):
Version numbers are incorrect - use `1` major version numbering instead. This will also need to be applied to any other deprecation notices in this module.

_legacy/GraphQL/CreateFileMutationCreator.php
```diff
@@ -15,7 +16,7 @@
 }
 /**
  * @todo Allow file upload (https://github.com/silverstripe/silverstripe-graphql/issues/19)
- * @deprecated 4.8..5.0 Use silverstripe/graphql:^4 functionality.
+ * @deprecated 1.8.0 Use _graphql directory functionality instead
```
ChloeSartorelli: The original message was more correct - the thing to use is the latest graphql, the directory is irrelevant.
The same message will need to be changed in a couple of other classes below as well.

ChloeSartorelli (changes requested):

_legacy/GraphQL/CreateFileMutationCreator.php
emteknetnz: Updated

ChloeSartorelli (approved):
MOG

PR #1304, opened 2022-11-03
API Stop using deprecated API

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10542

### DIFF
--- a/code/Extensions/UsedOnTableExtension.php
+++ b/code/Extensions/UsedOnTableExtension.php
@@ -34,7 +34,6 @@ class UsedOnTableExtension extends Extension
      */
     public function updateUsage(ArrayList &$usage, DataObject &$record)
     {
-        Deprecation::notice('1.7.0', 'Use updateUsageExcludedClasses() instead');
         // noop
     }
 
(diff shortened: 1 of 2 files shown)


code/Forms/FileSearchFormFactory.php
```diff
@@ -21,11 +21,6 @@ class FileSearchFormFactory implements FormFactory
     use Extensible;
     use Injectable;
 
-    public function __construct()
-    {
-        $this->constructExtensions();
```
emteknetnz, line 26: Removing call to deprecated noop method Extensible::constructExtensions()

code/Extensions/UsedOnTableExtension.php
```diff
@@ -34,7 +34,6 @@ public function updateUsageExcludedClasses(array &$excludedClasses)
      */
     public function updateUsage(ArrayList &$usage, DataObject &$record)
     {
-        Deprecation::notice('1.7.0', 'Use updateUsageExcludedClasses() instead');
```
emteknetnz, line 37: This deprecation notice just creates noise and doesn't need to be here since it's an extension method that's never called directly

sabina-talipova (approved):

PR #1305, opened 2022-11-04
NEW: Add extension hook for field-specific validation

lozcalver:
Depends on https://github.com/silverstripe/silverstripe-framework/pull/10569

### DIFF
--- a/code/Forms/UploadField.php
+++ b/code/Forms/UploadField.php
@@ -330,7 +330,7 @@ class UploadField extends FormField implements FileHandleField
         $count = count($this->getItems() ?? []);
 
         if ($maxFiles < 1 || $count <= $maxFiles) {
-            return true;
+            return $this->extendValidationResult(true, $validator);
         }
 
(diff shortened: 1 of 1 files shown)


lozcalver:
Would I need to update composer.json to `"silverstripe/framework": "^4.12"` as part of this? I seem to remember we used to do that automatically for minor releases, but I think that stopped?

michalkleiner:
> Would I need to update composer.json to `"silverstripe/framework": "^4.12"` as part of this? I seem to remember we used to do that automatically for minor releases, but I think that stopped?

That would be my expectation to require a version that introduced what's used here, unless we update this PR to include a `$this->hasMethod('extendValidationResult')` check.

lozcalver:
Removed the composer.json update as both PRs are now targeting 5.0.0

ChloeSartorelli:
Rebased on top of 2.0 to get rid of the unrelated CI failures.

ChloeSartorelli (approved):

PR #1310, opened 2022-12-08
FIX Cast absoluteUrl() argument to string

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/642

Similar to https://github.com/silverstripe/silverstripe-elemental/pull/1029


ChloeSartorelli (approved):
See https://github.com/silverstripe/silverstripe-cms/pull/2806#discussion_r1046613902 for a discussion on this approach - it'd be nicer for `AssetAdmin::getFileEditLink()` to always return a string, but strong-typing is out of scope for now (ironically, given strongly typing caused the problem in the first place)

PR #1316, opened 2023-01-10
FIX Correctly reset file details after bulk publish

ChloeSartorelli:
Fixes an error from https://github.com/silverstripe/silverstripe-asset-admin/actions/runs/3726649898/jobs/6320236803
> An element matching css ".font-icon-rocket[name='action_publish']" appears on this page, but it should not. (Behat\Mink\Exception\ExpectationException)

When bulk publishing or bulk unpublishing files, the file details screen wasn't being updated.
This is because react-router now slightly defers its navigation events. If two navigation events occur immediately after one another, the pending navigation is effectively cancelled. So we need to defer the second navigation until the first has finished.

Ultimately it would be better to just re-render the details screen whenever details change,taking advantage of react's state system, but that would require an extensive refactor and partial rewrite of the gallery that I'm just not prepared to do right now.

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/655


maxime-rainville (approved):
LGTM

PR #1320, opened 2023-01-12
MNT Fix js tests

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1421

This is far from complete, I got as far as getting the admin js to be transformed for jest.  There's still a bunch of broken jest tests, most notably AssetAdmin-test.js and AssetAdminRouter-test.js

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,0 +1,6 @@
+{
+    "presets": [
+      "@babel/preset-env",
+      "@babel/preset-react"
+    ]
+}
--- a/client/src/components/AssetDropzone/tests/AssetDropzone-test.js
+++ b/client/src/components/AssetDropzone/tests/AssetDropzone-test.js
@@ -24,15 +24,10 @@ describe('AssetDropzone', () => {
     let item = null;
 
     beforeEach(() => {
-      item = ReactTestUtils.renderIntoDocument(
-        <AssetDropzone {...props} />
-      );
+      item = new AssetDropzone();
     });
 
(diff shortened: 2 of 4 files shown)


babel.config.json
```diff
@@ -0,0 +1,6 @@
+{
```
emteknetnz, line 1: For non-admin modeles it seems that this config needs to be in babel.config.json rather than in package.json/.babelrc - https://babeljs.io/docs/en/configuration#whats-your-use-case

client/src/components/AssetDropzone/tests/AssetDropzone-test.js
```diff
@@ -24,15 +24,10 @@ describe('AssetDropzone', () => {
     let item = null;
 
     beforeEach(() => {
-      item = ReactTestUtils.renderIntoDocument(
-        <AssetDropzone {...props} />
-      );
+      item = new AssetDropzone();
```
emteknetnz, line 27: Cannot no longer call constructor() directly on react components, need to use the `new` keyword

package.json
```diff
@@ -44,7 +44,8 @@
     ],
     "transform": {
       ".*": "babel-jest"
-    }
+    },
+    "testEnvironment": "jsdom"
```
emteknetnz, line 48: This is required for things like 'document' to be globally available

ChloeSartorelli:
Closed in favor of https://github.com/silverstripe/silverstripe-asset-admin/pull/1330

PR #1322, opened 2023-01-18
FIX update react render

ChloeSartorelli:
This should have been done at the same time the main js work was done for installer modules.

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/644


ChloeSartorelli:
js CI failure is to be fixed in https://github.com/silverstripe/silverstripe-admin/issues/1421

emteknetnz (approved):

PR #1303, opened 2022-11-03
DEP Upgrade build stack

ChloeSartorelli:
## Parent issue 
- https://github.com/silverstripe/silverstripe-admin/issues/1312

PR #1330, opened 2023-02-09
FIX Get JS tests passing again

ChloeSartorelli:
Using `FIX` prefix because it affects src files and therefore dist files.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1421


babel.config.json
```diff
@@ -0,0 +1,6 @@
+{
```
ChloeSartorelli, line 1: For non-admin modeles it seems that this config needs to be in babel.config.json rather than in package.json/.babelrc - https://babeljs.io/docs/en/configuration#whats-your-use-case

client/src/components/AssetDropzone/tests/AssetDropzone-test.js
```diff
@@ -24,15 +24,10 @@ describe('AssetDropzone', () => {
...
-      item = ReactTestUtils.renderIntoDocument(
-        <AssetDropzone {...props} />
-      );
+      item = new AssetDropzone();
     });
 
     it('should set this.dropzone to null', () => {
-      item.dropzone = 1;
-      item.constructor(props);
-
```
ChloeSartorelli, line 35: Can no longer call `constructor()` directly on react components, need to use the new keyword

client/src/components/BackButton/tests/BackButton-test.js
```diff
@@ -1,4 +1,4 @@
-/* global jest, jasmine, describe, it, expect, beforeEach */
+/* global jest, describe, it, expect, beforeEach */
```
ChloeSartorelli, line 1: jasmine isn't used or installed

client/src/components/BulkActions/tests/BulkActions-test.js
```diff
@@ -131,7 +131,7 @@ describe('BulkActions', () => {
       const callbackMockFn = jest.fn();
 
       bulkActions.getOptionByValue
-        .mockReturnValueOnce({ confirm: Promise.resolve(), callback: callbackMockFn });
+        .mockReturnValueOnce({ confirm: () => Promise.resolve(), callback: callbackMockFn });
```
ChloeSartorelli, line 134: This wasn't actually hitting the bit of code it's meant to be testing, because `Promise.resolve()` returns an object, and the `handleChangeValue()` method in the `BulkActions` component has a `typeof option.confirm === 'function'` condition.

client/src/components/BulkActions/tests/BulkActions-test.js
```diff
@@ -141,9 +141,9 @@ describe('BulkActions', () => {
       const callbackMockFn = jest.fn();
 
       bulkActions.getOptionByValue
-        .mockReturnValueOnce({ confirm: Promise.reject(), callback: callbackMockFn });
+        .mockReturnValueOnce({ confirm: () => Promise.reject('cancelled'), callback: callbackMockFn });
       return bulkActions.handleChangeValue(event).then(() => {
-        expect(callbackMockFn).toBeCalled();
+        expect(callbackMockFn).not.toBeCalled();
```
ChloeSartorelli, line 146: This test was literally not testing the right thing before. I only caught this because the test started failing with a `UnhandledPromiseRejection` because the rejected promise wasn't being caught.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -58,11 +54,10 @@ class AssetAdmin extends Component {
     this.handleMoveFilesSuccess = this.handleMoveFilesSuccess.bind(this);
   }
 
-  static getDerivedStateFromProps(props) {
-    if (!props.loading && props.folder && props.folderId !== props.folder.id) {
+  componentWillReceiveProps(props) {
+    if ((typeof props.onReplaceUrl === 'function') && !props.loading && props.folder && props.folderId !== props.folder.id) {
       props.onReplaceUrl(props.folder.id, props.fileId, props.query, props.viewAction);
     }
-    return null;
```
ChloeSartorelli, line 65: This was originally changed to `getDerivedStateFromProps` in https://github.com/silverstripe/silverstripe-asset-admin/pull/1303 - but that's causing issues in the tests (for whatever reason they're complaining that the state hadn't been set when it _very_ clearly had) - reverted because this works. We can worry about cleaning up deprecated lifecycle events later (see https://github.com/silverstripe/silverstripe-admin/issues/1414)

client/src/containers/AssetAdmin/tests/AssetAdminRouter-test.js
```diff
@@ -17,19 +17,14 @@ describe('AssetAdminRouter', () => {
...
+        location: {
+          pathname: '',
+          query: {},
+          search: '',
         },
-      },
-      history: {
-        push: jest.fn(),
+        navigate: jest.fn(),
+        params: {},
```
ChloeSartorelli, line 27: The proptypes for the router props changed in https://github.com/silverstripe/silverstripe-admin/pull/1389

client/src/containers/Gallery/tests/Gallery-test.js
```diff
@@ -471,21 +471,6 @@ describe('Gallery', () => {
...
-      props.onCreateFolder = jest.fn();
-
-      gallery = ReactTestUtils.renderIntoDocument(
-        <Gallery {...props} />
-      );
-      gallery.handleCreateFolder();
-
-      expect(props.onCreateFolder).toBeCalled();
-    });
-  });
```
ChloeSartorelli, line 487: This test is trying to call a method `gallery.handleCreateFolder()` that doesn't exist - I have _no_ idea why this didn't scream at us in cms 4 but my guess it is just silently fails.
There is a proper test for the behaviour I think this was intended to test in `client/src/components/GalleryToolbar/Buttons/tests/AddFolderButton-test.js`

client/src/state/imageLoad/tests/ImageLoadActionHandler-test.js
```diff
@@ -36,7 +36,9 @@ describe('ImageLoadActionHandler', () => {
   });
 
   it('eternally failing images end up failed', () => {
-    jest.useFakeTimers();
+    jest.useFakeTimers({
+      legacyFakeTimers: true,
+    });
```
ChloeSartorelli, line 41: See the note about `legacyFakeTimers` in https://jestjs.io/docs/28.x/jest-object#fake-timers
Needed because the tests here treat `setTimeout` as a mock function and this was easier than refactoring all the tests.

client/src/state/imageLoad/tests/ImageLoadActionHandler-test.js
```diff
@@ -49,7 +51,7 @@ describe('ImageLoadActionHandler', () => {
     // Create handler which always fails
     const loadImageHandler = new ImageLoadActionHandler(
       options,
-      (url, resolve, reject) => { reject(); }
+      (url, resolve, reject) => { reject(4); }
```
ChloeSartorelli, line 54: Added to help debug an issue where one of the promise rejections wasn't being caught. Left in so it's easier to find in the future.
The output of the error includes whatever is passed in here, so if it said the issue was with a reject that had `4` passed in we can just search for `reject(4)` and there it is.

client/src/state/uploadField/tests/UploadFieldReducer-test.js
```diff
@@ -122,7 +122,7 @@ describe('uploadFieldReducer', () => {
       });
       expect(nextState.fields.anotherfield.files.length).toBe(2);
       expect(nextState.fields.anotherfield.files[1]).toEqual(
-        jasmine.objectContaining({
+        expect.objectContaining({
```
ChloeSartorelli, line 125: Jasmine isn't installed - `expect.objectContaining()` does the same thing that used to.

package.json
```diff
@@ -29,9 +29,15 @@
...
   "jest": {
+    "testEnvironment": "jsdom",
     "roots": [
       "client/src"
     ],
+    "moduleNameMapper": {
+      "^react-dom/client$": "react-dom-16",
+      "^react-dom((/.*)?)$": "react-dom-16$1",
+      "^react((/.*)?)$": "react-16$1"
+    },
```
ChloeSartorelli, line 40: This and the inclusion of react16 in dev deps reflects how I got tests working in admin in https://github.com/silverstripe/silverstripe-admin/pull/1389/
see https://github.com/silverstripe/silverstripe-admin/issues/1419

package.json
```diff
@@ -54,7 +60,12 @@
     "@storybook/react": "^6.5.13",
     "babel-jest": "^29.3.0",
     "copy-webpack-plugin": "^11.0.0",
+    "enzyme": "^3.11.0",
+    "enzyme-adapter-react-16": "^1.15.6",
```
ChloeSartorelli, line 64: The tests use enzyme. This should have already been here.

package.json
```diff
@@ -99,11 +110,5 @@
   },
   "browserslist": [
     "defaults"
-  ],
-  "babel": {
-    "presets": [
-      "@babel/preset-env",
-      "@babel/preset-react"
-    ]
-  }
```
ChloeSartorelli, line 108: We're using the `babel.config.json` file instead.

emteknetnz (approved):

PR #1331, opened 2023-02-10
ENH Add delete buttons for images/embeds in wysiwyg

ChloeSartorelli:
The stretch goals are shockingly non-trivial, so I'm not touching those. This PR just adds nice delete buttons when you select an image or embed (and behat tests for the selection context toolbars).

I've added the delete button to embeds _and_ images, to keep things consistent.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1410


tests/behat/features/wysiwyg-embed-toolbar.feature
```diff
@@ -0,0 +1,29 @@
+@assets
+Feature: Use the WYSIWYG embed toolbar
+  As a cms author
+  I want to edit and delete embeds using the context toolbar
+  So that I can administer my content efficiently
+
+  Background:
+    Given a "page" "About Us" has the "Content" "[embed url="https://www.example.com/watch?v=4Eyiz8wxC8E" thumbnail="https://via.placeholder.com/200x113" class="embed" width="200" height="113"]https://www.example.com/watch?v=4Eyiz8wxC8E[/embed]"
```
ChloeSartorelli: Shortcode syntax for storing the embed in the db.
Note I'm not using a real youtube or other actually embeddable URL here because I don't want this test to rely on some arbitrary resource actually existing and being accessible at the time the test is run. We don't need that for this test.

tests/behat/src/FixtureContext.php
```diff
@@ -65,19 +65,23 @@ public function iShouldNotSeeTheGalleryItem($name)
     }
 
     /**
-     * @Then /^I should see the "([^"]*)" form$/
+     * @Then /^I should (not |)see the "([^"]*)" form$/
      * @param string $id HTML ID of form
      * @param integer $timeout
      */
-    public function iShouldSeeTheForm($id, $timeout = 3)
+    public function iShouldSeeTheForm($not, $id, $timeout = 3)
```
ChloeSartorelli, line 72: Added the ability to check a form is _not_ present, and using this to check if the insert image modal is present.
Note that the embed modal _doesn't_ check for a form, because a form is only present in that modal if a real embed resource URL is used.

tests/behat/src/FixtureContext.php
```diff
@@ -450,26 +454,44 @@ public function iShouldSeeTheTableGalleryFolderInPosition(string $name, string $
...
-jQuery(doc).find("img[src*='$filename']").each(function() {
-    if(!matched) {
-        rng.selectNode(this);
-        sel.removeAllRanges();
-        sel.addRange(rng);
-        matched = true;
-    }
-});
-JS;
+        var editor = jQuery('#$inputFieldId').entwine('ss').getEditor(),
```
ChloeSartorelli, line 480: Added indents to make this easier to read

tests/behat/src/FixtureContext.php
```diff
@@ -450,26 +454,44 @@ public function iShouldSeeTheTableGalleryFolderInPosition(string $name, string $
...
+        $this->selectInTheHtmlField("div.embed[data-url='$url']", $field);
+    }
+
+    /**
+     * Selects the first match of $select in the given HTML editor (tinymce)
+     */
+    protected function selectInTheHtmlField(string $select, string $field)
     {
         $inputField = $this->getHtmlField($field);
+        $inputField->getParent()->find('css', 'iframe')->click();
```
ChloeSartorelli, line 477: We have to click on the editor first so it has focus and knows this is the "active" editor, or the toolbar doesn't appear.

ChloeSartorelli:
js test failure is expected until https://github.com/silverstripe/silverstripe-asset-admin/pull/1330 is merged in

tests/behat/features/wysiwyg-embed-toolbar.feature
michalkleiner: > Note I'm not using a real youtube or other actually embeddable URL here because I don't want this test to rely on some arbitrary resource actually existing and being accessible at the time the test is run.

But still relying on placeholder.com. I guess there's no point using a local image instead?

ChloeSartorelli: Yeah that's a fair point. I didn't want to hardcode an image URL but I guess there's no harm in it - we'll always be running this test with default config presumably so the URL should be reliable. I'll (try to remember to) make that change on Monday.

emteknetnz (commented):
This works good, **do we need to update the behat test re placeholder.com?**

**Before**
image

**After**
image

tests/behat/features/wysiwyg-embed-toolbar.feature
emteknetnz: Do we need to make an update here?

ChloeSartorelli: Yes, sorry, I forgot to do that. Will do along with any other changes you recommend when you've finished your review

emteknetnz: Just need to sort out placeholder.com (and maybe example.com)

tests/behat/features/wysiwyg-embed-toolbar.feature
```diff
@@ -0,0 +1,29 @@
...
+      And the "group" "EDITOR" has permissions "Access to 'Pages' section"
+      And I am logged in as a member of "EDITOR" group
+      And I go to "/admin/pages"
+      And I click on "About Us" in the tree
+    # Validate that everything is ready for the test scenarios
+    Then I should not see a ".tox-pop__dialog .tox-toolbar" element
+      And I should not see the ".insert-embed-react__dialog-wrapper" element
+
+  Scenario: I can open the media edit modal from the context toolbar
+    When I select the media "https://www.example.com/watch?v=4Eyiz8wxC8E" in the "Content" HTML field
```
emteknetnz, line 19: Also here (maybe)

tests/behat/features/wysiwyg-embed-toolbar.feature
```diff
@@ -0,0 +1,29 @@
...
+    Then I should see a ".tox-pop__dialog .tox-toolbar button[title='Edit media']" element
+    When I press the "Edit media" button
+    Then I should see "Media from the web" in the ".insert-embed-react__dialog-wrapper" element
+
+  Scenario: I can delete the media from the context toolbar
+    When I select the media "https://www.example.com/watch?v=4Eyiz8wxC8E" in the "Content" HTML field
+    Then I should see a ".tox-pop__dialog .tox-toolbar button[title='Delete media']" element
+    When I press the "Delete media" button
+    Then the "Content" HTML field should not contain "https://www.example.com/watch?v=4Eyiz8wxC8E"
```
emteknetnz, line 28: And here (maybe)

emteknetnz (changes requested):

tests/behat/features/wysiwyg-embed-toolbar.feature
ChloeSartorelli: example.com is okay - this is a domain that's explicitly intended to be used for example URLs and won't cause problems for us even if it suddenly becomes unavailable with the way this test is set up. Will replace the placeholder URL with some relative file.

ChloeSartorelli: Done

ChloeSartorelli: Done

ChloeSartorelli:
Rebased so js tests can pass, and resolved feedback.

emteknetnz (approved):

PR #1332, opened 2023-02-21
FIX Ensure img is fetched and exists in tinymce plugin

ChloeSartorelli:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1333

This PR avoids this scenario by:
1. Trying to get the image from the child of the currently selected node (in debugging I found the `<p>` tag wrapping the image was selected when the error occurs)
2. Not calling `setAttribute()` if the result is still `undefined`.


emteknetnz (approved):
Test locally seems good. I wasn't able to replicate the original issue, though this PR also doesn't appear to break anything so I'm happy

PR #1340, opened 2023-03-24
Versioned module required

alessandromarotta:
Module broken without Silverstripe/Versioned installed but there's not dependency in composer.json

@see https://github.com/silverstripe/silverstripe-versioned/pull/399

### DIFF
--- a/_config/versionedfiles.yml
+++ b/_config/versionedfiles.yml
@@ -1,0 +1,6 @@
+---
+Name: versionedfiles
+---
+SilverStripe\Assets\File:
+  extensions:
+    - SilverStripe\Versioned\Versioned
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,8 @@
         "php": "^8.1",
         "silverstripe/framework": "^5",
         "silverstripe/admin": "^2",
-        "silverstripe/graphql": "^5"
+        "silverstripe/graphql": "^5",
+        "silverstripe/versioned": "^2"
     },
     "require-dev": {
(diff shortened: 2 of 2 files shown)


_config/versionedfiles.yml
```diff
@@ -0,0 +1,6 @@
+---
+Name: versionedfiles
```
michalkleiner, line 2: For the tests to pass you'd need to rename this section, otherwise it clashes with the file within the versioned module where the removal isn't merged yet.

michalkleiner:
I wonder whether adding the extension in this module is the right place as well since it's being added to the `File` class from the `assets` module. The assets module can function without versioning even though it's rarely used that way. It's not a hard dependency there. The assets admin (this) module should also work without versioning, even though it's probably not fully decoupled in all places.

ChloeSartorelli:
This seems very much like it belongs in versioned, where it already is. 
Framework requires assets. 
Assets doesn't (theoretically) need versioned.
And this admin module as Michal points out is definitely the wrong place for this config, which is about how assets and versioned work together. 

I'm going to close this and https://github.com/silverstripe/silverstripe-versioned/pull/399 - but if you have a strong argument for why this change should be made please do comment and we'll see if we need to reconsider.

alessandromarotta:
Assets doesn't require versioning but the Versioned module (installed by the Admin module) force you to have it. Actually which need the Versioned module is the Assets Admin because has code coupled with Versioned module.

PR #1350, opened 2023-04-19
DOC Update README.md for CMS 5

ChloeSartorelli:
## Issue

- https://github.com/silverstripe/.github/issues/20

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,9 @@ The Asset Admin module provides a interface for managing files within Silverstri
 
 ## Installation
 
+```sh
+composer require silverstripe/asset-admin
 ```
-$ composer require silverstripe/asset-admin
-```
-
-You'll also need to run `dev/build`.
 
 ## Versioning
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,11 +9,9 @@ The Asset Admin module provides a interface for managing files within Silverstri
 
 ## Installation
 
+```sh
+composer require silverstripe/asset-admin
 ```
-$ composer require silverstripe/asset-admin
-```
-
-You'll also need to run `dev/build`.
```
sabina-talipova, line 16: I've already merge couple another PR with similar changes, but small question.
Is it correct that we remove recommendation about `dev/build`?

ChloeSartorelli: Yeah I noticed this was present in a couple of modules but not in others. I don't think it really matters too much one way or another... it's pretty common knowledge and is documented in the main ss docs (or at least should be) that this is needed any time database schema is updated. And I'd argue nobody should be installing modules without knowing what sort of affect that module will have.

tl;dr: 🤷‍♀️

sabina-talipova (approved):
LGFM

Issue #1352, opened 2023-04-20
v2 UploadField Regression: "You cannot upload any more files"


chrispenny:
## Setup

```php
private static array $has_one = [
    'HeroImage' => Image::class,
];
```

```php
$fields->addFieldsToTab(
    'Root.Hero',
    [
        UploadField::create('HeroImage', 'Hero Image'),
        ...
    ]
);
```

## Description

1. Select a file in your `UploadField`
2. Save your Page/DO
3. Reload edit form
4. Remove the file from the `UploadField`
5. Select "Upload new"
6. Select a file to upload

Result: "You cannot upload any more files"

https://user-images.githubusercontent.com/505788/233233955-ea414c81-2c88-428e-95f1-098cff40191f.mov

## v1

Tested the same behaviour in a Silverstripe 4 project, and did not experience this behaviour.

## Acceptance criteria
- You can select or upload a new file after removing a file from a single-file upload field.
- You can not upload or select more than 1 file in a single file upload field
- Add a Behat test to cover this scenario
- If you have a multi file upload field with a file limit, that limit is still honored.

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1353

ChloeSartorelli:
Just for the sake of clarity, I'm assuming this is with a `has_one` relationship?
Did you set any additional config (either globally or per instance) for the upload field or validator?

chrispenny:
Thanks, @GuySartorelli. Correct, it's a `has_one` (added to the description).

The project is also running [KFC](https://github.com/silverstripe-terraformers/keys-for-cache) and Fluent, but KFC doesn't impact anything until after a write is performed, and for Fluent this field is not localised.

ChloeSartorelli:
I can definitely reproduce this, but I can't figure out what's causing it.
I can see that the number of files that our react components are aware of is correct all throughout the process.
I've also console logged the files array from the underlying dropzone instance and _that_ seems to always be correct as well. So I'm not sure what's causing it to think we've hit the limit.

ChloeSartorelli:
PR merged
@chrispenny You can use `2.0.x-dev` as a dependency constraint for now if you need to, but we're intending to launch CMS 5 stable pretty soon.

chrispenny:
Thanks, team!!

purplespider:
I'm still getting a seemingly related "You can not upload any more files." bug in 5 stable.

To reproduce:
[...]

PR #1353, opened 2023-04-25
FIX Ensure fresh options are passed to dropzone

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1352

Reverts part of this commit - https://github.com/silverstripe/silverstripe-asset-admin/commit/45e7461a549ae858999eacd8d11eaeb4bbe13515#


ChloeSartorelli (changes requested):
The behat test passes even without the fix - please update the behat test so that it fails without the fix, but passes with it (probably explicitly check that "You can not upload any more files." is not there after uploading the second file)

emteknetnz:
I don't think this is possible unfortunately.

This regression only happens when you click "Upload new" rather that "Choose existing', and you select a file from the (native) file selector dialog.

Unfortunately there isn't a way to get behat to "select a file from the file selector" - when you get behat to click "Upload new", there is no file selector, though the link "Upload new" is underlined.

There are no existing tests where "Upload new is clicked", in fact all the "upload" tests will "Choose existing" i.e. pre-uploaded files - so I'm not the first person to run into this issue.

I think we'll just have to make do with the behat test as it is.

ChloeSartorelli:
Bugger. Well, we can't win em all 😅

ChloeSartorelli (approved):
Works great locally.

PR #1359, opened 2023-05-18
Adds Folder ID for TinyMCE_sslink-file to allow us to target a specif…

gazzayeatman:
Targeting 1.11 as thats the version we're using in this case.

Copied the code from TinyMCE_ssmedia to allow us to send a folderID to TinyMCE_sslink-file. 

Tested locally for our specific use case and it appears to work. 

The problem we are trying to solve. Our client has a website hosts pages for specific centres across Australia. They have a folder structure where each centre has a directory to upload files to. They also have set up a permissions model to only allow a specific centre to view their own directory and no others.

Previously when they were trying to link a file they would be met with an exception because they were defaulted to the root assets directory which they did not have access to. We would like to be able to point them to their centre's directory by default so they don't get met with this exception and improve the UX. I hope that helps.


ChloeSartorelli (changes requested):
Thank you for contributing, this seems like it could be useful in some cases - though we'll probably need to review it from a UX perspective to make sure content authors don't accidentally link to a folder when they don't mean to since in most projects that would actually be an error state.

We'll probably need some configuration to enable this feature, false by default, to avoid causing problems in that more common case.

I haven't taken the time to review this properly yet but just noting a pre-requisite for this to be merged - please target the `2` branch. As a new feature, this cannot be merged into a CMS 4 compatible branch - it'll have to be released in a future minor release of CMS 5 if it is merged at all.

While you're at it, can you please also update the commit name to conform with our [contributing guidelines](https://docs.silverstripe.org/en/5/contributing/code/#commit-messages)?

michalkleiner:
@GuySartorelli I think the change is about being able to provide a default folder for file upload fields so that the user don't need to always select it, not about allowing linking to a folder instead of a file. Correct me if I'm wrong @gazzayeatman. With that, I wouldn't expect this needing specific UX review. Perhaps some UI test, but also not critical from my perspective.

ChloeSartorelli:
Ahh okay - in that case @gazzayeatman can you please write a clear description of what problem this is solving in the pr description? Just to clear up exactly what this is intended to do.

gazzayeatman:
@GuySartorelli Sorry for the delay. I've updated the description

ChloeSartorelli (approved):
Thanks for that. It works well locally.
I would normally request some documentation for this, but it doesn't look like there's any documentation for this functionality for the media plugin either so that's fine.

I've re-targetted this to `1.13` as we don't support earlier versions and there's administrative overhead with merge-ups etc if we were to release it on an earlier minor version.

ChloeSartorelli:
It could be argued that this is a new feature, but honestly it seems broken to me that the media plugin would respect the folder for the wysiwyg config but the link to file plugin didn't - so I'm considering this a bug fix which allows us to release it as a patch for CMS 4.

Released as [1.13.2](https://github.com/silverstripe/silverstripe-asset-admin/releases/tag/1.13.2). It'll also get a patch release against the `2.0` minor at some stage after merge-ups are done.

emteknetnz:
This has been released in [2.0.1](https://github.com/silverstripe/silverstripe-asset-admin/releases/tag/2.0.1)

Issue #1362, opened 2023-05-29
UploadField Regression: "You can not upload any more files" on new object


purplespider:
User unexpectedly gets a "_You can not upload any more files_" message when trying to upload a file in a new Page/DO.

To reproduce:
1. **Upload** a file to an empty UploadField.
2. **Save/Create**
3. Create a **new** Page/DO
4. **Upload** a file to the empty UploadField.

Result: **"You can not upload any more files."**

If I reload the page, then it will let me upload as expected.

Alternative reproduction explanation from @chrispenny: https://github.com/silverstripe/silverstripe-asset-admin/issues/1352#issuecomment-1566234301

Issue in Silverstripe 5 stable.

Seems to be a different manifestation of https://github.com/silverstripe/silverstripe-asset-admin/issues/1352

![Screenshot 2023-05-26 at 15 03 24](https://github.com/silverstripe/silverstripe-asset-admin/assets/329880/15e1c4ba-fc06-4f40-a9b3-edeff8d30093)

## Acceptance criteria
- Upload field on new data object doesn't block you from uploading files.

## PR
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1368

rspcactucker:
Replicating on my SS5 repo as well. Seems to be originating entirely from the Dropzone client-side code (no network activity as far as I can see?)

emteknetnz:
Linked pull-request has been merged and has been released as part of [2.0.1](https://github.com/silverstripe/silverstripe-asset-admin/releases/tag/2.0.1)

PR #1368, opened 2023-06-15
FIX Upload files issue in AssetDropzone

sabina-talipova:
## Description
We pass `canEdit` permission to the `AssetDropzone Component` as `canUpload` props.
Then in the conditional block `if (this.props.canUpload && prevProps.canUpload !== this.props.canUpload) {}` in `componentDidUpdate` function we compare `prevProps.canUpload = true|false` with `this.props.canUpload = true|false` and get `false`.
Replaced with `prevProps.options !== this.props.options`.

## Parent issue
- https://github.com/silverstripe/silverstripe-asset-admin/issues/1362


emteknetnz (approved):
Tested locally, works good. Also tested there was no regression of https://github.com/silverstripe/silverstripe-asset-admin/issues/1352

PR #1369, opened 2023-06-15
FEAT update CMS fields to allow user-specific permissions

andrewandante:
Companion to https://github.com/silverstripe/silverstripe-framework/pull/10819 and https://github.com/silverstripe/silverstripe-assets/pull/556

Adds CMS field to Asset admin to allow for user-specific permission management.

## Issue
- silverstripe/silverstripe-framework#10817

### DIFF
--- a/code/Forms/AssetFormFactory.php
+++ b/code/Forms/AssetFormFactory.php
@@ -16,6 +16,7 @@ use SilverStripe\Forms\FormAction;
 use SilverStripe\Forms\FormFactory;
 use SilverStripe\Forms\HeaderField;
 use SilverStripe\Forms\HiddenField;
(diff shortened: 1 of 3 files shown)


ChloeSartorelli (approved):
We really need a core implementation of displaylogic that works in both entwine and react contexts 😅 But that's way out of scope - for now continuing to show the fields is suboptimal but acceptable.

This works great locally. Will merge after the assets and framework PRs are both good to go.

ChloeSartorelli (changes requested):
Oh! There is one change required actually. Can you please bump the framework constraint in composer.json to `^5.1`?

ChloeSartorelli (approved):
LGTM

PR #1371, opened 2023-06-28
FIX ESLint issues

sabina-talipova:
## Description
Turn on some of the ESLint.
## Parent issue
-  https://github.com/silverstripe/silverstripe-admin/issues/1433


client/src/components/GalleryItem/tests/SelectGalleryItemTracker.js
```diff
@@ -16,9 +16,7 @@ class SelectGalleryItemTracker extends Component {
 
   handleSelect(event, item) {
     triggerAction(this.state.selected ? 'unchecked' : 'checked', item.title);
-    this.setState({
-      selected: !this.state.selected,
-    });
+    this.setState((prevState) => ({ selected: !prevState.selected }));
```
emteknetnz, line 19: Is this sort of change required to get linting to pass? Otherwise I'd prefer to not have any actual code changes.

emteknetnz (changes requested):

emteknetnz (changes requested):
.

client/src/components/GalleryItem/tests/SelectGalleryItemTracker.js
sabina-talipova: This is the rule [react/no-access-state-in-setstate](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md).
```txt
Usage of this.state inside setState calls might result
in errors when two state calls are called in batch and thus referencing 
old state and not the current state.
``` 
This problem describes [there](https://legacy.reactjs.org/docs/react-component.html#setstate)
```txt
setState() does not always immediately update the component.
It may batch or defer the update until later. This makes reading this.state right after
calling setState() a potential pitfall. Instead, use componentDidUpdate or a setState callback
(setState(updater, callback)), either of which are guaranteed to fire after the update has been applied. 
If you need to set the state based on the previous state, read about the updater argument below.
```

emteknetnz (approved):

PR #1376, opened 2023-08-07
MNT Behat test to test special characters in shortcode

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10451

### DIFF
--- a/tests/behat/features/insert-an-image.feature
+++ b/tests/behat/features/insert-an-image.feature
@@ -105,6 +105,36 @@ Feature: Insert an image into a page
       # Required to avoid "unsaved changed" browser dialog
       And I press the "Save" button
 
+  Scenario: I can add text with special characters as Alternative text
+    When I press the "Insert from Files" HTML field button
+      And I select the file named "folder1" in the gallery
+      And I click on the file named "file1" in the gallery
+    Then I should see the "Form_fileInsertForm" form
+    When I fill in "Alternative text (alt)" with "My alt updated & saved"
+    When I fill in "Title text (tooltip)" with "My title text updated & saved"
+      And I press the "Insert" button
+    Then the "Content" HTML field should contain "file1.jpg"
+      And the "Content" HTML field should contain "My alt updated &amp; saved"
+      And I press the "Save" button
+    # We need this to update DB with new value after it was created in the previous step
+    When I press the "Save" button
+      Then the "Content" HTML field should contain "file1.jpg"
+      And the "Content" HTML field should contain "My alt updated & saved"
+      And the "Content" HTML field should contain "My title text updated & saved"
+    When I select the image "file1.jpg" in the "Content" HTML field
+      And I press the "Insert from Files" HTML field button
+      And I should see the "Update file" button
+    When I fill in "Caption" with "My caption updated & saved"
+      And I press the "Update file" button
+    Then the "Content" HTML field should contain "My alt updated &amp; saved"
+      And the "Content" HTML field should contain "My title text updated &amp; saved"
+      And the "Content" HTML field should contain "My caption updated &amp; saved"
+      And I press the "Publish" button
+    Then I go to "/about-us"
+      And I should see an "img[alt='My alt updated & saved']" element
+      And I should see an "img[title='My title text updated & saved']" element
+      And I should see "My caption updated & saved"
+
   Scenario: I can link to a file
     Given I select "awesome" in the "Content" HTML field
     When I press the "Insert link" HTML field button


sabina-talipova:
Require [PR](https://github.com/silverstripe/silverstripe-assets/pull/565)

ChloeSartorelli:
I've rerun the behat tests but they're failing. Have confirmed that behat is running against the recent changes, so there's either something wrong with the test or the new logic.

sabina-talipova:
Fixed

ChloeSartorelli (changes requested):
It looks like this still isn't working as expected.

I would expect to see `My alt updated &amp; saved` in the HTML, but this test shows `My alt updated & saved` which is incorrect.

We should also add to the behat test what happens on the front-end, to validate that the frontend markup has `alt="My alt updated &amp; saved"` and not `alt="My alt updated & saved"` or `alt="My alt updated &amp;amp; saved"`.

sabina-talipova:
Follow [TinyMC documentation] (https://www.tiny.cloud/docs/tinymce/6/content-filtering/#entities)  and [this](https://www.tiny.cloud/docs/tinymce/6/content-filtering/#entity_encoding):

>  The base entities <, >, &, ', and " will always be entity encoded into their named equivalents.

So my assumption, when we just start to work on content and add new special characters TinyMC will always encoded them into their named equivalents. So this we see this here.
**First string that I get before "Save" (named equivalents):**
```js
document.querySelector('textarea[name="Content"]').value
'<div class="captionImage leftAlone" style="width: 444px;">[image src="http://localhost:90/sink413x/assets/Uploads/GitHub-avatar-v2.jpeg" id="3" width="444" height="444" class="leftAlone ss-htmleditorfield-file image" alt="Batman &amp; Robin"]\n<p class="caption leftAlone">Batman &amp; Robin</p>\n</div>'
```
**Second string that I get after "Save" (entities):**
```js
document.querySelector('textarea[name="Content"]').value
'<div class="captionImage leftAlone" style="width: 444px;">[image src="/sink413x/assets/Uploads/GitHub-avatar-v2.jpeg" id="3" width="444" height="444" class="leftAlone ss-htmleditorfield-file image" alt="Batman & Robin"]\n<p class="caption leftAlone">Batman & Robin</p>\n</div>'
```

**Third string that I get after I made some changes, but not "Save" yet (named equivalents):** 
```js
document.querySelector('textarea[name="Content"]').value
'<div class="captionImage leftAlone" style="width: 444px;">[image src="/sink413x/assets/Uploads/GitHub-avatar-v2.jpeg" id="3" width="444" height="444" class="leftAlone ss-htmleditorfield-file image" alt="Batman &amp; Robin"]\n<p class="caption leftAlone">Batman &amp; Robin &amp; Co</p>\n</div>'
```
This string we pass to the server from the client :
```txt
<div class="captionImage leftAlone" style="width: 444px;">[image src="/sink413x/assets/Uploads/GitHub-avatar-v2.jpeg" id="3" width="444" height="444" class="leftAlone ss-htmleditorfield-file image" alt="Batman &amp; Robin"]
<p class="caption leftAlone">Batman &amp; Robin &amp; Co</p>
</div>
```
This string we store in DB:
```txt
<div class="captionImage leftAlone" style="width: 444px;">[image src="/sink413x/assets/Uploads/GitHub-avatar-v2.jpeg" id="3" width="444" height="444" class="leftAlone ss-htmleditorfield-file image" alt="Batman &amp; Robin"]
<p class="caption leftAlone">Batman &amp; Robin &amp; Co</p>
```
It looks like TinyMC convert html entities into named equivalents in whole content every time when we edit it. When we return `&` as is from DB (I updated `&amp;` with `&` in string in DB), then TinyMC shows it as is, but encode it to `&amp;` as soon as we make any small changes (add 'Space'). 

So for behat test we should expect that after "Save" we will get `My alt updated & saved` instead of `My alt updated &amp; saved`. I add some test to show all this explanation above.

tests/behat/features/insert-an-image.feature
```diff
@@ -105,6 +105,36 @@ Feature: Insert an image into a page
...
+      And the "Content" HTML field should contain "My alt updated & saved"
+      And the "Content" HTML field should contain "My title text updated & saved"
+    When I select the image "file1.jpg" in the "Content" HTML field
+      And I press the "Insert from Files" HTML field button
+      And I should see the "Update file" button
+    When I fill in "Caption" with "My caption updated & saved"
+      And I press the "Update file" button
+    Then the "Content" HTML field should contain "My alt updated &amp; saved"
+      And the "Content" HTML field should contain "My title text updated &amp; saved"
+      And the "Content" HTML field should contain "My caption updated &amp; saved"
```
ChloeSartorelli, line 131: This doesn't seem right - when we click "update file", we should see the same thing we filled in, i.e. the unchanged `My title text updated & saved` value.
Looks like we still have a little bit of work left to do here in the implementation.

tests/behat/features/insert-an-image.feature
```diff
@@ -105,6 +105,36 @@ Feature: Insert an image into a page
...
+      And I press the "Publish" button
+    Then I go to "/about-us"
+      And I should see an "img[alt='My alt updated &amp; saved']" element
+      And I should see an "img[title='My title text updated &amp; saved']" element
```
ChloeSartorelli: It may not seem like it, but this actually means we're _still double encoding_. These should be queryable as `img[alt='My alt updated & saved']` and `img[title='My title text updated & saved']`

I've gone back and re-run my manual tests. It turns out that the inspector in both firefox and chrome shows the decoded version of the string. So I saw this:
Screenshot from 2023-08-10 09-48-40
`alt="some &amp; thing"`

But the actual value (seen by right clicking and selecting "Edit as HTML") was this:
Screenshot from 2023-08-10 09-48-52
`alt="some &amp;amp; thing"`

So we're still double encoding. Sorry I didn't catch it before merging the changes.

ChloeSartorelli (changes requested):

tests/behat/features/insert-an-image.feature
sabina-talipova: All changes that were made are correct, but they fixed problem in CMS itself. Current problem happens now when we display Content on a page.

sabina-talipova: I'll fix it and update Behat tests.

sabina-talipova: Could you, please review [this PR](https://github.com/silverstripe/silverstripe-assets/pull/567)

sabina-talipova:
Tests require [this PR](https://github.com/silverstripe/silverstripe-assets/pull/567)

ChloeSartorelli (approved):
Looks good - it's green now the changes are merged.

PR #1377, opened 2023-08-08
Allows display of validation messages on assets

### LINKED ISSUE #967: ValidationResult::addError() doesn't show up anywhere in the UI
**Version: 4.4.x-dev**

## Steps to reproduce

Add this to `Image`:

```php
    public function validate()
    {
        $validation = parent::validate();
        if ($this->isInDB() && empty($this->Title)) {
            $validation->addError('Testing error');
        }
        return $validation;
    }
```

Upload an image. This is successful (as expected) because the isInDB() check skips validation.

Open the image and remove the title, then try to save the image.

## Expected result

The image should not save. I should see "Testing error" somewhere in an error alert.

## Result

The image doesn't save (correct). No validation error is shown in the UI (bug).

---

This works with `$validation->addFieldError('Title', 'Testing error')`:

image

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1377 (for new files only - doesn't address editing existing files)
- https://github.com/silverstripe/silverstripe-admin/pull/1587

blueo:
Fixes #967 introduced in https://github.com/silverstripe/silverstripe-asset-admin/pull/1138

Adds an optional xhr parameter to the getStatusCodeMessage function to allow displaying custom validation messages from the CMS.


ChloeSartorelli:
Thanks for submitting this. It's a good start, and handles the case for uploading new files - but it seem to resolve the main problem described in the issue you've linked to, which is that no error is displayed when you save a file which _already exists_.

Would you like to try tackling that part of the issue? Or would you prefer this fix to be merged as-is, as it is an objective improvement over the current state of affairs.

If you'd like to have this merged as is, please retarget the `1.13` branch so it can be released as a patch.

blueo:
> which is that no error is displayed when you save a file which _already exists_

thanks @GuySartorelli tbh I was only trying to address the new file scenario as that's the immediate use case we've got. I'm not opposed to looking into the existing file but ideally as a separate change if possible? I'll retarget the branch

ChloeSartorelli:
> I'm not opposed to looking into the existing file but ideally as a separate change if possible? 

Yup no worries. I'll leave the issue open as it isn't resolved, and I'll merge this once ci goes green.

ChloeSartorelli (approved):
Looks good, works well locally

ChloeSartorelli:
Thisnwill be auto-tagged once CI has finished running on the branch.

PR #1395, opened 2023-09-06
FIX Image tag is converted to shortcode even if it doesn't have parent node

sabina-talipova:
## Description
HTML Editor Content is wrapped into `div` tag before additional processing. After processing function returns content without wrapper.

## Additional test preparations
- Add the following code to `app/_config.php`:
```php
HTMLEditorConfig::get('cwp')->setOptions([
    'forced_root_block' => '',
...
     'extended_valid_elements' =>
        'img[class|src|alt|title|hspace|vspace|width|height|align|name|usemap|data*],'
...
        . 'script[src|async|defer|type|charset],'
])
```  

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10884


client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -108,8 +107,8 @@ const filter = 'img[data-shortcode="image"]';
         // Note that <script> tags might be sanitized separately based on editor config.
         o.content = '';
         content.each(function () {
-          if (this.outerHTML !== undefined) {
-            o.content += this.outerHTML;
+          if (this.innerHTML !== undefined) {
```
emteknetnz, line 110: There's a comment above this "Insert outerHTML in order to retain all nodes incl. <script>" - will need to update that

emteknetnz (changes requested):
Tested locally, works good

Will just need to update inline comment that mentions outerHTML

client/src/entwine/TinyMCE_ssmedia.js
sabina-talipova: I've done some tests with `<script>` tag and it looks like work. Since now we have additional `div` wrapper in line 78.
If I leave `outerHTML` we also save `<div>` element, that we don't need to save.

emteknetnz: Sorry I mean update the inline comment on line 105 and just change `outerHTML` to `innerHTML` since that's what we're using now

emteknetnz (changes requested):

client/src/entwine/TinyMCE_ssmedia.js
sabina-talipova: Updated

emteknetnz (approved):

PR #1399, opened 2023-09-11
fix: expose ShowInSearch field in access UI

wilr:
Upgrading from SilverStripe 3 to SilverStripe 5 it appears the ShowInSearch field is no longer displayed in the UI. 

The File record still contains this data and the implementation was to match Page instances which contain the 'Show in search' option to CMS editors.

## Issue
- https://github.com/silverstripe/silverstripe-asset-admin/issues/1400

### DIFF
--- a/code/Forms/AssetFormFactory.php
+++ b/code/Forms/AssetFormFactory.php
@@ -9,6 +9,7 @@ use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Forms\CheckboxField;
 use SilverStripe\Forms\FieldGroup;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\Form;
@@ -152,13 +153,13 @@ abstract class AssetFormFactory implements FormFactory
     {
         if ($record && $record->isInDB() && $record->canEdit()) {
             /** @var FormAction $action */
-            $action = FormAction::create('save', _t(__CLASS__.'.SAVE', 'Save'))
+            $action = FormAction::create('save', _t(__CLASS__ . '.SAVE', 'Save'))
                 ->setIcon('save')
                 ->setSchemaState([
                     'data' => [
-                        'pristineTitle' => _t(__CLASS__.'.SAVED', 'Saved'),
+                        'pristineTitle' => _t(__CLASS__ . '.SAVED', 'Saved'),
                         'pristineIcon' => 'tick',
-                        'dirtyTitle' => _t(__CLASS__.'.SAVE', 'Save'),
+                        'dirtyTitle' => _t(__CLASS__ . '.SAVE', 'Save'),
                         'dirtyIcon' => 'save',
                         'pristineClass' => 'btn-outline-primary',
                         'dirtyClass' => '',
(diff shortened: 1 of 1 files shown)


ChloeSartorelli:
There's something odd happening with the behat tests... can you please rebase your PR to see if that resolves it?

wilr:
Rebased now.

ChloeSartorelli:
This PR doesn't seem to fix it - I still can't see the ShowInSearch field when editing a file in the asset admin:
image

Can you please advise if there's some additional steps I need to take, or if I'm looking in the wrong place, etc?

wilr:
@GuySartorelli should be under 'Permissions'

ChloeSartorelli:
Oh, interesting. Was it under "permissions" in CMS 3? That doesn't seem like an intuitive place for it, but if that's where it was in CMS 3 then I'd be okay with it.

Please add an exception for Folder - there is a comment in the $db definition that says it shouldn't be inherited for folder, so we shouldn't show the field for folders.

wilr:
@GuySartorelli Done added a check to exclude folder.

ChloeSartorelli:
Awesome, thanks for making that change. Can you please also answer the question from my previous comment?
> Was it under "permissions" in CMS 3? That doesn't seem like an intuitive place for it, but if that's where it was in CMS 3 then I'd be okay with it.

code/Forms/AssetFormFactory.php
```diff
@@ -152,13 +153,13 @@ protected function getSaveAction($record)
     {
         if ($record && $record->isInDB() && $record->canEdit()) {
             /** @var FormAction $action */
-            $action = FormAction::create('save', _t(__CLASS__.'.SAVE', 'Save'))
+            $action = FormAction::create('save', _t(__CLASS__ . '.SAVE', 'Save'))
```
ChloeSartorelli, line 156: Can you please either avoid making unrelated changes like this (and raise a separate PR for them) or at least do these in a separate commit? It is very hard to see which parts of the PR are relevant changes vs just small nothing-changes like this when you have them all in one commit.

wilr: @GuySartorelli Pro-tip just chuck a `?w=1` on the URL and you can see the changes sans the Vscode auto-linting fix. Ideally we should have some sort of hooks so things are linted by default when coming in.

ChloeSartorelli: Thanks for the tip - my request still stands for future pull requests (you can leave as is for this one). I don't like ignoring anything when reviewing pull requests.

ChloeSartorelli:
Can you please resolve the behat failures, and also respond to my question in the previous comments?

wilr:
> Can you please resolve the behat failures, and also respond to my question in the previous comments?

Not sure about 3 but I think it's fine under Permissions as it relates to some form of Permission (is it permitted to be in search).

ChloeSartorelli:
Permissions is more about _user_ permissions than what the record is "permitted" to do. I wouldn't like to make a precedent that this sort of thing is a "permission".

If this is where it was for CMS 3 then I'm fine with it. If it wasn't there in CMS 3 or we can't verify one way or another, please move it out of permissions.

ChloeSartorelli:
This pull request hasn't had any activity for a while. If you're still interested in having this merged, can you please either move it out of permissions or else validate where it was in CMS 3? If there's no more action after a few weeks I will close the PR, and you can reopen it when you're ready to work on it again.

wilr:
Hi @GuySartorelli happy to move it out to main. In this 3 build there were no tabs. Everything is under 'Main'.

ChloeSartorelli:
Sounds good - when that change is made, I'll be happy to re-review and merge.

wilr:
@GuySartorelli done.

ChloeSartorelli (approved):
LGTM, will merge on green.

Issue #1401, opened 2023-09-12
The "Link to a File" option in the TinyMCE WYSIWYG allows links with no link text


sabina-talipova:
## Description

<img width="521" alt="Screenshot 2023-09-12 at 2 28 23 PM" src="https://github.com/silverstripe/silverstripe-asset-admin/assets/87288324/a8c91b94-654a-4476-9d83-66bb7c5ebe15">

User can add link to file without "Link text". This field is not required. But it looks like link was added, since I can see additional actions like "Edit link / Remove Link". But if I save page a link disappeared.

## PR
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1403

emteknetnz:
Linked PR has been merged

PR #1403, opened 2023-09-25
FIX Insert file link text is required

sabina-talipova:
## Description
There is not proper validation for file link text, and it means that user can insert link to the file as empty string without any content.

## Parent issue
- https://github.com/silverstripe/silverstripe-asset-admin/issues/1401


emteknetnz (approved):
Tested locally, worked nicely

PR #1405, opened 2023-10-03
MNT Fix behat test

ChloeSartorelli:
Probably caused by a bad merge-up. The TinyMCE 4 selector was being used instead of the TinyMCE 6 one.

## Issue
- https://github.com/silverstripe/.github/issues/125


client/dist/styles/bundle.css
ChloeSartorelli, line 1: Not sure why the css file was changed, but the js job failed at first and after running `yarn build` I got this changed file which matches the sha1sum of the original js build (see https://github.com/silverstripe/silverstripe-asset-admin/actions/runs/6387816339/job/17337061941)

emteknetnz (approved):

PR #1436, opened 2024-01-25
FIX TinyMCE Insert link from Files

sabina-talipova:
## Description
- It looks like using `$` conflicts with another library, so it was replaced with `jQuery`;
- Fix existing `getSelection` function issue. ~~[See](https://github.com/silverstripe/silverstripe-admin/issues/1548)~~   
## Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2915


client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -230,13 +230,7 @@ jQuery.entwine('ss', ($) => {
       const { url, ...attrs } = this.getOriginalAttributes();
       const fileSelected = attrs.hasOwnProperty('ID') && attrs.ID !== null;
       const folderId = this.getFolderId();
-      const editor = this.getElement().getEditor();
-      const selection = editor.getInstance().selection;
-      const selectionContent = editor.getSelection();
-      const tagName = selection.getNode().tagName;
-      // Unsupported media insertion will use insert link form instead
```
ChloeSartorelli: We need to keep this comment - it gives context about why we even have a `requireLinkText` variable in a js file that doesn't seem to be related to links

client/src/entwine/TinyMCE_ssmedia.js
```diff
@@ -230,13 +230,7 @@ jQuery.entwine('ss', ($) => {
...
-      // treat image tag selection as blank content
-      const requireLinkText = tagName !== 'A' && (tagName === 'IMG' || selectionContent.trim() === '');
+      const requireLinkText = this.getRequireLinkText();
```
ChloeSartorelli: This comment, and the `tagName === 'IMG'` portion of this logic, are both important.

If you add an image, and then try to replace that image with a pdf, the "Link text" field should be present - but this PR removes that.

Before this PR:
[Screencast from 25-01-24 15:40:50.webm](https://github.com/silverstripe/silverstripe-asset-admin/assets/36352093/ca0ac7a1-f018-4aa2-845d-1a5a2cf49007)

With this PR:
[Screencast from 25-01-24 15:43:05.webm](https://github.com/silverstripe/silverstripe-asset-admin/assets/36352093/1e0c7db3-bdfe-4812-b382-c60bfd168135)




Also note that the `getRequireLinkText()` function is defined for elements with the CSS selector `'.insert-link__dialog-wrapper'` in https://github.com/silverstripe/silverstripe-admin/pull/1579/. If the function exists to be called for this element, it's purely coincidence and shouldn't be relied upon.

ChloeSartorelli (changes requested):
The fix for the bug reported in the issue is fine - but the stuff you've added for the required link text doesn't work as expected. Details below.

client/src/entwine/TinyMCE_ssmedia.js
sabina-talipova: Updated

client/src/entwine/TinyMCE_ssmedia.js
sabina-talipova: Updated. Cancel changes.

ChloeSartorelli (approved):
LGTM

PR #1440, opened 2024-02-06
TLN Update translations

ChloeSartorelli:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)


emteknetnz (changes requested):
It's English instead of translations

tx-translator has "remove English strings from translation functionality", though possibly it's not working with javascript?

Issue #1209, opened 2021-06-14
UI inconsistency/unintuitive: having to click 'Details' to *edit* details (why not just load editable form in the first place?)


micschk:
Right now when you click the 'eye' icon on an uploadfield, a file selection/view modal will open. In this modal, the currently selected file/image is shown in sort-of-preview mode. That is, it is basically 1:1 the 'edit' formfields, but in read only mode. I've clicked maybe over a hundred times on one of the shown fields in order to edit some file detail, only to find I needed to click 'Details' first in order to deactivate read-only mode. 

I guess this initial read-only mode is meant as a sort of preview mode, but it looks 99% identical to the editable version. Besides the button not being very obvious UI-wise and the term 'Details' not indicating **edit me**, I cannot actually think of a reason why it would provide value to have read-only mode instead of just 'edit if you need to' mode (which is the mode when 'opening' files via the regular asset admin interface). It's basically just one extra step to click the edit button (two if you count the click on the read-only form field).

### I'd like to propose to just load the editable version by default and get rid of the read-only mode altogether.
This would coincidentally also fix/prevent some outstanding issues:
- #1136 
- [Cannot 'edit details' of file (UploadField > file selection modal)](https://github.com/silverstripe/silverstripe-assets/issues/445)

AND provides (what I'd consider) a better solution when also using FocusPoint module (currently this also looks 'clickable' but also requires 'edit details' first, which I find it *still* unintuitive & burdensome as clients keep mentioning it seems broken).
- #813  

<img width="880" alt="Screenshot 2021-06-14 at 12 24 07" src="https://user-images.githubusercontent.com/1005986/121878366-ebd6f680-cd0b-11eb-97ce-bcb02eb0f6ce.png">

lerni:
related to https://github.com/silverstripe/silverstripe-asset-admin/issues/1116 &  https://github.com/silverstripe/silverstripe-asset-admin/issues/1097

maxime-rainville:
I wish it was that simple. This modal is also used in many context where the user has to provide context information like when inserting an image in a WYSIWIG.

You also end up with somewhat ambiguous interaction .. if you click the "insert file" button in your upload field, you're would also be updating its data. You could also end up confusing some user who might not realise that they are actually updating the source file when changing that data.

micschk:
> This modal is also used in many context where the user has to provide context information like when inserting an image in a WYSIWIG.

[...]

Issue #1116, opened 2020-07-13
Rename "Details" button to "Edit Details"


purplespider:
_Initially reported in original PR: https://github.com/silverstripe/silverstripe-asset-admin/pull/1038_

New in Silverstripe 4.6 is the ability to edit file information directly in the modal.

When adjusting an image in an UploadField, the _Details_ fields are read-only, and the button to make them editable is confusing just labelled "Details":

![Screenshot 2020-06-02 at 09 43 37m@2x](https://user-images.githubusercontent.com/329880/83500600-e6b75880-a4b6-11ea-812a-46287ae5b9b3.png)

A primary use case of this button is when using the FocusPoint module.

---

The same button exists in a different context when editing a WYSIWYG image:

![Screenshot 2020-06-02 at 09 42 07@2x](https://user-images.githubusercontent.com/329880/83500385-9a6c1880-a4b6-11ea-9b44-6d33bc7c81f6.png)

Here, the Details fields are not visible, so changing it to just say "Edit" would be unclear.

---

Therefore I suggest renaming it to say **Edit Details** so that the action is clear in both contexts.

purplespider:
I should mention, that the ideal here is that when adjusting an UploadField image, the details fields are immediately editable, rather than read-only. Then we don't even need to display a button.

lerni:
To me click on a button to edit settings that belongs to a file makes sense and therefor just have fields like placement etc. editable is well OK. But "Details" doesn't make clear that you probable edit something that is used elsewhere. I agree "Edit Details" or even "Edit File Settings" would explain that better than just "Details".

micschk:
The '(Edit) Details' button isn't shown at all when this same modal gets loaded from an UploadField.
Only a cross is shown, without any label or title or any clue ("Cancel" aria-label, actually unlinks the current image) 🤦‍♂️.
Stunning UI 'improvements', all this used to work way more intuitive in a certain earlier versions...

lerni:
I've tried to reproduce what @micschk wrote with 4.8.x and 4.9.0-rc1  but either I do not understand or I cannot.

This button?
![Bildschirmfoto 2021-09-10 um 12 11 33](https://user-images.githubusercontent.com/472115/132839118-a1617b71-0273-42ff-b7ba-b1ee88ed2b91.png)


I wonder with what versions or other "Environment variables" like language etc. you used, to get that result?

**4.8.x**
```
silverstripe/framework (4.8.0)
[...]

PR #1446, opened 2024-03-12
TLN: change en.js word for "Details" to be "Edit"

sunnysideup:
Right now you see this when you select an image:

image

That does not make much sense, because you can already see the details and the button really is there to "EDIT" the image, as you can already see the details and because when you edit, you always edit "details"

image

For this purpose, I have added the english "phrase" for details now to be "edit".

### DIFF
--- a/client/lang/en.js
+++ b/client/lang/en.js
@@ -6,6 +6,7 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
   }
 } else {
   ss.i18n.addDictionary('en', {
+    'AssetAdmin.DETAILS': 'Edit',
     "AssetAdmin.ADD_FILES": "Add from files",
     "AssetAdmin.ADD_FOLDER_BUTTON": "Add folder",
     "AssetAdmin.BACK": "Back",


michalkleiner:
@sunnysideup is this addressing a particular issue? If there's one existing already, please link it, if not, please create one.

So far I can see there's https://github.com/silverstripe/silverstripe-asset-admin/issues/1209 and https://github.com/silverstripe/silverstripe-asset-admin/issues/1116 which both have relevant comments and info.

If we change the label, we probably also should change the translation key as we are changing the meaning so we need to let other languages know to re-translate it rather than just rely on the old meaning which may not be correct anymore.

sunnysideup:
Will  do.

sunnysideup:
Putting my management hat on here for a second.  If you look at the amount of discussion and the amount of time it has taken for this issue to be addressed then I think the ticket system is broken as a whole.  Or perhaps the ticket creators like me are just too much in rush to do their due diligence / deliver a pull request properly. 

In my books, it would be easier for someone who is "trusted" just to take up the issue, read the comments, make a decision, fix it (if required), and move on. 

This sort of thing should be picked up by a more design conscious person and just addressed as such. It could simply be a pencil button rather than "edit" OR "details".

ChloeSartorelli:
Sounds like there's a lot more to this than will be covered by this PR.
I'm going to close this PR, but I'll also highlight this issue internally as something we should prioritise since it has caused so much confusion.

sunnysideup:
@GuySartorelli  - does it matter that `AssetAdmin.DETAILS` is not currently in the `en` dictionary?

ChloeSartorelli:
Yes, that's a bug. Unfortunately the text collector doesn't currently collect text from JS so those have to be updated manually.
Feel free to raise an issue for it, with a linked PR if you're up for it.

ChloeSartorelli:
For clarity: You should be able to apply project-level override without it being added.

Issue #1465, opened 2024-06-05
Babel is devDependency


emteknetnz:
Babel is currently in `dependencies` in package.json, it should be in `devDependencies`

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1466

ChloeSartorelli:
PR merged

PR #1466, opened 2024-06-05
DEP Move babel to devDependencies

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1465


ChloeSartorelli (approved):
There's a lot of unrelated changes in yarn.lock but given we don't manually touch that file, and CI is happy, it's probably fine.
LGTM.

PR #1475, opened 2024-07-11
ENH Remove animation for thumbnails by default.

ChloeSartorelli:
Needs https://github.com/silverstripe/silverstripe-assets/pull/621 for CI to go green.

## Issue
- https://github.com/silverstripe/silverstripe-assets/issues/619


code/Model/ThumbnailGenerator.php
```diff
@@ -107,6 +110,14 @@ public function generateThumbnail(AssetContainer $file, $width, $height)
             $file = $file->existingOnly();
         }
 
+        // Try to remove the animation if we can
+        if (!$this->getAllowsAnimation() && $file->getIsAnimated() && ClassInfo::hasMethod($file, 'RemoveAnimation')) {
```
ChloeSartorelli: Note it doesn't make sense to add the `RemoveAnimation` method to `AssetContainer` interface, because it's about image manipulation.
Even though that interface is only used by `DBFile` and `File` which both use the `ImageManipulation` trait, we shouldn't assume the object has this method for the sake of strict correctness. (even though below we assume it has whatever `$method` is lol)

emteknetnz (changes requested):
There might be some future issue with copyright with that sample image provided, swap it to something more "sample image" like

ChloeSartorelli:
Swapped sample gif to one we explicitly own

code/Model/ThumbnailGenerator.php
```diff
@@ -107,9 +111,33 @@ public function generateThumbnail(AssetContainer $file, $width, $height)
...
+                $backend = $file->getImageBackend();
+                $origAllowAnimation = $backend->getAllowsAnimationInManipulations();
+                $backend->setAllowsAnimationInManipulations(false);
+            } elseif ($file->getIsAnimated() && ClassInfo::hasMethod($file, 'RemoveAnimation')) {
+                $noAnimation = $file->RemoveAnimation();
+                if ($noAnimation) {
+                    $file = $noAnimation;
+                }
+            }
+        }
```
ChloeSartorelli, line 128: See https://github.com/silverstripe/silverstripe-asset-admin/pull/1475#discussion_r1673291221 - this is now more complex logic, but that principle remains that we should adhere to the API of the `AssetContainer` interface. As such I've provided two ways to do this - one where there's an `Image_Backend` available and one where that's not available but `RemoveAnimation()` is.

emteknetnz (approved):

PR #1378, opened 2023-08-08
ENH Use archive text when file archiving is enabled

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/300

AssetAdmin::getClientConfig() is an array of PHP values that are made available in JS

This pull-request is NOT updating jest tests, instead they will be updated in https://github.com/silverstripe/silverstripe-asset-admin/pull/1379 instead

I'm doing this because I want to test https://github.com/silverstripe/gha-merge-up/pull/6 going cross major, though jest tests will definately cause merge-conflicts due to the jest tests being rewritten in `2` https://github.com/silverstripe/silverstripe-asset-admin/pull/1355

After merging - manually run the merge-up action on asset-admin using the default branch - https://github.com/silverstripe/silverstripe-asset-admin/actions/workflows/merge-up.yml

PR #1477, opened 2024-07-22
API Remove GraphQL

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799


client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -334,70 +331,45 @@ class AssetAdmin extends Component {
...
-      .then((resultItems) => {
-        const successes = resultItems.filter((result) => result).length;
-        const { archiveFiles } = this.props.sectionConfig;
-        if (successes !== ids.length) {
-          let transKey = 'AssetAdmin.BULK_ACTIONS_DELETE_FAIL_02';
-          let transDefault = '%s folders/files were successfully deleted, but %s files were not able to be deleted.';
-          if (archiveFiles) {
-            transKey = 'AssetAdmin.BULK_ACTIONS_ARCHIVE_FAIL_02';
-            transDefault = '%s folders/files were successfully archived, but %s files were not able to be archived.';
-          }
```
emteknetnz, line 364: I've removed the functionality above where if some but not all of the files were delete/archived, then show a message.

There's a few reasons why:
- I'd like the non-read endpoints to simply return a 204 (no content) on success, and a 4** code without a message on failure.  As much as possible I'd like to not return "custom success/failure messages" as that's not moving towards standard endpoints. Instead we should rely on standard HTTP status codes as much as possible.
- There are now plenty of ways to trigger 4** status codes before any files are actually deleted e.g. validating ids passed are on files that exist, canDelete() checks on all files, etc. Passing those checked and then failing to delete the file seems very unlikely. 
- If something like a 500 was returned due to an exception being thrown in PHP, the deleted JS code wouldn't have triggered anyway (would use a .catch() block for that as we're using lib/Backend)

_config/config.yml
```diff
@@ -21,6 +21,9 @@ SilverStripe\Core\Injector\Injector:
   SilverStripe\AssetAdmin\Controller\AssetAdmin:
     properties:
       ThumbnailGenerator: '%$SilverStripe\AssetAdmin\Model\ThumbnailGenerator.assetadmin'
+  SilverStripe\AssetAdmin\Controller\AssetAdminOpen:
+    properties:
+      ThumbnailGenerator: '%$SilverStripe\AssetAdmin\Model\ThumbnailGenerator.assetadmin'
```
ChloeSartorelli: I assume this is to replace `SilverStripe\AssetAdmin\Model\ThumbnailGenerator.graphql`

Why doesn't it have `Generates: false`? That was there to prevent trying to generate new thumbnails when loading asset admin if the thumbnail generation criteria change, because that would massively slow down the CMS.

ChloeSartorelli: In the current CMS 5 implementation, we have the two thumbnail generators:
- `SilverStripe\AssetAdmin\Model\ThumbnailGenerator.assetadmin` is used for upload fields despite its name. It generates thumbnails because there aren't expected to be many loaded on any given page load.
- `SilverStripe\AssetAdmin\Model\ThumbnailGenerator.graphql` is used for asset admin. It doesn't generate thumbnails (other than on upload) because we expect lots of images on any given page load.

client/src/components/UploadField/UploadField.js
```diff
@@ -575,11 +575,6 @@ UploadField.propTypes = {
   disabled: PropTypes.bool,
   data: PropTypes.shape({
     files: PropTypes.arrayOf(fileShape),
-    createFileEndpoint: PropTypes.shape({
-      url: PropTypes.string.isRequired,
-      method: PropTypes.string.isRequired,
-      payloadFormat: PropTypes.string.isRequired,
-    }),
```
ChloeSartorelli, line 582: Add new `endpoints` object in its place, even if its type is just `PropTypes.object`

client/src/containers/AssetAdmin/AssetAdmin.js
ChloeSartorelli, line 1: Why are we doing `refetchFolder` a lot now, when we weren't before?

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -276,21 +340,25 @@ class AssetAdmin extends Component {
...
-            return response;
-          });
+            this.handleOpenFolder(this.getFolderId());
+          }
+        } else if ((action === 'action_save' || action === 'action_publish')
+          && this.getFolderId() !== response.record.parent.id) {
+          // If the file was moved, open the folder containing the file was moved to
+          this.handleBrowse(response.record.parent.id, response.record.id, null);
+        }
+        return response;
```
ChloeSartorelli, line 372: Why have the conditions changed here? It's probably super obvious with the full context of how everything is put together but I haven't got that context fully.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -332,72 +400,50 @@ class AssetAdmin extends Component {
...
     const fileIDs = files.map(file => file.id);
-    const parentId = this.props.folder ? this.props.folder.id : 0;
-
-    return this.props.actions.files.deleteFiles(fileIDs, parentId)
-      .then(({ data: { deleteFiles } }) => {
+    const folder = this.state.folder;
+    const parentId = folder ? folder.id : 0;
+
+    const url = this.props.sectionConfig.endpoints.archive.url;
+    return backend.post(url, {
```
ChloeSartorelli, line 418: Some endpoints have an explicit `.method` - I haven't gotten as far as looking at the PHP than handles that yet but just philosophically, should we always have a `.method` that we dynamically respect in js?
If not, why *do* some of them have a `.method` that needs to be respected like that?

client/src/containers/AssetAdmin/AssetAdmin.js
ChloeSartorelli: Excellent reasoning, thank you for explaining

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -408,64 +454,69 @@ class AssetAdmin extends Component {
...
-            i18n.inject(
-              i18n._t(
-                'AssetAdmin.BULK_OWNED_WARNING_REMAINING',
-                'And {count} other file(s)'
-              ),
-              { count: rest.length },
-            )
+    if (fileIDs.length === 0) {
+      throw new Error('No files selected for unpublishing');
+    }
```
ChloeSartorelli: Should probably add this to all of them if we're adding it to any.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -673,7 +725,13 @@ class AssetAdmin extends Component {
   }
 
   render() {
-    const { folder, folderId, query, getUrl, type, maxFiles, toolbarChildren, SearchComponent, BulkDeleteConfirmationComponent } = this.props;
+    const { folderId, query, getUrl, type, maxFiles, toolbarChildren, SearchComponent, BulkDeleteConfirmationComponent } = this.props;
+
+    // return <h1>{ JSON.stringify(this.state.folder) }</h1>;
```
ChloeSartorelli: I assume this was debug code?
```suggestion
```

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -673,7 +725,13 @@ class AssetAdmin extends Component {
...
+
+    if (this.state.folder === null) {
+      return null;
+    }
```
ChloeSartorelli, line 756: What scenario does this happen in? This wasn't here before, why do we need it now?

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -704,7 +764,7 @@ class AssetAdmin extends Component {
           showBackButton={showBackButton}
           onBackButtonClick={this.handleBackButtonClick}
         >
-          <AssetAdminBreadcrumb {...breadcrumbProps} />
+          {folder && <AssetAdminBreadcrumb {...breadcrumbProps} />}
```
ChloeSartorelli, line 789: Why do we now need this condition?

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -738,7 +801,6 @@ AssetAdmin.propTypes = {
   onBrowse: PropTypes.func,
   onReplaceUrl: PropTypes.func,
   onInsertMany: PropTypes.func,
-  graphQLErrors: PropTypes.arrayOf(PropTypes.string),
```
ChloeSartorelli, line 741: I think you removed other things too? Like `folder` looks like it's in state now? Please make sure to update the propTypes.

This goes for _all_ components in _all_ PRs.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -812,9 +874,4 @@ export { AssetAdmin as Component };
 
 export default compose(
   connect(mapStateToProps, mapDispatchToProps),
-  injectGraphql('ReadFilesQuery'),
-  deleteFilesMutation,
-  unpublishFilesMutation,
-  publishFilesMutation,
-  withApollo
```
ChloeSartorelli, line 819: Love to see this

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
```diff
@@ -1,28 +1,47 @@
...
 import fileShape from 'lib/fileShape';
+import backend from 'lib/Backend';
+import Config from 'lib/Config';
 import DeletionModal from './DeletionModal';
 import BulkDeleteMessage from './BulkDeleteMessage';
-import { getFolderDescendantFileTotals, getFileTotalItems } from './helpers';
 
-/**
- * Wires the Redux store and Apollo result set with the DeletionModal.
- */
```
ChloeSartorelli: If it doesn't do that, what _does_ it do? Please update the comment instead of deleting it.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
         ]);
     }
 
+    public function apiReadDescendantCounts(HTTPRequest $request): HTTPResponse
+    {
+        $data = [];
+        $ids = $this->getGetDataValue($request, 'ids');
+        foreach ($ids as $id) {
+            $id = (int) $id;
+            /* @var Folder&AssetAdminFile $file */
```
ChloeSartorelli:
```suggestion
            /* @var File&AssetAdminFile $file */
```
At this point we don't know it's a folder, hence the `is_a()` check later on.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+            }
+        }
+        $this->extend('updateApiReadLiveOwnerCounts', $data, $request);
+        return $this->jsonSuccess(200, $data);
+    }
+
+    public function apiReadUsage(HTTPRequest $request): HTTPResponse
+    {
+        $id = (int) $request->param('ID');
+        /* @var Folder&AssetAdminFile $file */
```
ChloeSartorelli:
```suggestion
        /* @var File&AssetAdminFile $file */
```
Given the variable is named `$file` I assume we don't know this is a folder.

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
```diff
@@ -101,13 +118,10 @@ const ConnectedModal = compose(
     ['Loading'],
     (Loading) => ({ LoadingComponent: Loading }),
   ),
-  injectGraphql('readDescendantFileCountsQuery'),
-  withApollo
 )(BulkDeleteConfirmation);
 
 /**
  * Decide whether to render the Deletion Modal based on the information in the redux store.
- * This avoid firing off a GraphQL query for nothing.
```
ChloeSartorelli, line 110: Does it avoid firing off an XHR request for nothing?
If so just update instead of remove the comment. If not, are we introducing a regression?

client/src/containers/Editor/Editor.js
```diff
@@ -40,9 +41,32 @@ class Editor extends Component {
...
 
+  componentDidMount() {
+    this.refetchFile();
+  }
+
+  componentDidUpdate(prevProps) {
+    if ((prevProps.fileId !== this.props.fileId) && prevProps.fileId !== null) {
+      this.refetchFile();
+    }
+  }
```
ChloeSartorelli, line 56: Probably the same reason as above (if so just link to response) - but why are we refetching data that we weren't before?

client/src/containers/Editor/Editor.js
```diff
@@ -285,13 +314,6 @@ class Editor extends Component {
     const campaignTitle = i18n._t('Admin.ADD_TO_CAMPAIGN', 'Add to campaign');
     const Loading = this.props.loadingComponent;
 
-    // Most of the the time, the GraphQL data comes back first and the selected file data gets
-    // passed down to the Editor in time for the EditorHeader to be rendered correctly.
-    // Occasionnaly the FormSchema comes back first and the EditorHeader gets rendered without the
-    // necessary file data. Passing `file` to FormBuilderLoader will force a re-render when the
-    // GraphQL file data is filled in later on, which will cause a re-render of the EditorHeader.
-    const { file } = this.props;
```
ChloeSartorelli, line 293: We're still passing the file in... is that because this is still relevant (just for xhr instead of graphql)? If so, please keep the comment and just update it to match.

If that's not why we're passing file through now, why _are_ we still passing it through?

client/src/containers/Editor/Editor.js
```diff
@@ -323,7 +345,7 @@ class Editor extends Component {
 }
 
 Editor.propTypes = {
-  file: fileShape,
+  // file: fileShape,
```
ChloeSartorelli: Don't keep commented-out code

client/src/containers/Gallery/Gallery.js
```diff
@@ -870,23 +876,18 @@ class Gallery extends Component {
...
           <div className="gallery__error flexbox-area-grow">
             <div className="gallery__error-message">
               <h3>
                 {i18n._t('AssetAdmin.DROPZONE_RESPONSE_ERROR', 'Server responded with an error.')}
               </h3>
               {errorMessage && <p>{errorMessage}</p>}
-              {hasGraphQLErrors && graphQLErrors.map((error, index) => (
-                // eslint-disable-next-line react/no-array-index-key
-                <p key={index}>{error}</p>
-              ))}
```
ChloeSartorelli, line 889: Does `errorMessage` contain the errors from the controller which used to be in `graphQLErrors`? If not, what happens with errors now?

code/Controller/AssetAdmin.php
```diff
@@ -65,19 +70,52 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
+        'POST api/move' => 'apiMove',
+        'POST api/publish' => 'apiPublish',
+        'POST api/unpublish' => 'apiUnpublish',
+        'POST api/uploadFile' => 'apiUploadFile',
     ];
 
+    /**
+     * @var array
+     */
+    private static $allowed_actions = array(
```
ChloeSartorelli:
```suggestion
    private static array $allowed_actions = [
```

code/Controller/AssetAdmin.php
```diff
@@ -65,19 +70,52 @@ class AssetAdmin extends LeftAndMain implements PermissionProvider
...
+        'apiArchive',
+        'apiHistory',
+        'apiReadDescendantCounts',
+        'apiReadLiveOwnerCounts',
+        'apiReadUsage',
+        'apiMove',
+        'apiPublish',
+        'apiUnpublish',
+        'apiUploadFile',
+    );
```
ChloeSartorelli:
```suggestion
    ];
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+
+    /**
+     * Do not show this controller in the CMS menu
+     */
+    private static $ignore_menuitem = true;
+
+    /**
+     * @var ThumbnailGenerator
+     */
+    protected $thumbnailGenerator;
```
ChloeSartorelli:
```suggestion
    protected ThumbnailGenerator $thumbnailGenerator;
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+     *
+     * This must be false, not 'CMS_ACCESS' or else this will not work with silverstripe/subsites
+     * for users with limited permissions e.g. 'CMS_ACCESS_AssetAdmin' only
+     */
+    private static $required_permission_codes = false;
+
+    /**
+     * Do not show this controller in the CMS menu
+     */
+    private static $ignore_menuitem = true;
```
ChloeSartorelli:
```suggestion
    private static bool $required_permission_codes = false;

    /**
     * Do not show this controller in the CMS menu
     */
    private static bool $ignore_menuitem = true;
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+        Requirements::javascript('silverstripe/asset-admin:client/dist/js/bundle.js');
+        Requirements::css('silverstripe/asset-admin:client/dist/styles/bundle.css');
+    }
+
+    // public function canAccess(Member $member = null)
+    // {
+    //     // This method needs to be defined and return true to prevent LeftAndMainSubsites::canAccess()
+    //     // from being called and blocking access if the silverstripe/subsites modules is installed
+    //     return true;
+    // }
```
ChloeSartorelli:
```suggestion
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+     */
+    private static int $page_length = 50;
+
+    /**
+     * Allow any authenticated CMS user access to this controller
+     *
+     * This must be false, not 'CMS_ACCESS' or else this will not work with silverstripe/subsites
+     * for users with limited permissions e.g. 'CMS_ACCESS_AssetAdmin' only
+     */
+    private static $required_permission_codes = false;
```
ChloeSartorelli: I assume you've validated that unauthenticated users can't use the `api/read` endpoint to read files?

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+    // }
+
+    public function getClientConfig()
+    {
+        $baseLink = $this->Link();
+        return array_merge(parent::getClientConfig(), [
+            'reactRouter' => true,
+            'endpoints' => [
+                'read' => [
+                    'url' => Controller::join_links($baseLink, 'api/read'),
```
ChloeSartorelli:
```suggestion
                    'url' => $this->Link('api/read'),
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+    {
+        $id = $request->param('ID');
+        // When passing in an ID of 0, Silverstripe routing will have an ID value of
+        // "read", presumbably because 0 is falsey.
+        // Unlike most endpoints, we want to allow 0 as a valid ID here.
+        if ($id === 'read') {
+            $id = '0';
+        }
+        if (!ctype_digit($id)) {
+            $this->jsonError(404);
```
ChloeSartorelli, line 323:
```suggestion
            $this->jsonError(400);
```
If the ID isn't an integer, the client has sent a bad request.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    public function apiRead(HTTPRequest $request): HTTPResponse
+    {
+        $id = $this->getIDParam($request);
+        if ($id === 0) {
+            // This is required when clicking the "Create new folder" button in the CMS
+            $data = $this->getFileData(Folder::singleton(), $request, false, false);
+            $tmp = $this->getFileData(null, $request, false, true);
+            $data['children'] = $tmp['children'];
```
ChloeSartorelli: Why are we creating a new folder in the "Read" API?

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+
+    private function getIDParam(HTTPRequest $request): int
+    {
+        $id = $request->param('ID');
+        // When passing in an ID of 0, Silverstripe routing will have an ID value of
+        // "read", presumbably because 0 is falsey.
+        // Unlike most endpoints, we want to allow 0 as a valid ID here.
+        if ($id === 'read') {
+            $id = '0';
+        }
```
ChloeSartorelli, line 321: We should probably follow the pattern that's used elsewhere in the CMS, where either
1. `new` is used to indicate we're creating a new record (gridfield does it this way, so that makes it the pattern for literally all admins other than CMSMain), or
2. There is no ID value passed, and we know because of the context of the request that we're creating a new item (CMSMain does it this way)

Using an ID of `0` means if the JS breaks somehow and ends up passing through a falsy value of any sort because maybe it can't read the ID or something, we'll end up creating a new item.

That is assuming that's what the `0` is used for, if it's for something else please make it clear in the comment what it's actually used for.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+        if ($id === 0) {
+            // This is required when clicking the "Create new folder" button in the CMS
+            $data = $this->getFileData(Folder::singleton(), $request, false, false);
+            $tmp = $this->getFileData(null, $request, false, true);
+            $data['children'] = $tmp['children'];
+        } else {
+            $file = $this->getFileByID($id);
+            $data = $this->getFileData($file, $request, true, true);
+        }
+        $this->extend('updateApiRead', $data, $request);
```
ChloeSartorelli:
```suggestion
        $this->extend('updateApiRead', $data, $file ?? null, $request);
```
Save people the effort of re-fetching the file (presumably if they're updating the data, they're updating it for that specific file)

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+            'version' => $file->Version,
+            'isTrackedFormUpload' => $file->isTrackedFormUpload(),
+        ];
+    }
+
+    protected function getFileByID(int $id, bool $isFolder = false, bool $use404 = true): ?File
+    {
+        if ($id === 0) {
+            $this->jsonError($use404 ? 404 : 400);
+        }
```
ChloeSartorelli: An ID of 0 should _always_ result in a 400 error, because 0 is never a valid ID for a record which means it's a bad request by definition.

```suggestion
        if ($id === 0) {
            $this->jsonError(400);
        }
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    protected function getFileData(
+        ?File $file,
+        HTTPRequest $request,
+        bool $includeParents,
+        bool $includeChildren,
+    ): array {
+        // Parents - used for breadcrumbs
+        $parents = [];
```
ChloeSartorelli:
```suggestion
        bool $includeAncestors,
        bool $includeChildren,
    ): array {
        // Ancestors - used for breadcrumbs
        $ancestors = [];
```

It's not just the parent of the file, it's all ancestors. This tends to be the language used elsewhere, e.g. see the `Hierarchy` extension in framework.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+            while ($parent = $nextFile->Parent()) {
+                // Prevent infinite loops
+                if (!$parent->ID || in_array($parent->ID, $parentIDs)) {
+                    break;
+                }
+                $parentIDs[] = $parent->ID;
+                $parents[] = $this->getFileData($parent, $request, false, false);
+                $nextFile = $parent;
+            }
+            $parents = array_reverse($parents);
```
ChloeSartorelli:
```suggestion
            $ancestors = $file->getAncestors();
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+        }
+        // Children
+        $children = [];
+        $totalCount = 0;
+        if ($includeChildren) {
```
ChloeSartorelli, line 152: I haven't looked at all of the code below - but it seems likely you can use the `Children()`, `AllChildren()`, or `getDescendantIDList()` method here, and use regular ORM filter, limit/offset, and sort functionality.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+        $childObj->pageInfo->totalCount = $totalCount;
+        $childObj->nodes = $children;
+        // This is purely for the root folder
+        if (!$file) {
+            return [
+                'children' => $childObj
+            ];
+        }
+        // Values
+        /** @var File|Versioned $file */
```
ChloeSartorelli: Probably better to just put `@param File|Versioned $file` in the PHPDoc for the method instead of putting this way down here. Otherwise if we want versioning methods higher up in the method in the future we'd have to move the `@var`.

code/Controller/AssetAdminOpen.php
ChloeSartorelli, line 1: Please add PHPDoc comments to all public and protected methods in this class which aren't inherited from superclasses.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+        return [
+            'canDelete' => $file->canDelete(),
+            'canEdit' => $file->canEdit(),
+            'canView' => $file->canView(),
+            'category' => $file instanceof Folder ? 'folder' : $file->appCategory(),
```
ChloeSartorelli, line 207: "folder" isn't a category - what's that for?
We already have the type as folder further down, so this should probably be null or an empty string for non-files.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+            'title' => $file->Title,
+            'type' => $file instanceof Folder ? 'folder' : 'file',
+            'url' => $file->getURL(),
+            'visibility' => $file->getVisibility(),
+            'hasRestrictedAccess' => $file->hasRestrictedAccess(),
+            'children' => $childObj,
+            'parents' => $parents,
+            'draft' => $file->isOnDraftOnly(),
+            'extension' => $file->getExtension(),
+            'published' => $file->isPublished(),
```
ChloeSartorelli, line 223: Why do we have both `draft` and `published`? Surely it's only ever one or the other?
Should we have a "versionState" instead, [like we did for linkfield](https://github.com/silverstripe/silverstripe-linkfield/blob/ed8455648d9e2612115a30fe8d6c211c4e7f69f0/src/Models/Link.php#L170-L187)? Maybe even centralise that logic so we can reuse it anywhere we need the version state?

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+        if (!$file->canView()) {
+            $this->jsonError(403);
+        }
+        return $file;
+    }
+
+    protected function getFolderByID(int $id, bool $use404 = true): ?Folder
+    {
+        return $this->getFileByID($id, true, $use404);
+    }
```
ChloeSartorelli: This seems completely superfluous. Just call `getFileByID()` directly instead.
```suggestion
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+            'modified' => $file->isModifiedOnDraft(),
+            'size' => $file->getSize(),
+            'smallThumbnail' => $this->getSmallThumbnail($file),
+            'thumbnail' => $this->getThumbnail($file),
+            'version' => $file->Version,
+            'isTrackedFormUpload' => $file->isTrackedFormUpload(),
+        ];
+    }
+
+    protected function getFileByID(int $id, bool $isFolder = false, bool $use404 = true): ?File
```
ChloeSartorelli: PHPDoc should explain why we'd ever want `$use404` to be false

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+        }
+        return $file;
+    }
+
+    protected function getFolderByID(int $id, bool $use404 = true): ?Folder
+    {
+        return $this->getFileByID($id, true, $use404);
+    }
+
+    protected function getGetDataValue(HTTPRequest $request, string $key, bool $allowMissing = false): mixed
```
ChloeSartorelli:
```suggestion
    protected function getDataValue(HTTPRequest $request, string $key, bool $allowMissing = false): mixed
```
I don't think we need to distinguish that this is for a GET request here. `getGet` looks like a typo so it's doing more harm than good.

ChloeSartorelli: Maybe `getValueFromRequest` is better. Much clearer what's happening, imo. And then if you really want the fact that it's a GET request specifically to be in there, you can have `getValueFromGetRequest`

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+    {
+        $getVars = $request->getVars();
+        if (!array_key_exists($key, $getVars)) {
+            if ($allowMissing) {
+                return null;
+            }
+            $this->jsonError(404);
```
ChloeSartorelli, line 265:
```suggestion
            $this->jsonError(400);
```
404 is "we couldn't find the data on _our_ end". In this case the request didn't send the data we need, so it's a bad request.

ChloeSartorelli: Same with the other errors in this method - in all of them it's the request that didn't send the necessary data (or send it in the wrong format), so it's a bad request.


We haven't even started _looking_ for records on our end yet so it can't be a 404.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    private function getSmallThumbnail(File $file): ?string
+    {
+        $width = UploadField::config()->uninherited('thumbnail_width');
+        $height = UploadField::config()->uninherited('thumbnail_height');
+        return $this->getThumbnailGenerator()->generateThumbnailLink($file, $width, $height);
+    }
+
+    private function getThumbnail(File $file): ?string
```
ChloeSartorelli:
```suggestion
    private function getSmallThumbnailUrl(File $file): ?string
    {
        $width = UploadField::config()->uninherited('thumbnail_width');
        $height = UploadField::config()->uninherited('thumbnail_height');
        return $this->getThumbnailGenerator()->generateThumbnailLink($file, $width, $height);
    }

    private function getThumbnailUrl(File $file): ?string
```
We're getting a URL, not the actual thumbmnail itself.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    private function filterFiles(DataList $list, array $filter): DataList
+    {
+        // ID filtering
+        if (isset($filter['id']) && (int)$filter['id'] > 0) {
+            $list = $list->filter('ID', $filter['id']);
+
+            if ($list->count() === 0) {
+                $this->httpError(400);
```
ChloeSartorelli:
```suggestion
                $this->httpError(404);
```
404 - we looked for the records and we couldn't find them. The request itself was fine.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+            }
+        } elseif (isset($filter['id']) && (int)$filter['id'] === 0) {
+            // Special case for root folder
+            $list = new ArrayList([new Folder([
```
ChloeSartorelli:
```suggestion
            $list = ArrayList::create([Folder::create([
```

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+    private function filterFiles(DataList $list, array $filter): DataList
+    {
+        // ID filtering
+        if (isset($filter['id']) && (int)$filter['id'] > 0) {
+            $list = $list->filter('ID', $filter['id']);
+
+            if ($list->count() === 0) {
+                $this->httpError(400);
+            }
+        } elseif (isset($filter['id']) && (int)$filter['id'] === 0) {
```
ChloeSartorelli: What about `if (isset($filter['id']) && $filter['id'] < 0)` or is otherwise not coercable into an int? That should throw a 400.

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,418 @@
...
+        }
+        // Filter unknown id by known child if search is not applied
+        if (!$search && isset($filter['anyChildId'])) {
+            $child = File::get()->byID($filter['anyChildId']);
+            $id = $child ? ($child->ParentID ?: 0) : 0;
+            if ($id) {
+                $list = $list->filter('ID', $id);
+            } else {
+                // Special case for root folder, since filter by ID = 0 will return an empty list
+                $list = new ArrayList([new Folder([
```
ChloeSartorelli:
```suggestion
            $fromDate = DateField::create(null, null, $filter['lastEditedFrom']);
            $list = $list->filter("LastEdited:GreaterThanOrEqual", $fromDate->dataValue().' 00:00:00');
            $search = true;
        }
        if (!empty($filter['lastEditedTo'])) {
            $toDate = DateField::create(null, null, $filter['lastEditedTo']);
            $list = $list->filter("LastEdited:LessThanOrEqual", $toDate->dataValue().' 23:59:59');
            $search = true;
        }
        // Date filtering created
        if (!empty($filter['createdFrom'])) {
            $fromDate = DateField::create(null, null, $filter['createdFrom']);
            $list = $list->filter("Created:GreaterThanOrEqual", $fromDate->dataValue().' 00:00:00');
            $search = true;
        }
        if (!empty($filter['createdTo'])) {
            $toDate = DateField::create(null, null, $filter['createdTo']);
            $list = $list->filter("Created:LessThanOrEqual", $toDate->dataValue().' 23:59:59');
            $search = true;
        }
        // Categories (mapped to extensions through the enum type automatically)
        if (!empty($filter['appCategory'])) {
            $list = $list->filter('Name:EndsWith', $filter['appCategory']);
            $search = true;
        }
        // Filter unknown id by known child if search is not applied
        if (!$search && isset($filter['anyChildId'])) {
            $child = File::get()->byID($filter['anyChildId']);
            $id = $child ? ($child->ParentID ?: 0) : 0;
            if ($id) {
                $list = $list->filter('ID', $id);
            } else {
                // Special case for root folder, since filter by ID = 0 will return an empty list
                $list = ArrayList::create([Folder::create([
```

ChloeSartorelli: Did you copy most of this from somewhere?

If so we can leave it more-or-less as is for now and refactor later. But if this is all new code we should just set up `serachable_fields` and use `SearchContext` instead now.

code/Controller/AssetAdmin.php
```diff
@@ -198,30 +209,69 @@ public function getClientConfig()
...
-            ],
+            'endpoints' => array_merge($parentClientConfig['endpoints'], [
+                'createFile' => [
+                    'url' => Controller::join_links($baseLink, 'api/createFile'),
+                    'method' => 'post',
+                    'payloadFormat' => 'urlencoded',
+                ],
+                'archive' => [
+                    'url' => Controller::join_links($baseLink, 'api/archive'),
+                    'method' => 'POST',
```
ChloeSartorelli: There's a mix between uppercase and lowercase for the method. We should make it consistent - looks like it used to be lowercase so presumably that's what all the other controllers are using too.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
         ]);
     }
 
+    public function apiReadDescendantCounts(HTTPRequest $request): HTTPResponse
+    {
+        $data = [];
+        $ids = $this->getGetDataValue($request, 'ids');
+        foreach ($ids as $id) {
+            $id = (int) $id;
```
ChloeSartorelli, line 336:
```suggestion
```
No need to explicitly cast - PHP is quite happy for us to pass string reprensentations of ints around even to methods with strong typing - it'll cast for us.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+    }
+
+    public function apiReadLiveOwnerCounts(HTTPRequest $request): HTTPResponse
+    {
+        $ids = $request->getVar('ids');
+        if (!is_array($ids)
+            || count(array_filter($ids, 'ctype_digit')) !== count($ids)
+            || count(array_filter($ids, fn($id) => (int) $id > 0)) !== count($ids)
+        ) {
+            $this->jsonError(404);
```
ChloeSartorelli:
```suggestion
            $this->jsonError(400);
```
As mentioned elsewhere, at this stage we haven't tried to find records so we can't say we didn't find them. It's just a bad request.

ChloeSartorelli: Shouldn't we be using `getGetDataValue()` here instead?

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+        $fileIDs = [];
+        foreach ($files as $file) {
+            if (is_a($file, Folder::class) || !$file->hasExtension(RecursivePublishable::class)) {
+                continue;
+            }
+            if (!$file->canView()) {
+                $this->jsonError(403);
+            }
+            $fileIDs[] = $file->ID;
+        }
```
ChloeSartorelli: Abstract this into a `getFilesByID()` method similar to the `getFileByID()` method in the `AssetAdminOpen` class.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+            $this->jsonError(404);
+        }
+        $files = File::get()->filter(['ID' => $ids]);
+        if ($files->count() !== count($ids)) {
+            $this->jsonError(404);
+        }
+        $data = [];
+        $fileIDs = [];
+        foreach ($files as $file) {
+            if (is_a($file, Folder::class) || !$file->hasExtension(RecursivePublishable::class)) {
```
ChloeSartorelli: > `is_a($file, Folder::class)`

Is it even valid to pass IDs of folders into this request, if we're ignoring them? Shouldn't we return a 400 here instead?

> `!$file->hasExtension(RecursivePublishable::class)`

Extensions are added by class, not by instance or record, so we can probably just check this up front and potentially save some db calls.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+            /** @var Versioned|RecursivePublishable|File $liveFile */
+            foreach ($liveFiles as $liveFile) {
+                $count = $liveFile->findOwners(false)->count();
+                // This translation key is from a class that has since been removed
+                $class = 'SilverStripe\\AssetAdmin\\GraphQL\\Resolvers\\PublicationResolver';
+                $data[] = [
+                    'id' => $liveFile->ID,
+                    'count' => $count,
+                    'message' => _t(
+                        "$class.OWNER_WARNING",
```
ChloeSartorelli:
```suggestion
                $data[] = [
                    'id' => $liveFile->ID,
                    'count' => $count,
                    'message' => _t(
                        __CLASS__ . '.OWNER_WARNING',
```
Just use a new key

code/Controller/AssetAdmin.php
```diff
@@ -522,6 +758,15 @@ public function apiHistory(HTTPRequest $request)
...
     }
 
+    private function getPostedJsonValue(HTTPRequest $request, string $key): mixed
+    {
+        $data = json_decode($request->getBody(), true);
+        if (!array_key_exists($key, $data)) {
+            $this->jsonError(400);
+        }
+        return $data[$key];
+    }
```
ChloeSartorelli: We're doing lots of repeated checks like the below after calling this method - just embed those checks into this method instead (mirroring `AssetAdminOpen::getGetDataValue()`)
```php
        if (!is_array($ids) || empty($ids)) {
            $this->jsonError(400);
        }
```

code/Controller/AssetAdminOpen.php
ChloeSartorelli: **UPDATE** Looks like you have `getPostedJsonValue` in `AssetAdmin` - so I recommend the following:

|old name|new name|
|---|---|
|getGetDataValue|getValueFromGetRequest|
|getPostedJsonValue|getValueFromPostRequest|

Then it's clear how these methods relate to each other, and it's really clear what they're doing.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+    public function apiArchive(HTTPRequest $request): HTTPResponse
+    {
+        $this->checkJsonCsrfToken($request);
+        $ids = $this->getPostedJsonValue($request, 'ids');
+        if (!is_array($ids) || empty($ids)) {
+            $this->jsonError(400);
+        }
+        $files = File::get()->filter(['ID' => $ids]);
+        if ($files->count() !== count($ids)) {
+            $this->jsonError(400);
```
ChloeSartorelli: We decided this was 400 instead of 404 because the data is in the post body rather than the "resource" AKA URL, right?

Probably worth adding a code comment explaining that, because I still look at this and go "but we didn't find it, so it's 404"

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+        if (!$file->canView()) {
+            $this->jsonError(403);
+        }
+        $count = $file->getFilesInUse()->count();
+        $data = ['count' => $count];
+        $this->extend('updateApiReadUsage', $data, $request);
+        return $this->jsonSuccess(200, $data);
+    }
+
+    public function apiArchive(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli: Should probably be `apiDelete` - versioning is technically an add-on which changes the delete behaviour, rather than being the default.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+                $this->jsonError(403);
+            }
+        }
+        foreach ($files as $file) {
+            if ($file->has_extension(Versioned::class)) {
+                $file->doArchive();
+            } else {
+                $file->delete();
+            }
+        }
```
ChloeSartorelli, line 429: I think looping through twice like this will run the query twice, won't it? Is there a way we can avoid that?

Same thing with all other methods you've added where that's happening.

code/Controller/AssetAdmin.php
ChloeSartorelli, line 1: Please add PHPDoc comments to all new public and protected methods in this class which aren't inherited from superclasses.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+        foreach ($files as $file) {
+            if (!$file->canEdit()) {
+                $this->jsonError(403);
+            }
+        }
+        foreach ($files as $file) {
+            $file->ParentID = $folder->ID;
+            $file->write();
+        }
+        return $this->jsonSuccess(204);
```
ChloeSartorelli, line 454: Why not 200?

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+            }
+        }
+        return $this->jsonSuccess(204);
+    }
+
+    public function apiMove(HTTPRequest $request): HTTPResponse
+    {
+        $this->checkJsonCsrfToken($request);
+        $ids = $this->getPostedJsonValue($request, 'ids');
+        $folderID = (int) $this->getPostedJsonValue($request, 'folderID');
```
ChloeSartorelli:
```suggestion
        $folderID = $this->getPostedJsonValue($request, 'folderID');
```
We don't need to cast it, PHP will do that for us when relevant.

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+        if (!is_array($ids) || empty($ids)) {
+            $this->jsonError(400);
+        }
+        $folder = $this->getFolderByID($folderID, false);
+        if (!$folder) {
+            $this->jsonError(400);
+        }
```
ChloeSartorelli: `getFileByID` already does this 
```suggestion
```

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+        foreach ($files as $file) {
+            if (!$file->canEdit()) {
+                $this->jsonError(403);
+            }
+        }
+        foreach ($files as $file) {
+            $file->ParentID = $folder->ID;
+            $file->write();
```
ChloeSartorelli, line 452: The GraphQL resolver did an explicit `$file->writeToStage(Versioned::DRAFT);` - why are we doing a full `write()` here?

code/Controller/AssetAdmin.php
```diff
@@ -273,6 +323,195 @@ public function getClientConfig()
...
+        }
+        return $this->jsonSuccess(204);
+    }
+
+    public function apiPublish(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 460: In at least one other place we're checking if files have the `PublishRecursive` extension - but here we're not. We should be consistent and either always check for it before relying on it, or always implicitly assuming it's there.

`silverstripe/versioned` isn't a dependency of this module so we should probably always check.

code/Controller/AssetAdmin.php
ChloeSartorelli: Same question applies for `apiPublish` and `apiUnpublish` and likely others.

code/Controller/AssetAdmin.php
```diff
@@ -287,7 +526,7 @@ public function apiCreateFile(HTTPRequest $request)
         $upload = $this->getUpload();
         $upload->setReplaceFile(false);
 
-        // CSRF check
+        // CSRF check - note that token is sent in POST body, not header
```
ChloeSartorelli, line 512: We should probably change that - but if that's not related to this work say so and I'll create a spin-off card.

code/Controller/AssetAdmin.php
```diff
@@ -355,20 +594,21 @@ public function apiCreateFile(HTTPRequest $request)
      * and 'Name' as form filename value
      * @return HTTPRequest|HTTPResponse
      */
-    public function apiUploadFile(HTTPRequest $request)
+    public function apiUploadFile(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 575: If we're adding return types, probably add to `apiCreateFile` as well, and any others I've not noticed are currently missing them.

code/Controller/AssetAdmin.php
```diff
@@ -355,20 +594,21 @@ public function apiCreateFile(HTTPRequest $request)
      * and 'Name' as form filename value
      * @return HTTPRequest|HTTPResponse
```
ChloeSartorelli:
```suggestion
```

code/Controller/AssetAdmin.php
```diff
@@ -355,20 +594,21 @@ public function apiCreateFile(HTTPRequest $request)
...
+    public function apiUploadFile(HTTPRequest $request): HTTPResponse
     {
         $data = $request->postVars();
 
         // When updating files, replace on conflict
         $upload = $this->getUpload();
         $upload->setReplaceFile(true);
-
-        // CSRF check
+
```
ChloeSartorelli:
```suggestion

```
Remove the horizontal space

tests/behat/features/manage-files.feature
```diff
@@ -97,8 +97,8 @@ Feature: Manage files
     Then I should see the "Form_fileEditForm" form
     When I press the "Other actions" button
       And I press the "Delete" button
-    Then I should see a modal titled "Confirm deletion"
-      And I press the Delete button inside the modal
+    Then I should see a modal titled "Confirm archive"
+      And I press the Archive button inside the modal
```
ChloeSartorelli: It should only say "archive" if `keep_archived_assets` is set to true. We did an exercise in https://github.com/silverstripe/silverstripe-asset-admin/pull/1378 to ensure "delete" and "archive" were used appropriately based on whether files are kept or not.

ChloeSartorelli (changes requested):

_config/config.yml
emteknetnz: Have set the assetadminopen one to false and added your comments as inline comments

client/src/components/UploadField/UploadField.js
emteknetnz: Done

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: Previously graphql was a HOC around AssetAdmin (readFilesQuery I believe) and it did the equivalent of refetchFolder a whole lot (though not enough, we have very longstanding bugs with Apollo caching). I don't actually know exactly what triggered it, that's part of the mystery that is the GraphQL HOC / Apollo.

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: From memory this was do to with editing an individual file/folder. I think I changed the logic around to handle both navigating to new folder after it's created, and also moving to the new file location after moving it. Something along those lines.  It was just to make it behave like it was previously under graphql

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: There was some old config that has .method on them, though honestly it's pretty pointless since it just makes the JS harder to read, and we set the required method in $url_handlers so you'll know very quickly if it's typed in wrong

Getting rid of the method would probably require refactoring some other JS code that does dynamically use those methods and that's very unrelated to these PRs. If we want to do that I think it should happen in a separate card

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: I don't follow? Add what to of all what?

Do you mean throwing Errors? There's plenty of methods doing that in this class

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: Removed

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: this.state.folder is set to null in the constructor.  It's set to non-null in refetchFolder(), though that's an XHR request to there's a brief period early on where this.state.folder is null.

Previously folder was a prop

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: Because folder (which is just this.state.folder) can be null

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: Updated

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
emteknetnz: Have added back in "This avoid firing off a GraphQL query for nothing.", as that's still true

client/src/containers/BulkDeleteConfirmation/BulkDeleteConfirmation.js
emteknetnz: Updated. It's still true expect for the Apollo bit which I've removed

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: Changed to lowercase

code/Controller/AssetAdmin.php
emteknetnz: It's also being cast as an int for the response json

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: Done

code/Controller/AssetAdmin.php
emteknetnz: We had this discussion offline the other day where I remember correctly we decided that data passed by query string, and hence the URL, for GET requests, which is the case here, is part of the HTTP "resource" and 404 would be correct response.

Whereas for POST request the data is passed in the request body and is not part of the "resource", so 400 is appropriate there.

I do think that's correct as broadly speaking, as for instance a CDN would cache a GET response making no differentiation between a path and a query string, it's all just part of the URL

I've been consistent in all of these PRs to have a convention of

GET / read:
Bad data = 404
Response = 200 with JSON body

POST / write
Bad data = 400
Response = 204 with empty body

emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: > Is it even valid to pass IDs of folders into this request, if we're ignoring them? Shouldn't we return a 400 here instead?

Yes it's valid, there's nothing stopping a user from selecting them in the file manager along with a bunch of regular file

> Extensions are added by class, not by instance or record, so we can probably just check this up front and potentially save some db calls.

Projects may have several subclasses of File, some of which do and some of which don't have the extension

code/Controller/AssetAdmin.php
emteknetnz: Done

code/Controller/AssetAdmin.php
emteknetnz: Done

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: I don't think we should be documenting why we use the correct HTTP status codes - that's already defined in elsewhere on the internet

code/Controller/AssetAdmin.php
emteknetnz: Probably will, though if we fail a permission check half-way through the list of files, looping twice prevents only some of the records being deleted. That would be an awkward situation, because then you'd need to display both a success and failure toast.

Given we're displaying paginated results and not deleting thousands of records at a time, I think the performance penalty as felt by the content editor is going to be minuscule

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: Updated

code/Controller/AssetAdmin.php
emteknetnz: From what I can tell there's no functional difference from calling `$file->writeToStage(Versioned::DRAFT)` and regular `$file->write()` other than a calling a couple of extension hooks `onBeforeWriteToStage` and `onAfterWriteToStage`, but those doesn't seem relevant to moving files.  

Maybe graphql has some weird "Live mode" or something so it needed to explicitly have things to draft mode?

I think it's much cleaner just using regular $file->write()

code/Controller/AssetAdmin.php
emteknetnz: 204 "No Content" mean success with an empty response body.  I've consistently used this as the success code for write requests

code/Controller/AssetAdmin.php
emteknetnz: That's a pretty broad topic, because we put SecurityID in as a hidden field all over the place :-)

I don't think this is worth changing at this stage honestly, if/when we go fully API driven that's would be good time to look at this

code/Controller/AssetAdmin.php
emteknetnz: I'd say assume it's always there, there's just so much UI in asset-admin that assumes things are Versioned there really zero support for unversioned files in practice, though who knows maybe things sort of work. Will remove the checks to see if a file has the RecursivePublishable extension

code/Controller/AssetAdmin.php
emteknetnz: Done. I've also removed the redundant `return null` underneath the $this->jsonError(400) as they're unreachable code and just make the response type look like it should be nullable when it doesn't need to be

code/Controller/AssetAdmin.php
emteknetnz: Done

client/src/containers/Editor/Editor.js
emteknetnz: Yeah basically the same reason, previous response was https://github.com/silverstripe/silverstripe-asset-admin/pull/1477/files#r1722712939

client/src/containers/Editor/Editor.js
emteknetnz: We're not passing in the file as a prop.  The component is constructed in AssetAdmin.js `<EditorComponent ...>` there's no file prop

client/src/containers/Editor/Editor.js
emteknetnz: Removed

client/src/containers/Gallery/Gallery.js
emteknetnz: errorMessage comes from redux - see mapStateToProps

code/Controller/AssetAdminOpen.php
emteknetnz: Done

code/Controller/AssetAdminOpen.php
emteknetnz: Yes, class extends LeftAndMain which enforces authentication

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: Removed

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: Comment seems wrong, have changed to "The root folder has an ID of 0"

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: Updated to use $file->getAncestors(), though kept the bit where we call `$this->getFileData($parent, $request, false, false);` because we use that to build the returned json

code/Controller/AssetAdminOpen.php
emteknetnz: I don't think it's that simple due to nuanced way the 3 different scenarios need to behave, and that Folders need to be sorted before files. I don't think we'd reduce much code even if we used those method. I'd rather not alter this right now.

code/Controller/AssetAdminOpen.php
emteknetnz: I don't like the idea of putting having solitary docblock `@params`, it seems like it goes against how we're generally dealing with adding strong types.  These inline `@vars` are only to support type hinting, and they're easy enough to move up higher if needed in the future

code/Controller/AssetAdminOpen.php
emteknetnz: There's bunch of existing JS logic that looks at `category === 'folder'` including in components not included in this PR. Refactoring that is out of scope

code/Controller/AssetAdminOpen.php
emteknetnz: There's existing JS logic for things like versioned state badges that relies on this. I don't want to refactor that in this PR

code/Controller/AssetAdminOpen.php
emteknetnz: Have changed so that the param is now required and is $missingFileErrorCode - reads a lot better when calling because the 404 or 400 is an argument e.g. `$file = $this->getFileByID($id, false, 404);`

code/Controller/AssetAdminOpen.php
emteknetnz: If it's a GET request then it should be a 404 - the ID is part of the URL/resource

code/Controller/AssetAdminOpen.php
emteknetnz: Removed

code/Controller/AssetAdminOpen.php
emteknetnz: Have changed to `getQueryStringValue()` instead

code/Controller/AssetAdminOpen.php
emteknetnz: As mentioned in other comments, 404 because it's a GET request and the query string is part of the URL/resource

code/Controller/AssetAdminOpen.php
emteknetnz: 0 is the ID used for the rootFolder

code/Controller/AssetAdminOpen.php
emteknetnz: Using 404 as the general error code for GET requests

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: Have added ctype_digit() check above

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: Updated. Yes copied from elsewhere

tests/behat/features/manage-files.feature
emteknetnz: Updated

_config/config.yml
```diff
@@ -14,13 +14,27 @@ SilverStripe\Admin\Forms\UsedOnTable:
...
+  SilverStripe\AssetAdmin\Model\ThumbnailGenerator.assetadminopen:
+    class: SilverStripe\AssetAdmin\Model\ThumbnailGenerator
+    properties:
+      Generates: false
   SilverStripe\AssetAdmin\Controller\AssetAdmin:
     properties:
       ThumbnailGenerator: '%$SilverStripe\AssetAdmin\Model\ThumbnailGenerator.assetadmin'
+  SilverStripe\AssetAdmin\Controller\AssetAdminOpen:
+    properties:
+      ThumbnailGenerator: '%$SilverStripe\AssetAdmin\Model\ThumbnailGenerator.assetadminopen'
```
ChloeSartorelli, line 41: Since AssetAdmin is a subclass of AssetAdminOpen, this will result in overriding the thumbnail generator for AssetAdmin.

I haven't tested this yet, but it seems like that would mean a bunch of operations that shouldn't be creating thumbnails now will be?

client/src/components/UploadField/UploadField.js
```diff
@@ -590,6 +585,7 @@ UploadField.propTypes = {
   AssetDropzone: PropTypes.elementType,
   InsertMediaModal: PropTypes.elementType,
   getItemProps: PropTypes.func,
+  endpoints: PropTypes.object,
```
ChloeSartorelli: Wrong place - this should be nested inside `data`

client/src/containers/AssetAdmin/AssetAdmin.js
ChloeSartorelli: Sorry, that wasn't very well explained of me lol. I meant add validation of the fileIDs to all of the `do*` functions. We should either consistently do that everywhere or consistently not bother.

client/src/containers/Editor/Editor.js
ChloeSartorelli: We are passing the file as a prop to `FormBuilderLoaderComponent` right below this.
The comment specifically is about that:
> Passing `file` to FormBuilderLoader will force a re-render...

client/src/containers/Gallery/Gallery.js
ChloeSartorelli: Right, and that was already happening. I'm asking if that includes the errors that _used_ to be in `graphQLErrors`.
If it doesn't include those, do we have an equivalent set of errors that used to be in `graphQLErrors`?

code/Controller/AssetAdminOpen.php
ChloeSartorelli: Any time we have additional information about a param beyond what the name and strong typehint can provide, the PHP docblock is used for that purpose. That's exactly the sort of thing `@param` is for.

code/Controller/AssetAdminOpen.php
ChloeSartorelli: Ahh fair enough. That's a shame.

code/Controller/AssetAdminOpen.php
ChloeSartorelli: Darn. But fair enough.

code/Controller/AssetAdminOpen.php
ChloeSartorelli: That's better than `getget` at least

code/Controller/AssetAdmin.php
ChloeSartorelli: Still isn't using it? (now called `getQueryStringValue()`)

code/Controller/AssetAdmin.php
ChloeSartorelli: It's not though - not in a way that gives a clear objective correct code to use for these scenarios. That's why we had such long discussions about it :p

code/Controller/AssetAdmin.php
ChloeSartorelli: Good point about the pagination, it's probably okay given that.

code/Controller/AssetAdmin.php
ChloeSartorelli: Love how MDN doesn't really clarify that lol. I'll allow it.

code/Controller/AssetAdmin.php
ChloeSartorelli: Fair

code/Controller/AssetAdminOpen.php
```diff
@@ -0,0 +1,416 @@
...
+            $ancestors = array_reverse($ancestors);
+        }
+        // Children
+        $children = [];
+        $totalCount = 0;
+        if ($includeChildren) {
+            // Scenarios for children/childFiles:
+            // a) No filter - child files
+            // b) Filter - all files
+            // c) Filter with currentFolderOnly - descendant files
```
ChloeSartorelli: Probably each of these scenarios should be its own private method, to keep things tidy.

Right now to parse what this method is doing, I have to parse this whole method and keep all of it in my head.

If these three separate scenarios are their own private methods, I can look at them in isolation, and only keep a summary of them in my head when looking at this method.

ChloeSartorelli (changes requested):

_config/config.yml
emteknetnz: Seems fine to me:

*app/_config.php*
```php
var_dump([
    SilverStripe\AssetAdmin\Controller\AssetAdmin::create()->getThumbnailGenerator()->getGenerates(),
    SilverStripe\AssetAdmin\Controller\AssetAdminOpen::create()->getThumbnailGenerator()->getGenerates()
]);
die;
// array(2) { [0]=> bool(true) [1]=> bool(false) } 
```

client/src/components/UploadField/UploadField.js
emteknetnz: Updated

client/src/containers/AssetAdmin/AssetAdmin.js
emteknetnz: Removed

client/src/containers/Editor/Editor.js
emteknetnz: No longer relevant, at the top of the render method we have

```js
    if (!this.state.file) {
      return null;
    }
```

client/src/containers/Gallery/Gallery.js
emteknetnz: I'm not sure what graphQLErrors actually was, presumably it was an error message of some kind? Maybe it came from Apollo? I don't know.

I guess the equivalent we have now would be actual HTTP status codes whereas before with GraphQL it was always a 200 :-)

code/Controller/AssetAdminOpen.php
emteknetnz: Updated

code/Controller/AssetAdminOpen.php
emteknetnz: I've moved the confusing filtering stuff to a single private method `getFilteredChildFiles()`, I think it makes it a lot clearer having that isolated from everything else

code/Controller/AssetAdmin.php
emteknetnz: Updated

ChloeSartorelli:
Few problems found while testing edge cases:

1. It seems like list view in the asset admin isn't using thumbnails. If I delete the thumbnail variants and then view the list view, I still see the image. If I right click on it and view image in new tab, it's the original full resolution image.
1. If `AssetAdminOpen::apiRead()` throws a json error, the asset admin doesn't load correctly and I see this in the dev console (yes it does just end abruptly - there's no more info other than a js stack trace):
    ```text
    Uncaught (in promise) Error: 
    ```
1. if `AssetAdmin::apiReadDescendantCounts()` throws a json error, the confirm deletion modal is stuck loading with the same error as above in the dev console

I stopped checking at that point - it seems likely that all of the endpoints suffer from similar problems.
I didn't check elemental, but suspect it would probably have the same type of problems.

emteknetnz:
> It seems like list view in the asset admin isn't using thumbnails. If I delete the thumbnail variants and then view the list view, I still see the image. If I right click on it and view image in new tab, it's the original full resolution image.

Replicated this on 5.2, so looks like it's an existing issue

client/src/containers/Gallery/Gallery.js
ChloeSartorelli: Looks like this was probably doing what the jsonerror responses are now doing. That is discussed separately from this discussion thread so I'll mark resolved.

ChloeSartorelli (changes requested):
The new json error handling is great! 

Found one problem when testing this time around.

## Cannot cancel deleting a file

1. Upload a file.
2. Select the file.
3. Click the `...` button
4. Click delete
5. Try to cancel the modal (it won't go away, and there's no errors in the console, no network requests, nothing)

The fact this wasn't caught by behat means we are missing a test scenario for this.

emteknetnz:
Have fixed and added behat tests for cancelling and closing the modal

ChloeSartorelli (approved):
LGTM, works as expected locally.

PR #1483, opened 2024-08-13
ENH Don't emit deprecation warnings for unavoidable API calls

ChloeSartorelli:
`Versioned::canArchive()` is deprecated in https://github.com/silverstripe/silverstripe-versioned/pull/461

We can't stop calling it directly until CMS 6 because people may have implemented extensions that update the result of this permission check.

## Issue
- https://github.com/silverstripe/silverstripe-versioned/issues/447

### DIFF
--- a/code/GraphQL/Resolvers/AssetAdminResolver.php
+++ b/code/GraphQL/Resolvers/AssetAdminResolver.php
@@ -16,6 +16,7 @@ use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\Filterable;
 use SilverStripe\Versioned\Versioned;
 use InvalidArgumentException;
+use SilverStripe\Dev\Deprecation;
 
 class AssetAdminResolver
 {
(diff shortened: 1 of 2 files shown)


ChloeSartorelli:
behat CI failure is pre-existing and not related to this PR

code/GraphQL/Resolvers/AssetAdminResolver.php
```diff
@@ -126,7 +127,7 @@ public static function resolveDeleteFiles($object, array $args, $context, Resolv
         $deletedIDs = [];
         $member = UserContextProvider::get($context);
         foreach ($files as $file) {
-            if ($file->canArchive($member)) {
+            if (Deprecation::withNoReplacement(fn() => $file->canArchive($member))) {
```
emteknetnz:
```suggestion
            $canArchive = Deprecation::withNoReplacement(fn() => $item->canArchive());
            if ($canArchive) {
```
Makes this a lot easier to read

emteknetnz (changes requested):

code/GraphQL/Resolvers/AssetAdminResolver.php
ChloeSartorelli: Done

emteknetnz (approved):

PR #1484, opened 2024-08-15
API Deprecate GraphQL

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799

### DIFF
--- a/code/GraphQL/FileFilter.php
+++ b/code/GraphQL/FileFilter.php
@@ -10,9 +10,20 @@ use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Forms\DateField;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\Filterable;
+use SilverStripe\Dev\Deprecation;
 
+/**
+ * @deprecated 5.3.0 Will be moved to the silverstripe/graphql module
+ */
 class FileFilter
 {
+    public function __construct()
+    {
+        Deprecation::withNoReplacement(function () {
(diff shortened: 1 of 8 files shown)


code/GraphQL/FileFilter.php
```diff
@@ -10,9 +10,20 @@
...
 
+/**
+ * @deprecated 5.3.0 Will be moved to the silverstripe/graphql module in CMS 6
+ */
 class FileFilter
 {
+    public function __construct()
+    {
+        Deprecation::withNoReplacement(function () {
+            Deprecation::notice('5.3.0', 'Will be moved to the silverstripe/graphql module in CMS 6', Deprecation::SCOPE_CLASS);
```
ChloeSartorelli:
```suggestion
 * @deprecated 5.3.0 Will be moved to the silverstripe/graphql module
 */
class FileFilter
{
    public function __construct()
    {
        Deprecation::withNoReplacement(function () {
            Deprecation::notice('5.3.0', 'Will be moved to the silverstripe/graphql module', Deprecation::SCOPE_CLASS);
```

When we did the deprecations for CMS 5, we decided to be fairly standardised in the messages, and part of that was to not include a reference to the specific major release we expect the change to happen in.

Please make this change to other deprecation warnings in this PR and in the related PRs.

ChloeSartorelli (changes requested):

code/GraphQL/FileFilter.php
emteknetnz: Updated

ChloeSartorelli (approved):
LGTM

PR #1485, opened 2024-08-18
API Standardise extension hooks

ChloeSartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11258

### DIFF
--- a/tests/php/Forms/FileFormBuilderTest.php
+++ b/tests/php/Forms/FileFormBuilderTest.php
@@ -120,7 +120,7 @@ class FileFormBuilderTest extends SapphireTest
         // Add extension to simulate different permissions
         File::add_extension(FileExtension::class);
 
-        $this->logInWithPermission('ADMIN');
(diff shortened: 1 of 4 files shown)


tests/php/Forms/FileFormBuilderTest.php
```diff
@@ -120,7 +120,7 @@ public function testEditFileFormWithPermissions()
         // Add extension to simulate different permissions
         File::add_extension(FileExtension::class);
 
-        $this->logInWithPermission('ADMIN');
+        $this->logInWithPermission('CMS_ACCESS_CampaignAdmin');
```
ChloeSartorelli, line 123: Having admin access results in an early `return true` in versioned `canPublish()` so our extension wasn't being called which resulted in incorrectly failing unit tests.

That's only just not become a thing because the path for `public` methods on extensions is different - it was being called "directly" (earlier than the versioned method) instead of being used as an extension hook.

emteknetnz (approved):

PR #1501, opened 2024-10-10
ENH Suppress deprecation notices for API we can't avoid

ChloeSartorelli:
Deprecations added in https://github.com/silverstripe/silverstripe-framework/pull/11420

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322


ChloeSartorelli:
No longer needed

PR #1514, opened 2024-11-21
API Deprecate campaign admin integrations

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/884

### DIFF
--- a/code/Extensions/CampaignAdminExtension.php
+++ b/code/Extensions/CampaignAdminExtension.php
@@ -7,15 +7,23 @@ use SilverStripe\Assets\File;
 use SilverStripe\Core\Extension;
 use SilverStripe\Forms\FormAction;
 use SilverStripe\Security\Permission;
+use SilverStripe\Dev\Deprecation;
 
 /**
  * Extension that updates the Popover menu of `FileFormFactory`.
  * This extension will only be applied if the `campaign-admin` module is installed.
  *
  * @extends Extension<FileFormFactory>
+ * @deprecated 2.4.0 Will be removed without equivalent functionality to replace it
  */
 class CampaignAdminExtension extends Extension
 {
+    public function __construct()
+    {
+        Deprecation::noticeWithNoReplacment('2.4.0', scope: Deprecation::SCOPE_CLASS);
+        parent::__construct();
+    }
+
     /**
      * Update the Popover menu of `FileFormFactory` with the "Add to campaign" button.
      *


code/Extensions/CampaignAdminExtension.php
```diff
@@ -7,15 +7,23 @@
...
  * This extension will only be applied if the `campaign-admin` module is installed.
  *
  * @extends Extension<FileFormFactory>
+ * @deprecated 5.4.0 Will be removed without equivalent functionality to replace it
  */
 class CampaignAdminExtension extends Extension
 {
+    public function __construct()
+    {
+        Deprecation::notice('5.4.0', 'Will be removed without equivalent functionality to replace it', Deprecation::SCOPE_CLASS);
```
GuySartorelli, line 23: Use the `noticeWithoutReplacement` so it gets wrapped. Anything without a replacement should be wrapped.

ChloeSartorelli: Use the `noticeWithNoReplacment` so it gets wrapped. Anything without a replacement should be wrapped.
```suggestion
       Deprecation::noticeWithNoReplacment('5.4.0', scope: Deprecation::SCOPE_CLASS);
```

ChloeSartorelli (changes requested):

code/Extensions/CampaignAdminExtension.php
ChloeSartorelli:
```suggestion
 * @deprecated 2.4.0 Will be removed without equivalent functionality to replace it
 */
class CampaignAdminExtension extends Extension
{
    public function __construct()
    {
        Deprecation::notice('2.4.0', 'Will be removed without equivalent functionality to replace it', Deprecation::SCOPE_CLASS);
```

ChloeSartorelli (changes requested):

code/Extensions/CampaignAdminExtension.php
ChloeSartorelli: `2.4.0` not `5.4.0`

emteknetnz: Updated

emteknetnz: Updated

ChloeSartorelli: You missed the PHPDoc

emteknetnz: Opps, done

ChloeSartorelli (approved):
LGTM

PR #1524, opened 2024-12-16
MNT Fix behat test

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/349

Fixes https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/12342589564/job/34442717488#step:12:777

    And I should see the file named "folder1" in the gallery     # SilverStripe\AssetAdmin\Tests\Behat\Context\FixtureContext::iShouldSeeTheGalleryItem()
      File named folder1 could not be found

### DIFF
--- a/tests/behat/features/manage-files.feature
+++ b/tests/behat/features/manage-files.feature
@@ -281,6 +281,7 @@ Feature: Manage files
       Then I should see "No items found"
       And I should see "File type: Archive"
     Then I press the "Close" button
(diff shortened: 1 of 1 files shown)


ChloeSartorelli (approved):
Not a fan of adding seconds to the build without checking why it's needed but for now this is fine.
I might recommend we look at adding "and I wait until I see" for specific elements to replace fixed-time waiting in the new year.

Issue #1016, opened 2019-10-01
Dragged file is the wrong size


newleeland:
When dragging a file. The file that is "picked up" by the mouse is much bigger than originally designed. This makes it harder to see the highlighted folders that one can drop the file into. Additionally, when dragging more than one item, the drag items look broken

![Screen Shot 2019-10-01 at 4.44.02 PM.png](https://images.zenhubusercontent.com/596d365561b4ca3b0a74bd86/66140320-ac4e-4d11-8988-49aff9b89105) 

Its should be about this size:
![Screen Shot 2019-10-01 at 4.45.23 PM.png](https://images.zenhubusercontent.com/596d365561b4ca3b0a74bd86/94e7c26f-be73-4588-98d6-c8a6167a3b72)

ChloeSartorelli:
Fixed in CMS 6.

PR #1527, opened 2024-12-17
DEP Replace react-dnd with dnd-kit

ChloeSartorelli:
This PR replaces the `react-dnd` with `dnd-kit` with like-for-like functionality, except:

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1016 has been fixed - I had to alter the styling anyway because of the way dragging the `GalleryItemDragLayer` works now. The changes aren't fully compatible with CMS 5 so I won't be backporting the changes there, so `#1016` needs to remain open.
- When dragging an item on top of a droppable zone, the cursor doesn't update. The one used by `react-dnd` is actually coming from the browser itself with default HTML5 drag-and-drop functionality (see https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect). That cursor _is not selectable_ with CSS as far as I can tell.
    - We _could_ use the `no-drop` cursor when dragging anywhere else, and only use `grabbing` for when we're over a valid drop target - but that wouldn't be consistent with cursors used in other drag-and-drop scenarios in the CMS. I'd recommend if we want to do that, it's handled in a wider-reaching card that standardises this approach.

Ran `yarn upgrade` to get https://github.com/silverstripe/eslint-config/pull/29

## Issue
- https://github.com/silverstripe/.github/issues/343


client/src/containers/Gallery/GalleryDND.js
ChloeSartorelli, line 1: Note that I didn't split converting this or other components to functional as its own PR or commit because the lifycycle functions needed to be removed anyway. Converting it with the lifecycle events in-tact and then removing them separately would have been more work.

client/src/containers/Gallery/GalleryDND.js
```diff
@@ -1,80 +1,77 @@
...
+    setDragging(false);
+    setDraggingItemID(null);
+    setDraggingItemProps(null);
+  });
 
-    return (
-      <div className={classnames(className, { 'gallery__main--dragging': this.state.dragging })}>
+  return (
+    <div className={classnames(className, { 'gallery__main--dragging': dragging })}>
+      <DndContext sensors={sensors} onDragStart={handleDragStart} onDragEnd={handleDragEnd} modifiers={[snapCenterToCursor]}>
```
ChloeSartorelli, line 61: I tried using built-in modifiers (including `restrictToWindowEdges`) to restrict where we can drag the files, but none of them worked how I'd expect them to, so I opted to go without. The end result is the same unbound dragging as the old approach so it's not worse than it was before, at least.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -135,6 +135,10 @@ class GalleryItem extends Component {
       thumbnailClassNames.push('gallery-item__thumbnail--no-preview');
     }
 
+    if (this.props.item.type === 'folder') {
+      thumbnailClassNames.push('gallery-item__thumbnail--folder');
+    }
```
ChloeSartorelli, line 140: Needed for styling the draglayer for folders

client/src/components/GalleryItem/GalleryItem.scss
```diff
@@ -298,23 +298,23 @@ $gallery-item-label-height: 40px;
 }
 
 .gallery-item--archive .gallery-item__thumbnail {
-  background: $white url("../../images/icon_archive.png") center center no-repeat;
+  background: transparent url("../../images/icon_archive.png") center center no-repeat;
```
ChloeSartorelli, line 301: These changes were required to fix a problem where the background was being drawn over top of the border of the drag layer.
Note that I can't just use `background-image` because that seems to be overridden by something elseand results in no thumbnail being displayed

client/src/components/GalleryItem/GalleryItemDragLayer.js
```diff
@@ -1,76 +1,38 @@
...
-class GalleryItemDragLayer extends Component {
-  getOffset() {
-    const {
-      offset,
-      dragged,
-    } = this.props;
-    return {
-      transform: offset && `translate(${offset.x + dragged.x}px, ${offset.y + dragged.y}px)`,
-    };
-  }
```
ChloeSartorelli, line 16: Offset stuff is handled by the `snapCenterToCursor` modifier now

client/src/components/GalleryItem/README.md
```diff
@@ -4,7 +4,8 @@ Displays a file/folder as a thumbnail with relevant actions.
 
 ## Example
 ```js
-<GalleryItem item={{id: 2}} selectable={true} />
+<GalleryItem item={{id: 2, exists: true, type: 'file', category: 'document'}} selectable={true} />
+<GalleryItem item={{id: 2, exists: true, type: 'folder', category: 'folder'}} selectable={true} />
```
ChloeSartorelli, line 8: While updating styling, I had to render out these directly so I could mess with css in the browser.
I found that just an ID wasn't enough, so I've updated for an example for a file and for a folder, for future reference.

client/src/components/GalleryItem/draggable.js
```diff
@@ -1,75 +1,29 @@
...
-      }
-
-      return { selected, props };
-    },
-    endDrag(props) {
-      const { id } = props.item;
-      if (typeof props.onDrag === 'function') {
-        props.onDrag(false, id);
-      }
-    },
```
ChloeSartorelli, line 28: Handling what happens on drag/drop is done inside `GalleryDND` instead.

client/src/containers/Gallery/Gallery.js
```diff
@@ -952,7 +950,12 @@ class Gallery extends Component {
         ref={gallery => { this.gallery = gallery; }}
       >
         {this.renderTransitionBulkActions()}
-        <GalleryDND className={galleryClasses.join(' ')}>
+        <GalleryDND
+          onDragStartEnd={(dragging) => this.handleEnableDropzone(!dragging)}
```
ChloeSartorelli, line 954: Replaces `handleDrag` in `ThumbnailView`

tests/behat/src/FixtureContext.php
```diff
@@ -499,4 +499,29 @@ public function stepCreateMaximumFileSizeStep($size): void
...
+     * Example: When I drag the file "file1" to the folder "my folder"
+     *
+     * @When /^I drag the file "([^"]*)" to the folder "([^"]*)"$/
+     */
+    public function stepDragTheFileToTheFolder(string $fileName, string $folderName): void
+    {
+        $file = $this->getGalleryItem($fileName)->find('css', '.gallery-item__draggable');
+        $folder = $this->getGalleryItem($folderName)->find('css', '.gallery-item__droppable');
+        $file->dragTo($folder);
+    }
```
GuySartorelli, line 513: I couldn't use the existing `I drag the "some element" element to the "some other element" element` method, because the only selectors available relied on fixed IDs which works locally but fails in CI.

tests/behat/src/FixtureContext.php
```diff
@@ -499,4 +499,33 @@ public function stepCreateMaximumFileSizeStep($size): void
...
+     * @When /^I drag the (?:file|folder) named "([^"]+)" to the folder "([^"]+)"$/
+     */
+    public function stepDragTheFileToTheFolder(string $fileName, string $folderName): void
+    {
+        $file = $this->getGalleryItem($fileName)->find('xpath', "/ancestor-or-self::div[contains(@class, 'gallery-item__draggable')]");
+        Assert::assertNotNull($file, "File named {$fileName} could not be found or isn't draggable");
+        $folder = $this->getGalleryItem($folderName)->find('xpath', "/ancestor-or-self::div[contains(@class, 'gallery-item__droppable')]");
+        Assert::assertNotNull($folder, "Folder named {$folderName} could not be found or isn't droppable");
+        $file->dragTo($folder);
+    }
```
ChloeSartorelli, line 515: I couldn't use the existing `I drag the "some element" element to the "some other element" element` method, because the only selectors available rely on fixed IDs which works locally but fails in CI.

tests/behat/features/drag-and-drop.feature
```diff
@@ -0,0 +1,34 @@
+@javascript @assets @retry @job3 @gsat
```
emteknetnz:
```suggestion
@javascript @assets @retry @job3
```

emteknetnz (changes requested):
Couple of weird things happening when I tested locally in firefox, neither of which happen in CMS 5
- After moving, the thumbnail animates back to its original location
- There's a red toast error when attempting to move to the parent folder. Only happens with moving files, not moving folders.

https://www.youtube.com/watch?v=vnHESa15Q8g

ChloeSartorelli:
> There's a server error when attempting to move to the parent folder. Only happens with moving files, not moving folders.

If the parent folder is the root, [this is a known regression](https://github.com/silverstripe/silverstripe-asset-admin/issues/1529) that was already linked in the issue description.

emteknetnz:
Server error still happens when trying to move files to the parent folder in deeply nested folders

ChloeSartorelli:
Can't reproduce. Please:
1. Compare against the `3.0` branch without this PR to rule out an existing regression
2. If it _is_ an existing regression, open a new card about it
3. If it _is not_ an existing regression, please provide server error message and stack trace - I may even ask for a copy of your DB and assets since as I said I can't reproduce.

tests/behat/features/drag-and-drop.feature
ChloeSartorelli: Done

ChloeSartorelli:
> After moving, the thumbnail animates back to its original location

Good spot. Fixed.

emteknetnz:
Server error is [this issue](https://github.com/silverstripe/silverstripe-asset-admin/issues/1529)

emteknetnz (approved):

PR #1533, opened 2025-01-06
adding keywords to the composer.json file

sunnysideup:


michalkleiner:
@sunnysideup we probably don't need to reformat the whole file just to add the keywords.

Issue #1529, opened 2024-12-19
CMS 6 REGRESSION - Can't move files/folders up to the root level


ChloeSartorelli:
Dragging files and folders over the back button in the asset admin (![back button](https://github.com/user-attachments/assets/cf815a72-d148-4f45-ae94-cf49ecb60984)) will move those files up to the parent folder.

In CMS 5 this also works with moving them to the root folder (id `0`), but in CMS 6 trying to move them to the root folder results in an error.
This is likely a result of the refactor that replaced the GraphQL queries with regular POSTs to controller routes.

Make sure to add a behat test for this when fixing it.

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1534
- https://github.com/silverstripe/silverstripe-elemental/pull/1301

ChloeSartorelli:
PRs merged

PR #1534, opened 2025-01-06
FIX Handle moving files to root folder

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1529

Note that while the parent issue says "Make sure to add a behat test for this when fixing it." since this is a PHP issue rather than a JS issue, a regular unit test is more appropriate


ChloeSartorelli (approved):
It would have been good to also add a behat test as there are other things that could go wrong that would only be caught in an end-to-end test, but I suppose it's not strictly necessary.

Works well locally.

PR #1537, opened 2025-01-15
DEP Upgrade bootstrap and reactstrap

ChloeSartorelli:
Had to remove `react` as a direct dependency because it got confused between the version being installed locally and the one included from admin, which resulted in false-negatives with errors saying "you can't use hooks" when they were in fact being used correctly.

Update to eslint is because it defaulted to assuming the _latest_ version of react which resulted in a lint error, because we're not using the latest version of react.

See https://github.com/silverstripe/silverstripe-admin/pull/1889 which links to the specific upgrade guide entries for most changes made.

## Issue
- https://github.com/silverstripe/.github/issues/332


.eslintrc.js
ChloeSartorelli, line 1: Removing `react` from the dependencies resulted in jest defaulting to "react latest" which caused some tests to fail.

client/src/containers/Editor/LegacyPopoverField.scss
```diff
@@ -13,7 +13,7 @@
 
     &.top,
     &.bottom {
-      .arrow {
+      .popover-arrow {
```
ChloeSartorelli, line 16: https://getbootstrap.com/docs/5.0/migration/#popovers

> Renamed `.arrow` to `.popover-arrow` in our default popover template.

client/src/containers/InsertMediaModal/InsertMediaModal.js
```diff
@@ -102,12 +102,10 @@ class InsertMediaModal extends Component {
     return (
       <button
         type="button"
-        className="close modal__close-button insert-media-modal__close-button"
+        className="close btn-close modal__close-button insert-media-modal__close-button"
         onClick={this.props.onClosed}
         aria-label={i18n._t('FormBuilderModal.CLOSE', 'Close')}
-      >
-        <span aria-hidden="true">×</span>
```
ChloeSartorelli, line 109: This `x` was being rendered on top of the existing styling for the modal close button, so there were two `x` being rendered

emteknetnz (approved):

PR #1546, opened 2025-02-26
ENH Replace react-griddle with react-table

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1892


ChloeSartorelli:
Seems to work as expected but will hold off merging until this isn't in draft

tests/behat/features/table-view.feature
```diff
@@ -10,13 +10,13 @@ Feature: Change view for asset admin
       And I am logged in as a member of "EDITOR" group
       And I go to "/admin/assets"
 
+@sboyd
   @assets
   Scenario: I can switch the gallery between table and thumbnail mode
     When I click on the file named "folder1" in the gallery
       And I press the "table" button
       And I wait until I see the ".gallery__table-row" element
-    Then I should see a ".gallery__main-view--table" element
```
emteknetnz, line 18: This was a griddle class that wrapped the table. We're already asserting that we see a `.gallery__table-row` and a `.gallery__table-column--image` so we don't need to assert anything else.

ChloeSartorelli (approved):
LGTM

PR #1548, opened 2025-03-04
ENH Add campaign-admin support back in

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/895


client/src/containers/Editor/Editor.js
ChloeSartorelli, line 1: Just as a sanity check since you haven't provided any information - did you check if this can be done by adding code from the campaign admin module itself?

code/Controller/AssetAdmin.php
```diff
@@ -24,6 +25,7 @@
 use SilverStripe\Assets\Image;
 use SilverStripe\Assets\Storage\AssetNameGenerator;
 use SilverStripe\Assets\Upload;
+use SilverStripe\CampaignAdmin\AddToCampaignHandler;
```
ChloeSartorelli: Why are we adding this `use` statement, or the `RuntimeException` one?

code/Controller/AssetAdmin.php
```diff
@@ -212,7 +215,7 @@ public function getClientConfig(): array
 
         $parentClientConfig = parent::getClientConfig();
         $modalController = ModalController::singleton();
-        return array_merge($parentClientConfig, [
+        return ArrayLib::array_merge_recursive($parentClientConfig, [
```
ChloeSartorelli, line 216: What's this change for?

ChloeSartorelli (changes requested):

code/Controller/AssetAdmin.php
emteknetnz: That 'form.addToCampaignUrl' added in [AddToCampaignExtension::updateClientConfig()](https://github.com/silverstripe/silverstripe-campaign-admin/pull/346/files#diff-ba66fc15f570c5ac21be4327395879675bb715fa1285a66d7fa2aeb0a1c23042R35) isn't overridden - this method is calling `parent::getClientConfig();` before updating `$parentClientConfig()`

code/Controller/AssetAdmin.php
emteknetnz: Removed

client/src/containers/Editor/Editor.js
emteknetnz: I don't understand what you mean

Code added is what was originally removed - https://github.com/silverstripe/silverstripe-asset-admin/pull/1512/files#diff-135b0ccbc4091634cd30eb89d6876ea202900f53c84dd12667c7fb07a6b72745L8

ChloeSartorelli: I'm asking if there is a way to add this into the campaign admin module instead of here, since it's code that is specific to campaigns.
If you have looked into it and it's not feasible, please just explain why and that'll be that.

emteknetnz: There didn't seem like to get the javascript to work for like for like if it was moved to campaign admin. While it's usually easy to do extensions for PHP, it's not the case for our javascript which has UX.

ChloeSartorelli (approved):
Probably could use the JS injector here to move the code into campaign admin, but this will do for now give time constraints.

Issue #1528, opened 2024-12-18
Replace `react-selectable` dependency with something maintained


ChloeSartorelli:
[react-selectable](https://www.npmjs.com/package/react-selectable) looks like it may have been made specifically for asset-admin but is no longer being maintained.
It also has some strange coupling - e.g. see https://github.com/unclecheese/react-selectable/blob/802b1ec9ad1ef97fd241f48e612dd1b334b447b8/src/selectable-group.js#L126-L127

I recommend we replace it with https://www.npmjs.com/package/@air/react-drag-to-select

## Acceptance Criteria
- [x] `react-selectable` is replaced with a maintained alternative

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1550
- https://github.com/silverstripe/silverstripe-admin/pull/1913

emteknetnz:
No changelog PR for this one - we have a broad "Several JavaScript dependencies have been updated, and a few have been replaced." in the 6.0.0 changelog already.  I don't think this dep warrants its own mention because it's so specialized.

ChloeSartorelli:
PRs merged

PR #1550, opened 2025-03-12
ENH Replace react-selectable with react-drag-to-select

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1528


client/src/containers/Gallery/Gallery.js
```diff
@@ -548,28 +547,6 @@ class Gallery extends Component {
     this.handleGroupSelect(ids, new Event('na'));
   }
 
-  /**
-   * Pick if the selection started from inside the pagination. If it started from inside the
-   * pagination, cancel it to prevent inteference with the normal pagination.
-   * @param Event e
-   * @returns {boolean}
-   */
-  handleBeginSelection(e) {
```
emteknetnz, line 557: I'm not sure what these actually did, from what i can tell it didn't actually do anything. It doesn't appear to be necessary going forward.

client/src/containers/Gallery/Gallery.js
```diff
@@ -957,14 +934,9 @@ class Gallery extends Component {
...
-            enabled={this.props.view === 'tile' && this.props.type === ACTION_TYPES.ADMIN}
-            className="flexbox-area-grow fill-height gallery__main--selectable"
-            onSelection={this.handleGroupSelect}
-            onNonItemClick={this.handleClearSelection}
-            onBeginSelection={this.handleBeginSelection}
-            preventDefault={false}
-            fixedPosition
+          <Selectable
+            isEnabled={this.props.view === 'tile'}
+            onSelectionChange={this.handleGroupSelect}
```
ChloeSartorelli: Why was `&& this.props.type === ACTION_TYPES.ADMIN` dropped from the enabled condition?

client/src/containers/Gallery/Gallery.js
```diff
@@ -65,7 +65,6 @@ class Gallery extends Component {
     this.handleBulkUnpublish = this.handleBulkUnpublish.bind(this);
     this.handleBulkMove = this.handleBulkMove.bind(this);
     this.handleBulkInsert = this.handleBulkInsert.bind(this);
-    this.handleBeginSelection = this.handleBeginSelection.bind(this);
     this.handleGroupSelect = this.handleGroupSelect.bind(this);
     this.handleClearSelection = this.handleClearSelection.bind(this);
```
ChloeSartorelli, line 69: `handleClearSelection` still exists, but it isn't passed into the new `<Selectable>` component. Is it still used?

ChloeSartorelli (changes requested):
This has broken drag-and-drop to upload. This video shows me dragging a folder over the asset admin, but the display is completely wrong.

[Screencast from 18-03-25 16:59:34.webm](https://github.com/user-attachments/assets/c5f24222-1ef5-4a2b-9c13-1e13d0c1ebcc)

Compare with CMS 6 which allows me to drop the file(s) anywhere.

https://github.com/user-attachments/assets/04917149-4019-44f1-bba2-76dadd8d712c

client/src/containers/Gallery/Gallery.js
ChloeSartorelli: This did exactly what it said it did in the comment - if you try to click-and-drag from the "next" or "previous" page button at the bottom of the gallery, without this PR it doesn't start a selection, but with this PR it does.

_Very_ minor difference, but it is a little cleaner if it doesn't try selecting when your click starts on another interactive element.

ChloeSartorelli (changes requested):
Just noticed as well that if I have only a few files, and I start dragging from _below_ them, the dotted box isn't actually displayed until I drag up and my cursor reaches where the files are.

client/src/containers/Selectable/Selectable.js
```diff
@@ -0,0 +1,127 @@
...
+ * Component that wraps around a list of items to make them selectable by dragging a box around them
+ * and calls a callback when the selection changes
+ * Uses the react-drag-to-select library
+ */
+function Selectable(props) {
+  // State hooks
+  const [mouseEvent, setMouseEvent] = useState(null);
+  const [lastFileIDs, setLastFileIDs] = useState([]);
+
+  // Ref hooks
```
ChloeSartorelli:
```suggestion
  const [mouseEvent, setMouseEvent] = useState(null);
  const [lastFileIDs, setLastFileIDs] = useState([]);

```
Those are self-explanatory

client/src/containers/Selectable/Selectable.js
```diff
@@ -0,0 +1,127 @@
+import { boxesIntersect, useSelectionContainer } from '@air/react-drag-to-select';
+import React, { useState, useRef } from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * Component that wraps around a list of items to make them selectable by dragging a box around them
+ * and calls a callback when the selection changes
+ * Uses the react-drag-to-select library
```
ChloeSartorelli:
```suggestion
 * Uses the @air/react-drag-to-select library
```
When I first looked that up I found https://www.npmjs.com/package/react-drag-to-select and was worried because that's abandoned.

With any comments referencing this, can you please add the full reference i.e. `@air/react-drag-to-select` to avoid that confusion

client/src/components/GalleryItem/draggable.js
```diff
@@ -10,8 +10,8 @@ export default function draggable(Item) {
 
     return <div
       className="gallery-item__draggable"
-      // The `draggable` attribute is needed to tell react-selectable to not start selecting on mousedown
-      draggable
+      // The `data-draggable` attribute is needed to tell react-drag-to-select to not start selecting on mousedown
+      data-draggable="true"
```
ChloeSartorelli: Does the library actually handle that at all or is it just our custom `Selectable` component?

ChloeSartorelli (changes requested):

client/src/components/GalleryItem/draggable.js
emteknetnz: In theory it's the library https://github.com/AirLabsTeam/react-drag-to-select/blob/main/src/hooks/useSelectionContainer.tsx#L34

But in practice it didn't seem to work, so I manually implemented this in Selectable.shouldStartSelecting()

I'll update the comment

client/src/containers/Gallery/Gallery.js
emteknetnz: Yeah, it's passed into `<BulkActionsComponent>` in `Gallery.renderBulkActions()`

client/src/containers/Gallery/Gallery.js
emteknetnz: Have added back in though changed name to `handleShouldStartSelecting` to reflect what it's used on in `<Selectable>`

client/src/containers/Gallery/Gallery.js
emteknetnz: Didn't realise I had, added in

client/src/containers/Selectable/Selectable.js
emteknetnz: Updated

client/src/containers/Selectable/Selectable.js
emteknetnz: Updated

emteknetnz:
@GuySartorelli Have added a height: '100%' style to some of the div's, and adding 'display: flex' via a scss on another.  that resolved both the 'drop in file' and 'drag from below' issues

There is one issue, though probably very minor in practice, where the dashed box will get cut off if you drag to panel to the left where the links to model admins are (or above the gallery).  This has no functional impact though and most people probably wouldn't even drag their mouse there, or if they did notice the issue.  I don't see any easy fix to this so I think we should just live with this minor limitation

client/src/containers/Gallery/Gallery.js
```diff
@@ -944,6 +943,8 @@ class Gallery extends Component {
       cssClasses.push('gallery__main--has-opened-item');
     }
 
+    const isEnabled = this.props.view === 'tile' && this.props.type === ACTION_TYPES.ADMIN;
```
ChloeSartorelli:
```suggestion
    const canSelect = this.props.view === 'tile' && this.props.type === ACTION_TYPES.ADMIN;
```
`isEnabled` out here looks like it's talking about the enabled state of the gallery

ChloeSartorelli (changes requested):
## Actual regressions

These are bugs that this PR introduces that definitely need fixing.

1. After selecting some items, clicking in an empty space doesn't clear the selection.

## Minor bugs

These are all things that are clearly bugs, but I'm not sure if it's buggy enough to worry about.

1. This PR creates a horizontal scrolling area if you scroll off to the side - BUT that disappears as soon as you let go of the mouse.
image
1. This PR also makes it so that if you click and drag from the pagination number it starts selecting text which it didn't do before. Clicking in a blank area _does not clear the text selection_. Clicking and dragging to select files after that _also selects their text_.
1. It is a _little_ weird that the selection display gets cut off on the left earlier than the left menu, and earlier going upwards than the breadcrumb section.

client/src/containers/Gallery/Gallery.js
emteknetnz: Updated

emteknetnz:
> After selecting some items, clicking in an empty space doesn't clear the selection.

Fixed

> This PR creates a horizontal scrolling area if you scroll off to the side - BUT that disappears as soon as you let go of the mouse.

Fixed

> This PR also makes it so that if you click and drag from the pagination number it starts selecting text which it didn't do before.

Kind of weird how this started. I've created a [new PR](https://github.com/silverstripe/silverstripe-admin/pull/1913) to fix this.

> Clicking in a blank area does not clear the text selection. Clicking and dragging to select files after that also selects their text.

Fixed

> It is a little weird that the selection display gets cut off on the left earlier than the left menu, and earlier going upwards than the breadcrumb section.

I think we'll need to leave this as is as I don't know how to resolve this. IMO it's a pretty minor aesthetic issue that doesn't affect actual functionality at all.

ChloeSartorelli:
>> It is a little weird that the selection display gets cut off on the left earlier than the left menu, and earlier going upwards than the breadcrumb section.
>
> I think we'll need to leave this as is as I don't know how to resolve this. IMO it's a pretty minor aesthetic issue that doesn't affect actual functionality at all.

It is quite disorienting for me - but if there's not much we can do then there's not much we can do.

My one thought is - what if we made the `.gallery__outer` element the parent of the scrollable? 
image
If that's doable without much effort, I think it would resolve it, because then the boundaries for where it stops rendering are marked with clear lines rather than invisible ones.

If that's not doable or is considerable effort, I'm okay with not doing it.

emteknetnz:
Yeah that seemed to work by simply moving where the `<Selectable>` component in in the hieracrchy, the top and left boundaries aren't dotted lines unfortunately, though there's less "cutting off" when you drag above the area

I also removed the item list classname prop in the component as its direct children are no longer items

ChloeSartorelli:
That's way better🎉

ChloeSartorelli (approved):
LGTM, works as expected

Issue #1553, opened 2025-03-25
CMS 6 REGRESSION: File status displays incorrectly in table view


ChloeSartorelli:
In the table view of `/admin/assets` the "STATUS" column styling should be displayed the same as it is in other sections, but instead it has the blue background usually reserved for custom status flags. See https://docs.silverstripe.org/en/6/developer_guides/customising_the_admin_interface/status_flags/#status-flags

Note this column should also be sortable, but in CMS 6 it does not have the sortable indicator.

CMS 5:

![CMS 5 shows the file size](https://github.com/user-attachments/assets/e28fbb4f-cb46-4f80-bcbf-0dc446457a2d)

CMS 6:

![CMS 6 doesn't show the file size](https://github.com/user-attachments/assets/045cd30c-e8a2-441e-8de0-cc97550d99ad)

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1554

ChloeSartorelli:
@emteknetnz Please link to any PRs

ChloeSartorelli:
PR merged

Issue #1552, opened 2025-03-25
CMS 6 REGRESSION: File size not displayed in table view


ChloeSartorelli:
In the table view, the file size of all files should be displayed in the "SIZE" column, but "N/A" is displayed instead. Note it should also be a sortable column, though in CMS 6 the column does not have the sortable indicator.

CMS 5:

![CMS 5 shows the file size](https://github.com/user-attachments/assets/e28fbb4f-cb46-4f80-bcbf-0dc446457a2d)

CMS 6:

![CMS 6 doesn't show the file size](https://github.com/user-attachments/assets/045cd30c-e8a2-441e-8de0-cc97550d99ad)

## PR
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1554

ChloeSartorelli:
PR merged

PR #1554, opened 2025-03-25
FIX Table view status badge and file size

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1553

Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1552

Fixes:
- File status badge styling
- File size not rendering

Note that reported missing sortable headers for 'status' and 'size' are not regressions - those headers don't sort anything in CMS 5.4 so CMS 6 removing them is actually an improvement


client/src/containers/TableView/TableView.js
```diff
@@ -7,7 +7,6 @@ import Paginator from 'components/Paginator/Paginator';
 import FileStatusIcon from 'components/FileStatusIcon/FileStatusIcon';
 import VersionedBadge from 'components/VersionedBadge/VersionedBadge';
 import { galleryViewPropTypes, galleryViewDefaultProps } from 'containers/Gallery/Gallery';
-import { fileSize } from 'lib/DataFormat';
```
emteknetnz, line 10: This isn't needed - it converts a int in to something like "234 KB" - however the data some formatted from the server in the XHR request

I would update the server to send back an int instead, however I don't know what effect that would have on pagination, and sorting by file size in CMS 5 is already non-functional.  I don't want this bugfix PR to turn into an enhancement PR

client/src/containers/TableView/TableView.js
```diff
@@ -92,12 +90,14 @@ function TableView(_props) {
...
-          status: 'draft'
+          status: 'draft',
+          className: 'status-addedtodraft',
         });
       } else if (rowData.modified) {
         flags.push({
           key: 'status-modified',
-          status: 'modified'
+          status: 'modified',
+          className: 'status-modified',
```
ChloeSartorelli, line 100: We should probably be using the generic status flag functionality instead - which would let people add their own statuses to files. I would have done it when I introduced that functionality except I didn't realise this was here.

However, we shouldn't do that _now_ because that sort of change isn't after-beta material. I'll open a separate card to do that in a future minor release.

client/src/containers/TableView/TableView.js
ChloeSartorelli: Note: This comment has to do with sorting. The sorting capability doesn't work with the string. Changing it to an int might make the sort work, but could cause additional problems.

ChloeSartorelli (approved):
LGTM, works as expected. Confirmed that sorting by filesize and status didn't work in CMS 5 despite the visual icon on the column headers.

PR #1557, opened 2025-03-27
ENH Remove unnecessary calls to setUseSearchContext()

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11647

I'm removing these calls in CMS 6 rather than CMS 5 because the original PR targets CMS 5.3, and we'd need to add in a composer conflict to ensure the version of framework we're using is the latest patch, and the remove those conflicts in CMS 6.

I think it's cleaner to just target CMS 6 and leave the redundant though harmless calls in CMS 5

### DIFF
--- a/code/Forms/AssetFormFactory.php
+++ b/code/Forms/AssetFormFactory.php
@@ -395,8 +395,7 @@ abstract class AssetFormFactory implements FormFactory
                 _t(__CLASS__ . '.VIEWERMEMBERS', 'Viewer Users'),
                 Member::get()
             )
-                ->setIsLazyLoaded(true)
-                ->setUseSearchContext(true),
+                ->setIsLazyLoaded(true),
             OptionsetField::create(
                 "CanEditType",
                 _t(__CLASS__ . '.EDITHEADER', 'Who can edit this file?')
@@ -412,7 +411,6 @@ abstract class AssetFormFactory implements FormFactory
                 Member::get()
             )
                 ->setIsLazyLoaded(true)
-                ->setUseSearchContext(true)
         );
 
         return $tab;


ChloeSartorelli:
I understand why you don't want to do this in 5.3, but why not do it for 5.4?

emteknetnz:
I guess I could just target 5.4 and conflict <5.3.18 (assuming that's the correct framework patch), however as mentioned we'd need to the merge-up and remove conflict from CMS 6, it's just extra PRs

You want me to do that?

ChloeSartorelli:
I don't understand the need to add a "conflict" if we're doing this for CMS 5.4. But no, if that somehow is what would be needed, I'd prefer we don't do that.

If this is a breaking change, we need Jenn's approval to do it earlier than 6.1

emteknetnz:
Conflict is because you can use silverstripe without recipe-cms, therefore there's no lock-step between modules like asset-admin and framework.

This means you could install say asset-admin 5.4.0 with framework 5.3.0 which does not have the framework 5.3 patch that was just merged

Therefore you need to add a composer constraint in asset-admin 5.4 `"silverstripe/framework": "<5.3.18"` to ensure that we don't cause a regression

If it sound complicated it probably is, hence why I just said target CMS 6 because it's easier

Let me know what you'd like done

ChloeSartorelli:
We don't typically add conflicts in a minor release when we know there might be some mismatch like this. We rely on semantic versioning to tell people that "hey this is a minor release and may result in changes beyond this module that you will have to resolve in your project or by updating additional dependencies".

IMO we should do this in 5.4 (with approval from Jenn) and call it out in the changelog.

emteknetnz:
Talked offline, will keep targetting 6

ChloeSartorelli (approved):
LGTM

PR #1560, opened 2025-04-24
ENH Prepopulate version number cache

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11703

Requires https://github.com/silverstripe/silverstripe-versioned/pull/500

When viewing files in asset admin there are a large number of DB queries triggered by `$file->canView()` which called `Versioned->canViewVersioned()`. For every file in the paginated view, you will get 2 queries:

`SELECT "Version" FROM "File" WHERE "ID" = ?`
`SELECT "Version" FROM "File_Live" WHERE "ID" = ?`

This PR optimises the down to a pair of large queries whose results are cached in memory:

`SELECT "ID", "Version" FROM "File" WHERE "ID" IN (?, ?, ?, ?, ?, ?, ?, ?)`
`SELECT "ID", "Version" FROM "File_Live" WHERE "ID" IN (?, ?, ?, ?, ?, ?, ?, ?)`

### DIFF
--- a/code/Controller/AssetAdminOpen.php
+++ b/code/Controller/AssetAdminOpen.php
@@ -165,6 +165,10 @@ class AssetAdminOpen extends LeftAndMain
                 $query->sort('"IsFolderTmp" DESC', null, true);
             });
             $childFiles = $childFiles->limit($limit, $offset);
+            // Prepopulate the Versioned cache used by Versioned::canViewVersioned() which is called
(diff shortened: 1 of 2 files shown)


ChloeSartorelli (approved):
Local performance is about the same. As advertised, number of queries is dramatically reduced, which will improve performance in situations where there is latency communicating with the database.

Issue #1562, opened 2025-04-30
CMS 6 regression: Overlapping requests with non-blocking sessions gives outdated view of assets


ChloeSartorelli:
When non-blocking sessions are used, in CMS 6 specifically, saving a file will result in `refetchFolder()` being called in `AssetAdmin.js` while the save request is still in flight. That function fetches information about the assets in the current folder to display in the gallery view.

The result is that the outdated asset information is displayed in the gallery view after saving. For example changing the permissions for a file doesn't update the ![Image](https://github.com/user-attachments/assets/be491864-80eb-4d1e-9445-8d3204e7d83e) icon until you refresh.

This network waterfall graph shows the two requests overlapping when non-blocking sessions are used:

![image](https://github.com/user-attachments/assets/5cd1314f-e44b-45f9-912a-5b3a8c5629aa)

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1563

PR #1563, opened 2025-04-30
FIX Ensure refetching asset data happens after save is complete

ChloeSartorelli:
With non-blocking sessions, the asset admin refetch request was being fired off before the save request had finished processing. This resulted in the old information being displayed instead of the new post-save information.

Fixes https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/14743937873/job/41392697219.

Needed for https://github.com/silverstripe/silverstripe-framework/issues/11671 but likely also causes problems when dynamodb or hybridsessions is used.

> [!NOTE]
> Ignore the thumbnail request in the below waterfall graphs which is not related to this bug or the fix.

This network waterfall graph shows the two requests are overlapping before this PR with non-blocking sessions:
image

This second network waterfall graph shows the requests no longer overlap with this fix in place:
image

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1562


client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -370,6 +367,10 @@ class AssetAdmin extends Component {
           this.handleBrowse(response.record.parent.id, response.record.id, null);
         }
         return response;
+      }).finally(() => {
```
emteknetnz:
```suggestion
      }).then(() => {
```
Any reason why we're using finally() instead of regular then()?

Presumably if there's a hard failure in the previous then() block things are already in a broken so ensure the refetch happens probably won't fix things a whole lot

emteknetnz (changes requested):

client/src/containers/AssetAdmin/AssetAdmin.js
ChloeSartorelli: Before this PR, it was _always_ refetching. Using `finally` ensures that's still the case - it will still _always_ refetch. This PR just changes _when_ it refetches (i.e. after the other request has fully completed)

emteknetnz: Yeah I get that, it just looks weird having a finally(), it implies we almost expect something to fail in the above catch() block.  It just makes me read like it failing is OK somehow, or that we MUST fetch. Neither of those is really the case here.

I think it's cleaner to just have a regular then() here instead

ChloeSartorelli: I honestly don't know whether we need to always fetch or not - I wanted to keep that the same so I don't have to look into the failure states and whether refetching is necessary or not. My assumption is that the fact it always refetched before means there is a reason why we should always refetch.

`then()` is cleaner _if_ we shouldn't refetch on failure - but I don't know if we should do that or not.

ChloeSartorelli: Will use `then()` for now. Fingers crossed there's not an edge case this breaks 😬

emteknetnz (approved):

Issue #1434, opened 2024-01-18
UploadField doesn't validate allowed extensions with "choose existing" option


mark-a-j-adriano:
### Module version(s) affected

2.1.3

### Description

Allowed extensions validation for `UploadField` does not kick in when using the "choose existing" option

https://github.com/silverstripe/silverstripe-asset-admin/assets/11882563/9f2a02f8-ed41-4552-8359-684d4e418304

### How to reproduce

1. Add an upload field like this to a form:
    ```php
    /** @var UploadField $logo */
    $logo = UploadField::create('Logo', 'Logo');
    $logo->setAllowedExtensions(['png', 'svg'])
        ->setAllowedMaxFileNumber(1)
        ->setUploadEnabled(false)
        ->setDescription('Accepts PNG or SVG image format.');
    ```
1. Add a file (not `png` or `svg`) to the field using the "choose existing" option
1. Note that the file is added to the field without raising a validation error.

**NOTE** the `setAllowedMaxFileNumber` is probably unrelated, and `setUploadEnabled` is not necessary.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## Notes
- That validation is currently handled by the upload validator - which is for validating _uploaded_ files, not existing files
- It would make sense for the allowed extensions validation to be triggered for existing files. It's not uncommon for business logic to be "this relation should only be files of this type" (e.g. videos only, or documents only, etc)
- This is a fairly substantial change in logic, it should probably be introduced in a minor release rather than a patch so people can be alerted to this change via the changelog.

## Acceptance criteria
- User is blocked from selecting pre-existing files that do not match the allowed extensions.
- All file validation rules are honoured when selecting a pre-existing file
- The Browser upload file dialog stops you from uploading files that do not have one of the allowed extensions. (What should happen we have the default validation rules? Should all the file upload input include all the CMS allowed types?)

## UX questions
- What happens if you select a file that does not meet the validation criteria? How do we communicate to the user that they cannot select this file?

[Kitchen sink CI](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/15174484044)

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1565
- https://github.com/silverstripe/silverstripe-frameworktest/pull/234

[...]

PR #1565, opened 2025-05-15
FIX Validate file extension when selecting existing files

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1434

Requires https://github.com/silverstripe/silverstripe-frameworktest/pull/234 for behat


client/src/components/UploadField/UploadField.js
```diff
@@ -324,9 +325,20 @@ class UploadField extends Component {
...
+      this.props.actions.toasts.error(
+        i18n.sprintf(
+          // This matches the text used in Upload_Validator::validate()
+          i18n._t('AssetAdmin.UPLOADFIELD_INVALID_EXTENSION', "Extension '%s' is not allowed"),
+          file.extension.toUpperCase(),
+        )
+      );
+    } else {
+      this.props.actions.uploadField.addFile(this.props.id, file);
+    }
```
ChloeSartorelli: Do we also have server-side validation for this? If not, someone can bypass this by modifying the POST request. We shouldn't trust information coming from the client.

ChloeSartorelli (changes requested):
Haven't really reviewed beyond the below question as it may drastically change the implementation.

client/src/components/UploadField/UploadField.js
ChloeSartorelli: Ultimately I think the fact that file extension validation isn't happening is just a symptom - really _all_ validation that the `UploadField` has should trigger. E.g. if someone has some custom validation, that should also be triggered.

Imagine someone adds validation to only allow images of specific dimensions to be added to a field, because those dimensions are necessary for the image to be correctly displayed. Selecting an already-uploaded file shouldn't bypass that validation.

code/Forms/UploadField.php
ChloeSartorelli, line 1: It looks like you've reimplemented the logic for validating file extensions, is that correct?
It feels like we should be using the existing validator assigned to this field (the same one that validates uploads) if possible, which:
1. means we don't have two ways to validating the same thing
2. means we validate _all_ things equally, not just the file extension. See https://github.com/silverstripe/silverstripe-asset-admin/pull/1565#discussion_r2094649426

ChloeSartorelli (changes requested):

code/Forms/UploadField.php
emteknetnz: The problem with validating all the things is that the existing validator that's used, MimeUploadValidator which extends Upload_Validator, is expecting a tmpFile in a lot of places e.g. `$this->tmpFille['error']` - it's designed for Uploads, not selecting existing files.

We could refactor the whole thing so that validation logic was agnositic as to if it was tmpFile or an existing file, however that'll be a significant piece of work and out of scope for this bugfix pr

Note that the Upload_Validator logic for allowed extensions is basically just `in_array($this->getFileExtension(), $this->allowedExtensions)` so we're not exactly duplicating much code

client/src/components/UploadField/UploadField.js
emteknetnz: I've added server side validation

Re "all the validation" - discussion for that should be had [here](https://github.com/silverstripe/silverstripe-asset-admin/pull/1565/files#r2096798134)

code/Forms/UploadField.php
ChloeSartorelli: Looks like the only keys that you need in the tmp file array are
- `error` (which should be empty)
- `size` (contains the file size)
- `name` (the file name)
- `extension` (the file extension)

So can't you just craft that array and then ask the existing validator to validate it? Something like this:

```php
/** @var File $file */
foreach ($this->getItems() as $file) {
    $fileValidator = $this->getValidator();
    $fileValidator->setTmpFile([
        'error' => '',
        'name' => $file->getFilename(),
        'size' => $file->getSize(),
        'extension' => $file->getExtension(),
    ]);
    if (!$fileValidator->validate()) {
        $errors = $fileValidator->getErrors();
        // Do something with the errors.
    }
}
```

ChloeSartorelli: Actually I may need to think about this a little more, since obviously with a tmp file you have an actual file on the filesystem but with assets they may be stored anywhere, so existing custom validators may rely on that filesystem access.

ChloeSartorelli: Okay yeah https://github.com/BiffBangPow/silverstripe-imagesize-validator/ for example would stop working if we did that.

We've also missed the opportunity to rework file validation in CMS 6. I've [created a card](https://github.com/silverstripe/silverstripe-asset-admin/issues/1570) to look into it for CMS 7.

code/Forms/UploadField.php
```diff
@@ -326,28 +328,52 @@ public function performDisabledTransformation()
...
+        $invalidExts = array_values(array_unique($invalidExts));
+        foreach ($invalidExts as $ext) {
+            $result = false;
+            $validator->validationError($this->getName(), _t(
+                __CLASS__ . '.InvalidExtension',
+                "Extension '{ext}' is not allowed.",
+                [
+                    'ext' => $ext,
+                ]
+            ));
```
ChloeSartorelli: Use the existing i18n key from https://github.com/silverstripe/silverstripe-assets/blob/8b93103ce0aff0da16ac577c2616858aa5ec4511/src/Upload_Validator.php#L360-L367

code/Forms/UploadField.php
```diff
@@ -326,28 +328,52 @@ public function performDisabledTransformation()
...
     public function validate($validator)
     {
         $maxFiles = $this->getAllowedMaxFileNumber();
-        $count = count($this->getItems() ?? []);
-
-        if ($maxFiles < 1 || $count <= $maxFiles) {
-            return $this->extendValidationResult(true, $validator);
+        /** @var SS_List<File> $items */
+        $items = $this->getItems();
+        $count = count($items ?? []);
```
ChloeSartorelli:
```suggestion
        $count = count($items);
```
A new `ArrayList` is returned if there's no items, so this won't ever be null

ChloeSartorelli (changes requested):
A few problems stand out right now:

1. If I remove the client-side validation (to test the server-side validation) the invalid file is removed from the upload field, and my record gets saved. The toast says it was successful and there's no indication there were any validation errors.
2. If a record has multiple `UploadField` and one of them has a file that won't pass validation anymore, saving the record will only give a toast, but it won't be obvious what field had the invalid file.
3. I get a toast message for each invalid file even if they're the same extension.


Probably the easiest way forward is to get the server-side validation working (with displaying its messages next to the field) and just remove the client-side validation.

code/Forms/UploadField.php
emteknetnz: Updated

code/Forms/UploadField.php
emteknetnz: Changed to $items->count() since it's an SS_List

code/Forms/UploadField.php
```diff
@@ -119,6 +119,8 @@ public function getSchemaDataDefaults()
             'payloadFormat' => 'urlencoded',
         ];
 
+        // use array_values to ensure 0-based so does not get coverted to a JS object
+        $defaults['data']['allowedExtensions'] = array_values($this->getAllowedExtensions() ?: []);
```
ChloeSartorelli: Is this still needed? Was originally added for the client-side validation which has since been removed.

ChloeSartorelli:
When selecting an invalid file and saving: the invalid file is removed from the upload field, and my record gets saved. The toast says it was successful and there's no indication there were any validation errors

[Screencast from 22-05-25 12:35:32.webm](https://github.com/user-attachments/assets/c6610113-6739-4dbe-931f-f85d8c0e5ba7)

emteknetnz:
Works correctly on my loacl and on behat. Could you share the code you're using to test

ChloeSartorelli:
No custom code, just the upload field available in the profile controller (for single upload) and `Company` class (for multi-upload) when frameworktest is installed.

emteknetnz:
The profile field (not sure where this comes from, frameworktest?) doesn't disallow an extension? i.e. `setAllowedExtensions()` is not called. Nor does the Company image field in frameworktest? The new extension in frameworktest needs to be explicitly added to disallow file extensions (this happens in the behat test)

ChloeSartorelli:
They both only allow images. So any extensions other than image are not allowed.

The client-side code you removed was throwing toasts, because files with non-image extensions weren't allowed.

emteknetnz:
I'm inclined to say that should be treated as a different issue. After a lot of debugging where that's coming from is in `FileUploadReceiver::setValue()` (trait that's put on to UploadField) after being called from `Form::setSubmittedValue()` and `Form::loadDataFrom()` i.e. scaffolding the form pre validation

The specific line where it happens is `$items = DataObject::get($class)->byIDs($value['Files']);` where `$value = ['Files' => <FileID>]` and `$class = $this->getRelationAutosetClass();`, which in this case will be Image::class.

So basically when you upload say a PDF on an UploadField backed by a DBImage, because there's so corresponding record for that image, it simply doesn't find a record. Therefore the form field being scaffolded is never populated and therefore never validated that the file has the wrong extension

It would be wrong to throw a ValidationException at that point set it's only FormField::setValue()

A couple of solutions:
- Do it with frontend validation (not toasts which has already been rejected), which could messy for a UX point of view, particularly with multi-file uploads where some files are valid and some are not)
- Filter out all non-images in the AssetAdmin modal so that you cannot select an Image.  Possibly that could be done with file extensions rather than the dataclass to make it more flexible

Either way what this PR has right now is an improvement over the status quo in that setAllowExtensions() does something for selecting existing images now. I don't particularly feel like embarking on either of those other two solutions at this point, so we should make a new card if we want to do those.

ChloeSartorelli:
Ahhh okay. Of course something called `FileUploadReceiver` is messing with files that _aren't being uploaded_. 😮‍💨

Thank you for looking into that and the explanation. I agree that is a separate issue, because of the way it was implemented.
I would be tempted to even say it's a bug that `FileUploadReceiver` has _anything_ to do with pre-existing files, except that people are probably relying on that to some extent so changing that would end up being a breaking change.

I've added that inconsistency onto https://github.com/silverstripe/silverstripe-asset-admin/issues/1570. We will probably want to break that up when it comes to it but I think having all of the context about file validation inconsistencies in one place is best for now.

ChloeSartorelli (commented):
Works as expected, just one question (sorry I thought this had been sent last review cycle)

code/Forms/UploadField.php
emteknetnz: Removed

ChloeSartorelli:
CI failed after merging frameworktest PR

ChloeSartorelli (approved):
LGTM, works as expected locally.

Issue #1577, opened 2025-06-06
CMS 6 REGRESSION: Sort order is all over the place.


ChloeSartorelli:
Initial order seems to be fairly arbitrary. The pagination is done, and _then_ further sorting is done within the page.

The result is that within a page the sort order respects what the user selected, but across pages the order is effectively random.

Example here has limited to 10 items per page and has files named `file001` through to `file060`, and sort order _should_ be by title ascending. Note that several files are skipped (e.g. files 13 through 16) - those are on different pages.

![Image](https://github.com/user-attachments/assets/d23bc6c7-f093-4cbc-a96f-e08ae000cf48)

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1578

ChloeSartorelli:
Fixing this is quite easy - but we're faced with a new problem.

The default sort order for files is by the `Name` field - which is the name of the file and is _separate_ from the `Title` field.

The sort order options in the assets section show `Title A-Z` by default.
![Image](https://github.com/user-attachments/assets/995360ca-84f9-40d0-bb5f-18f9eedd55f0)

This is a bug that has been around for who-knows-how-long. The question is: Do we change the default sort order of `File` to be `Title ASC` - or leave it alone, knowing it doesn't match what the UI suggests it is?

Edit: There was already new CMS 6 logic that uses `Title ASC` only within AssetAdmin. While it would probably make more long-term sense to change the default sort order at a model level, that's not in scope.

PR #1578, opened 2025-06-08
FIX Make sure sort works correctly across pages

ChloeSartorelli:
Brings back the sort query logic from CMS 5 - see [`FolderTypeResolver::sortChildren()`](https://github.com/silverstripe/silverstripe-asset-admin/blob/12aba732ff17aa37fb9ee612a3526319c1ca5315/code/GraphQL/Resolvers/FolderTypeResolver.php#L164-L188)

I have retained the new CMS 6 logic that uses `Title ASC` as the default sort inside AssetAdmin when none is defined, instead of falling back to the default_sort.

## Issues

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1577

### DIFF
--- a/code/Controller/AssetAdmin.php
+++ b/code/Controller/AssetAdmin.php
@@ -116,14 +116,6 @@ class AssetAdmin extends AssetAdminOpen implements PermissionProvider
     ];
 
     /**
-     * Amount of results showing on a single page.
-     *
-     * @config
-     * @var int
-     */
-    private static $page_length = 50;
-
-    /**
      * @config
      *
      * @var int
--- a/code/Controller/AssetAdminOpen.php
+++ b/code/Controller/AssetAdminOpen.php
@@ -17,6 +17,7 @@ use stdClass;
 use SilverStripe\ORM\DataList;
 use SilverStripe\Forms\DateField;
 use SilverStripe\ORM\DataQuery;
+use SilverStripe\ORM\DB;
 use SilverStripe\View\Requirements;
 use SilverStripe\ORM\Hierarchy\Hierarchy;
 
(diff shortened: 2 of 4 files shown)


code/Controller/AssetAdmin.php
```diff
@@ -115,14 +115,6 @@ class AssetAdmin extends AssetAdminOpen implements PermissionProvider
         'apiUploadFile',
     ];
 
-    /**
-     * Amount of results showing on a single page.
-     *
-     * @config
-     * @var int
-     */
-    private static $page_length = 50;
```
ChloeSartorelli, line 124: This is already defined on `AssetAdminOpen`. Defining it here as well just makes it harder to override with no benefit. Done in this PR because I need to override it for the behat test.

code/Controller/AssetAdminOpen.php
```diff
@@ -151,43 +152,54 @@ protected function getFileData(
...
                 $offset = $limit * ($page - 1);
             }
-            $totalCount = $childFiles->count();
-            // Folders show first in paginated results, no matter the sort order
-            // it's assumed that no one will every subclass a Folder
-            $childFiles = $childFiles->alterDataQuery(function (DataQuery $query) {
-                $query->selectField('CASE WHEN "File"."ClassName" = \'SilverStripe\\\\Assets\\\\Folder\' THEN 1 ELSE 0 END AS "IsFolderTmp"');
-                $query->sort('"IsFolderTmp" DESC', null, true);
-            });
             $childFiles = $childFiles->limit($limit, $offset);
```
ChloeSartorelli, line 163: Moved limit up here - the actual order doesn't matter but we got the limit and offset here, so makes sense to use it immediately.

code/Controller/AssetAdminOpen.php
```diff
@@ -151,43 +152,54 @@ protected function getFileData(
         $totalCount = 0;
         if ($includeChildren) {
             $childFiles = $this->getFilteredChildFiles($file, $request);
+            $totalCount = $childFiles->count();
```
ChloeSartorelli, line 155: Moved the count to be directly here so we don't risk putting logic that alters the query before the total count.

code/Controller/AssetAdminOpen.php
```diff
@@ -151,43 +152,54 @@ protected function getFileData(
...
+            // Prepare sort order (default to Title ASC to match UI)
+            $sortArgs = $this->getQueryStringValue($request, 'sort', true);
+            $sortField = 'Title';
+            $sortDir = 'ASC';
+            if ($sortArgs) {
+                [$sortField, $sortDir] = explode(',', $sortArgs);
+                if (!in_array($sortField, ['title', 'lastEdited']) || !in_array($sortDir, ['asc', 'desc'])) {
+                    $this->jsonError(404);
+                }
+            }
```
ChloeSartorelli, line 174: Retained from existing CMS 6 logic

code/Controller/AssetAdminOpen.php
```diff
@@ -151,43 +152,54 @@ protected function getFileData(
...
+                    // Clear all existing sort
+                    true
+                );
+
+                if ($sortField && $sortDir) {
+                    $dataQuery->sort($sortField, $sortDir, false);
+                }
+
+                return $dataQuery;
+            });
```
ChloeSartorelli, line 194: CMS 5 logic to get folders first, then sort by the chosen sort order

tests/behat/features/change-sort-order.feature
```diff
@@ -4,59 +4,57 @@ Feature: Change view for asset admin
   I want to change the way I'm viewing files
 
   Background:
-    Given a "file" "folder1/document.pdf"
-      And a "image" "folder1/file1.jpg"
-      And a "image" "folder1/file2.jpg"
+    Given a "file" "file001" has "Filename"="folder1/document.pdf" and "Title"="file001"
+      And a "image" "file002" has "Filename"="folder1/file1.jpg" and "Title"="file002"
+      And a "image" "file003" has "Filename"="folder1/file2.jpg" and "Title"="file003"
```
ChloeSartorelli, line 9: Giving the files very clearly ordered names so it's easier to tell that the order is as expected

tests/behat/features/change-sort-order.feature
```diff
@@ -69,28 +67,60 @@ Feature: Change view for asset admin
...
+      And I go to "/admin/assets"
+      And I click on the file named "folder1" in the gallery
+    Then I should see the file named "zsubfolder2" in the gallery
+      And I should see the table gallery folder "xsubfolder1" in position "1"
+      And I should see the table gallery folder "zsubfolder2" in position "2"
+      And I should see the gallery item "file001" in position "1"
+      And I should see the gallery item "file002" in position "2"
+      # 10 total file and folders - so 8 files excluding folders
+      And I should see the gallery item "file008" in position "8"
+    When I press the "Next" button
```
ChloeSartorelli, line 112: The original tests are only working within a single page. This new scenario tests that sort works as expected _across_ pages which is what the regression was.

emteknetnz (approved):
Works well

Issue #1576, opened 2025-06-05
CMS 6 REGRESSION: Cannot filter by "file type"


ChloeSartorelli:
In CMS 5 you could filter assets by "file type" (e.g. only images, only documents).

In CMS 6 that option is there but always returns no results.

Note that other filters work as expected - it's just the "file type" option that's not working.

Part of fixing this bug includes adding a behat test for it.

## Note

Until https://github.com/silverstripe/silverstripe-asset-admin/issues/1573 is fixed, filtering in an UploadField modal will do _literally nothing_ - but the code for the actual filtering is all the same.

The filtering behat test will be updated in the fix PR for #1573 to include checking modals.

## Steps to reproduce

1. Go to `/admin/assets/` OR click "choose existing" inside an upload field
2. If no files have yet been uploaded, upload some files of various different types (images, pdfs, etc)
3. Click the magnifying glass
4. Click "search options"
5. Click the "file type" dropdown and select a type that you have uploaded
6. Click "Search"

Note that "No items found" is displayed, instead of a valid set of files.

## PRs

[**Installer CI run**](https://github.com/creative-commoners/silverstripe-installer/actions/runs/15524972974)

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1579
- https://github.com/silverstripe/silverstripe-behat-extension/pull/305

PR #1579, opened 2025-06-09
FIX Correctly filter by category

ChloeSartorelli:
The original code for this seems to have been pulled from [`FileFilter::filterList()`](https://github.com/silverstripe/silverstripe-asset-admin/blob/12aba732ff17aa37fb9ee612a3526319c1ca5315/code/GraphQL/FileFilter.php#L103-L107) - but the enum that comment references is a GraphQL enum which is created in [`Builder::updateSchema()`](https://github.com/silverstripe/silverstripe-asset-admin/blob/12aba732ff17aa37fb9ee612a3526319c1ca5315/code/GraphQL/Schema/Builder.php#L31-L42).

So in CMS 6 before this PR we're just doing a filter like `->filter('Name:EndsWith', 'IMAGE')` (we haven't mapped to the extensions yet).

Requires https://github.com/silverstripe/silverstripe-behat-extension/pull/305 to be merged for new CI to go green

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1576

### DIFF
--- a/code/Controller/AssetAdminOpen.php
+++ b/code/Controller/AssetAdminOpen.php
@@ -411,9 +411,20 @@ class AssetAdminOpen extends LeftAndMain
             $list = $list->filter("Created:LessThanOrEqual", $toDate->dataValue().' 23:59:59');
             $search = true;
         }
-        // Categories (mapped to extensions through the enum type automatically)
+        // Categories (mapped to extensions)
         if (!empty($filter['appCategory'])) {
-            $list = $list->filter('Name:EndsWith', $filter['appCategory']);
+            $category = $filter['appCategory'];
+            $categoryValues = File::config()->get('app_categories');
+            // Sanitise category names (some contain slashes) and make all upper case
+            foreach ($categoryValues as $key => $extensions) {
+                unset($categoryValues[$key]);
+                $newKey = strtoupper(preg_replace('/[^a-zA-Z0-9]/', '', $key));
+                $categoryValues[$newKey] = $extensions;
+            }
+            if (!isset($categoryValues[$category])) {
+                $this->jsonError(404);
+            }
+            $list = $list->filter('Name:EndsWith', $categoryValues[$category]);
             $search = true;
         }
         // Filter unknown id by known child if search is not applied
--- a/tests/behat/features/change-sort-order.feature
+++ b/tests/behat/features/change-sort-order.feature
@@ -62,7 +62,7 @@ Feature: Change view for asset admin
       And I am logged in as a member of "EDITOR" group
       And I go to "/admin/assets"
 
-  Scenario: I can switch the sorting order in table view
+  Scenario: I can switch the sorting order in gallery view
     When I click on the file named "folder1" in the gallery
       And I click on the ".gallery__sort a" element
       And I wait until I see the ".gallery__sort .chosen-results" element
--- a/tests/behat/features/filter-types.feature
+++ b/tests/behat/features/filter-types.feature
@@ -1,0 +1,120 @@
+@assets @retry @job1
+Feature: Filter in asset admin
+  As a cms author
+  I want to filter files in various ways
+
+  Background:
+    Given a "file" "file001" has "Filename"="folder1/document.pdf" and "Title"="file001"
+      And a "image" "file002" has "Filename"="folder1/file1.jpg" and "Title"="file002"
+      And a "image" "file003" has "Filename"="folder1/file2.jpg" and "Title"="file003"
+      # Adding multiple versions of folder/testfile.jpg is intentional
+      # We do this to avoid having to provide so many files in tests/behat/files/
+      And a "image" "file004" has "Filename"="folder1/testfile.jpg" and "Title"="file004"
+      And a "image" "file005" has "Filename"="folder1/testfile.jpg" and "Title"="file005"
+      And a "image" "file006" has "Filename"="folder1/testfile.jpg" and "Title"="file006"
+      And a "image" "file007" has "Filename"="folder1/testfile.jpg" and "Title"="file007"
+      And a "image" "file008" has "Filename"="folder1/testfile.jpg" and "Title"="file008"
+      And a "image" "file009" has "Filename"="folder1/testfile.jpg" and "Title"="file009"
+      And a "image" "file010" has "Filename"="folder1/testfile.jpg" and "Title"="file010"
+      And a "image" "file011" has "Filename"="folder1/testfile.jpg" and "Title"="file011"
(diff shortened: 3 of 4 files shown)


code/Controller/AssetAdminOpen.php
```diff
@@ -411,9 +411,19 @@ private function filterFiles(DataList $list, array $filter): DataList
...
+            $categoryValues = File::config()->get('app_categories');
+            // Sanitise category names (some contain slashes) and make all upper case
+            foreach ($categoryValues as $key => $extensions) {
+                unset($categoryValues[$key]);
+                $newKey = strtoupper(preg_replace('/[^[[:alnum:]]]*/', '', $key));
+                $categoryValues[$newKey] = $extensions;
+            }
+            if (!isset($categoryValues[$filter['appCategory']])) {
+                $this->jsonError(404);
+            }
```
ChloeSartorelli: In CMS 5 using an invalid category would return an error.
image

I honestly think error code `400` is more appropriate here, but for some reason we're using `404` for errors in this controller, so I'm using `404` here for consistency.

tests/behat/features/change-sort-order.feature
```diff
@@ -62,7 +62,7 @@ Feature: Change view for asset admin
       And I am logged in as a member of "EDITOR" group
       And I go to "/admin/assets"
 
-  Scenario: I can switch the sorting order in table view
+  Scenario: I can switch the sorting order in gallery view
```
ChloeSartorelli, line 65: This was confusing me because the "list" view is the one in a table. I checked the UI and it seems they're called "list view" (the one in a table, which _isn't_ this one), and "gallery view".

tests/behat/features/filter-types.feature
```diff
@@ -0,0 +1,120 @@
+@assets @retry @job1
```
ChloeSartorelli, line 1: job1 is the fastest job

tests/php/Controller/AssetAdminOpenTest.php
```diff
@@ -112,6 +112,14 @@ public static function provideApiRead(): array
...
                 'expectedCode' => 404,
             ],
+            'Reject invalid category' => [
+                'idType' => 'existing',
+                'fail' => '',
+                'filter' => 'filter[appCategory]=BANANA',
+                'page' => 0,
+                'pageLength' => 0,
+                'expectedCode' => 404,
+            ],
```
ChloeSartorelli, line 122: Can't sensibly test this with behat.
I was gonna write unit tests for the actual filtering as well but the existing tests aren't set up for that, so it didn't seem worth it given behat will test (most) of that.

tests/behat/features/filter-types.feature
ChloeSartorelli, line 1: Can't find a sensible way to test the last edited date filters.

code/Controller/AssetAdminOpen.php
```diff
@@ -411,9 +411,19 @@ private function filterFiles(DataList $list, array $filter): DataList
...
+        // Categories (mapped to extensions)
         if (!empty($filter['appCategory'])) {
-            $list = $list->filter('Name:EndsWith', $filter['appCategory']);
+            $categoryValues = File::config()->get('app_categories');
+            // Sanitise category names (some contain slashes) and make all upper case
+            foreach ($categoryValues as $key => $extensions) {
+                unset($categoryValues[$key]);
+                $newKey = strtoupper(preg_replace('/[^[[:alnum:]]]*/', '', $key));
+                $categoryValues[$newKey] = $extensions;
+            }
```
ChloeSartorelli: This is basically taken verbatim from [Builder::updateSchema()](https://github.com/silverstripe/silverstripe-asset-admin/blob/12aba732ff17aa37fb9ee612a3526319c1ca5315/code/GraphQL/Schema/Builder.php#L31-L42) - except that code puts the extensions into a sub-array with a `value` key which was specifically for the graphql enum and isn't needed here.

code/Controller/AssetAdminOpen.php
```diff
@@ -411,9 +411,19 @@ private function filterFiles(DataList $list, array $filter): DataList
...
         }
-        // Categories (mapped to extensions through the enum type automatically)
+        // Categories (mapped to extensions)
         if (!empty($filter['appCategory'])) {
-            $list = $list->filter('Name:EndsWith', $filter['appCategory']);
+            $categoryValues = File::config()->get('app_categories');
+            // Sanitise category names (some contain slashes) and make all upper case
+            foreach ($categoryValues as $key => $extensions) {
+                unset($categoryValues[$key]);
+                $newKey = strtoupper(preg_replace('/[^[[:alnum:]]]*/', '', $key));
```
emteknetnz:
```suggestion
                $newKey = strtoupper(preg_replace('/[^a-zA-Z0-9]/', '', $key));
```
We don't use `[:alnum:]` anywhere else in the codebase. I had to lookup what exact characters it represented. I think it's better to just explicitly list the chars.  We can also get rid of 2 sets of brackets and the asterisk this way

code/Controller/AssetAdminOpen.php
```diff
@@ -411,9 +411,19 @@ private function filterFiles(DataList $list, array $filter): DataList
...
+                $categoryValues[$newKey] = $extensions;
+            }
+            if (!isset($categoryValues[$filter['appCategory']])) {
```
emteknetnz:
```suggestion
            $appCategory = $filter['appCategory'];
            if (!isset($categoryValues[$appCategory])) {
```
Worth expanding out to make it easier to read

code/Controller/AssetAdminOpen.php
```diff
@@ -411,9 +411,19 @@ private function filterFiles(DataList $list, array $filter): DataList
...
+                $this->jsonError(404);
+            }
+            $list = $list->filter('Name:EndsWith', $categoryValues[$filter['appCategory']]);
```
emteknetnz:
```suggestion
            $list = $list->filter('Name:EndsWith', $categoryValues[$appCategory]);
```

emteknetnz (changes requested):

code/Controller/AssetAdminOpen.php
ChloeSartorelli: I don't think it is worth it - that's not hard to read at all.
But done just to avoid additional ping pong  since there was the other change anyway

code/Controller/AssetAdminOpen.php
ChloeSartorelli: This was pre-existing code from CMS 5 - but this seems a sensible enough change. Done.

emteknetnz (approved):

Issue #1585, opened 2025-06-30
"Drag and Drop" displays at top - cut off if scrolled down


ChloeSartorelli:
> [!IMPORTANT]
> See https://github.com/silverstripe/silverstripe-asset-admin/issues/1585#issuecomment-3199075401 below

If you scroll down before dragging the files to upload them into the asset admin section, the "Drag and Drop" image is displayed only at the top of the page and gets cut off.

![Image](https://github.com/user-attachments/assets/dc4266a1-ee8c-4aa9-a3e8-c711c94856ea)

This is exacerbated if you increase the number of files per page (configurable) or upload many files in a row, in which case you may not see the image at all.

Note this definitely happens in CMS 6 - but check if it happens in CMS 5 before implementing a fix.

## Acceptance criteria

- [x] The "Drag and Drop" image is displayed in full wherever you are scrolled to on the page.

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1599
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1606

ChloeSartorelli:
PR merged. Will be automatically tagged by GitHub Actions

ChloeSartorelli:
Reopening because the previous fix actually broke it worse lol.

Before the previous PR:

https://github.com/user-attachments/assets/8a6530e5-cb10-4285-b768-be510b8e267c

https://github.com/user-attachments/assets/5d7c02ef-8ad2-421d-a8df-9aebe4b01d1c

After the previous PR:

https://github.com/user-attachments/assets/8b412a16-c6f1-47de-a844-abe9ff74e24d

https://github.com/user-attachments/assets/f3865601-c4f8-4111-b563-cf9d8a01d830

ChloeSartorelli:
PR merged. Way better 👍

PR #1599, opened 2025-08-13
FIX Align drag and drop image with viewport scroll

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1585

Affects 5.4 as well as 6


ChloeSartorelli (approved):
LGTM, works as expected locally

Issue #839, opened 2018-09-18
Files list loads when clicking a file for the first time


ScopeyNZ:
When you first load assets, then click a file in the list for the first time, it will load the sidebar (great!) and it will also load the file list (why?). This won't happen on subsequent presses of the same item but will happen on the first click of any item in the list.

This was raised during a CMS author training. It wasn't particularly noticable until you have many files (and maybe some average hosting) and then the load time of the list was considerable enough for it to be a usability issue.

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1603

chillu:
Can reproduce this - I've highlighted the second list load (`/graphql` request). As far as I can tell it's not a high impact to the user - those two requests have separate loading indicators, so you can edit the file even though the list is still loading. Still, good catch, we should fix this!

<img width="569" alt="image" src="https://user-images.githubusercontent.com/111025/46268151-6af46280-c595-11e8-86de-d789a72297ac.png">

@bergice Since you're already in the depths of asset admin state management, might be a good one to pick up? :)

ChloeSartorelli:
Can reproduce in both CMS 5 and CMS 6.

In CMS 5 it seems to be related to the way the GraphQL injection works and I spent way too long trying in vain to find a solution before giving up.
Fixed for CMS 6.

emteknetnz:
PR works well, will be automatically tagged shortly

PR #1603, opened 2025-08-19
FIX Don't refetch files unnecessarily

ChloeSartorelli:
We don't need to refetch the full list of files just to view the edit form of a file.

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/839


tests/behat/features/file-upload-state.feature
```diff
@@ -15,6 +15,7 @@ Feature: Upload file state
     When I click on the file named "folder1" in the gallery
     Then I should see the gallery item "file2" in position "1"
     When I attach the file "testfile.jpg" to dropzone "gallery-container"
+    And I wait for 1 second
```
ChloeSartorelli: Without this, the mysql80 job1 behat build was failing (but mysql84 build passed).
The screenshot shows the file was in position 1. My local testing has it in position 1 very briefly before being moved to the correct position.

ChloeSartorelli: Nevermind, it's still broken even with this, for some reason...

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -602,11 +604,7 @@ class AssetAdmin extends Component {
   }
 
   handleUploadQueue() {
-    // A bit of coupling. If the editor isn't open, the gallery will automatically
-    // open a file and force a refresh, so we have to guard against a double refresh.
```
ChloeSartorelli, line 606: That double refresh doesn't happen anymore, since `handleBrowse()` and `componentDidUpdate()` no longer force a refresh when loading the edit form.

emteknetnz (approved):

PR #1605, opened 2025-08-21
ENH Disable sorting on queries where sort order doesn't matter

ChloeSartorelli:
Sorting queries can be slow, especially on large datasets and especially if indexes aren't optimised. We should avoid sorting when the sort order doesn't matter.

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11833

### DIFF
--- a/code/Controller/AssetAdmin.php
+++ b/code/Controller/AssetAdmin.php
@@ -720,10 +720,12 @@ class AssetAdmin extends AssetAdminOpen implements PermissionProvider
 
     /**
      * Retrieves a list of files by their IDs
+     * Does not respect default sort order
      */
     private function getFilesByIDs(array $ids, bool $includeFolders, int $missingFileErrorCode): ArrayList
     {
-        $files = File::get()->filter(['ID' => $ids]);
+        // Disabling sort improves performance
+        $files = File::get()->sort(null)->filter(['ID' => $ids]);
         if ($files->count() !== count($ids)) {
             $this->jsonError($missingFileErrorCode);
         }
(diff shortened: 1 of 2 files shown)


code/Controller/AssetAdmin.php
```diff
@@ -720,10 +720,12 @@ public function apiHistory(HTTPRequest $request): HTTPResponse
 
     /**
      * Retrieves a list of files by their IDs
+     * Does not respect default sort order
      */
     private function getFilesByIDs(array $ids, bool $includeFolders, int $missingFileErrorCode): ArrayList
     {
-        $files = File::get()->filter(['ID' => $ids]);
+        // Disabling sort improves performance
+        $files = File::get()->sort(null)->filter(['ID' => $ids]);
```
ChloeSartorelli, line 728: Only used in places where order doesn't matter, e.g. deleting, moving, publishing, unpublishing.

code/Controller/AssetAdminFile.php
```diff
@@ -225,7 +226,7 @@ private function getDescendantFileIDs(array $ids, int $maxDepth = 10): array
         if ($maxDepth === 0) {
             return $ids;
         }
-        $childIDs = File::get()->filter('ParentID', $ids)->column('ID');
+        $childIDs = array_values(File::get()->filter('ParentID', $ids)->getIDList());
```
ChloeSartorelli: Only used in a filter in `getDescendantFileCount()` above.

emteknetnz (approved):

Issue #1022, opened 2019-10-14
Screen reader use is problematic in the Files modal


sachajudd:
At the moment when the files modal is opened the focus is set on the whole dialog causing all of the information in the modal to be read out at once for a screen reader user.
It would be good to investigate if we could set the focus to the “Files” title when the modal is opened so a screen reader can tab through the appropriate fields themselves.

Possibly related: #958

sachajudd:
@clarkepaul could this be added to any future experience debt epics?

PR #1609, opened 2025-09-22
ENH Autofocus modal title

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1022


ChloeSartorelli (commented):
Used Orca screen reader - both before and after this PR it reads out a bunch of extra information, but with this PR it reads out a lot _more_ unnecessary information. Did you test this PR with a screen reader? At least with Orca, the experience seems to be worse with the PR than without it. Haven't tested with NVDA because I don't have windows at work, and haven't tested with JAWS because that's a paid product.

Another worry is that this change makes this one modal inconsistent with other modals, where the breadcrumbs won't have autofocus.

For some modals autofocus is intentionally given to the first input element in the modal, which is correct and consistent with [W3C's accessible modal example](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/).

For any modal without an input field though, having modals read out information inconsistently will probably add confusion.

emteknetnz:
Seems like we should just close all the PRs then?

ChloeSartorelli:
> Seems like we should just close all the PRs then?

Not necessarily? It depends if this is resolving a problem or not, which is why I asked "Did you test this PR with a screen reader?"

The Orca screenreader isn't great - ideally I'd like to see this tested with NVDA or JAWS (or both) since those are the two most popular screen readers.

If there is a real problem here, and this solves it for this modal, it's just a matter of making sure it's resolved in a consistent way for all modals.
If there is no problem (the issue was quite old, maybe screen readers got better since then) then in that case yes closing the PRs seems appropriate.

emteknetnz:
I've made a note on the parent issue that's I'm moving this issue back to our refinement column so that we can work out our screen reader testing and development approach first

PR #1610, opened 2025-09-23
FIX Show error on failed upload of TableView

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/454

CMS 6 version of https://github.com/silverstripe/silverstripe-asset-admin/pull/1607

Doing in a PR because it's a touch too complex to do in a merge-up


client/src/containers/TableView/TableView.js
```diff
@@ -213,16 +213,22 @@ function TableView(_props) {
...
     const styles = {};
-    classNames.push(`${baseClass}--${category}`);
-    if (category === 'image' && url) {
-      styles.backgroundImage = `url("${url}")`;
+    let errorMsg = null;
+    if (rowData.message?.type === 'error') {
+      errorMsg = rowData.message.value;
     }
     // If the url is falsey then show error on the thumbnail. The exception is
     // folder since it doesn't have to physically exist on the file system
```
ChloeSartorelli: This comment is removed in the PR this is meant to be a copy of.

client/src/containers/TableView/TableView.js
```diff
@@ -213,16 +213,22 @@ function TableView(_props) {
...
-    if (!uploading && !url && category !== 'folder') {
+    const noUrlError = (!uploading && !url && category !== 'folder');
+    if (errorMsg || noUrlError) {
```
ChloeSartorelli: What's this new `noUrlError`? This wasn't in the original PR, why do we need this here?

ChloeSartorelli (changes requested):

client/src/containers/TableView/TableView.js
emteknetnz: Removed

client/src/containers/TableView/TableView.js
emteknetnz: Removed - looks like I got a bit confused when I originally did the PR

ChloeSartorelli (approved):
LGTM, matches commit that needs to be merged up (not sure where the conflict was)

Issue #1611, opened 2025-09-29
Accessibility: Cannot tab to or interact with files/headers in table view


ChloeSartorelli:
In the table view in asset admin (which displays files in a gridfield-like display), it's not possible to navigate to anything inside the actual table using keyboard. The tab order skips the table completely, going from the "change view" button directly to pagination.

This means we cannot do the following:
1. Select any file
2. Edit any file
3. Delete any file
4. Sort the table

I took a crack at making the checkbox and headers interactable, but Steve and I couldn't agree on what the correct behaviour should be. See https://github.com/silverstripe/silverstripe-asset-admin/pull/1614 for the discussion.

We need to discover what the correct keyboard interaction and screen reader behaviour should be for table view - which likely also means looking at what the behaviour should be in the gallery view so that the behaviour can be similar wherever that's sensible.

Note that the behaviour doesn't have to be identical between these two views - the fact that the information is being displayed differently and has different semantic HTML also makes it okay to change the interaction to suit the different display/semantics.

## Some options

1. [The listbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)? May not be suitable for tabular data
2. [The grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/)? Requires that keyboard focus can move to every single cell which may not be suitable
3. Make interaction the same as gridfields (i.e. add a `...` button to each row which when activated provides various actions that can be performed on that row
4. ??? Open to suggestions

## old PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1614
- https://github.com/silverstripe/developer-docs/pull/825

emteknetnz:
Agreed offline with UX team that general approach is use tab-islands to quickly tab between components, then use arrows to navigate within the component. This would imply that only one of the files should be tabbable, and then the arrows keys are used to navigate between files.

The single file which is tabbable should probably dynamically change, it should be either:
- the first file in the table, or
- the file currently being edited, or
- the last file was selected using the arrows keys before tabbling out of the component

PR #1614, opened 2025-10-05
ENH Make assets table view keyboard accessible

ChloeSartorelli:
- Make sortable headers accessible with keyboard
- Make select checkboxes accessible with keyboard

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1611


client/src/containers/Gallery/Gallery.js
```diff
@@ -451,6 +451,10 @@ class Gallery extends Component {
    * @param {Event} event
    */
   toggleSelectConcat(event) {
+    // Tile view doesn't use the click-to-drag selection functionality
+    if (this.props.view !== 'tile') {
+      return;
+    }
```
ChloeSartorelli, line 457: Needed to add this to prevent additional re-rendering of the table view every time you press shift

client/src/containers/TableView/TableView.js
```diff
@@ -27,22 +27,27 @@ function TableView(_props) {
...
+    // if this column is for selecting, let the label/checkbox control it instead.
     if (cell.column.id === 'selected') {
-      evt.stopPropagation();
-      evt.preventDefault();
-      if (typeof props.onSelect === 'function') {
-        props.onSelect(evt, rowData);
-        return;
-      }
+      return;
     }
```
ChloeSartorelli, line 33: Event handling moved into the `input` element's `onChange` event instead. See new `handleSelect` function

client/src/containers/TableView/TableView.js
```diff
@@ -65,10 +70,14 @@ function TableView(_props) {
   }
 
   /**
-   * Avoids the browser's default focus state when selecting an item.
+   * Get the sort order of the current column (ascending, descending, or null)
    */
-  function preventFocus(evt) {
-    evt.preventDefault();
```
ChloeSartorelli, line 71: This was used to prevent focus of the `input` checkbox, which is exactly the opposite of accessible.

client/src/containers/TableView/TableView.js
```diff
@@ -169,14 +178,13 @@ function TableView(_props) {
...
+  function renderSelect(rowData, rowId) {
     if (props.selectableItems && (props.selectableFolders || rowData.type !== 'folder')) {
       const checkboxProps = {
         type: 'checkbox',
         title: i18n._t('AssetAdmin.SELECT'),
-        defaultChecked: rowData.selected,
-        tabIndex: -1,
-        onMouseDown: (evt) => preventFocus(evt),
+        checked: rowData.selected,
+        onChange: (evt) => handleSelect(rowData, evt),
```
ChloeSartorelli, line 187: - Change `defaultChecked` to `checked` because it represents the _current_ value, not the original value when the form first loads.
- Remove `tabIndex` because checkboxes should be accessible to keyboard.
- Remove `onMouseDown` because we explicitly don't want to change the default focus behaviour.
- Add `onChange` event handler so the checkbox handles its own selection instead of weird handler on the table row.

client/src/containers/TableView/TableView.js
```diff
@@ -265,10 +282,10 @@ function TableView(_props) {
     if (columnIsSortable(header.column.id)) {
       // The "font-icon-" class is intentional. It adds the necessary CSS for adding an icon.
       // The actual icon itself is added through explicit CSS depending on how the column's sorted.
-      return <>
+      return <button onClick={() => handleSort(header)} className="btn gallery__table-column__sort-button">
         <span className="gallery__table-column__label">{label}</span>
         <span className="gallery__table-column__sort-icon font-icon-" aria-hidden="true" />
-      </>;
+      </button>;
```
ChloeSartorelli, line 298: Add `button` so the sort headers are accessible to keyboard and screen readers. Also removed `pointer` cursor from table headers, which was making it look like even non-sortable headers were interactive when they weren't.

client/src/containers/TableView/TableView.js
```diff
@@ -321,15 +338,15 @@ function TableView(_props) {
     columnConfig.unshift({
       id: 'selected',
       header: '',
-      cell: (info) => renderSelect(info.getValue()),
+      cell: (info) => renderSelect(info.getValue(), info.row.id),
```
ChloeSartorelli, line 351: row id needed for `aria-labelledBy` attribute - this means that screen reader users are told which row they'd be selecting when they focus on a checkbox

client/src/containers/TableView/TableView.js
```diff
@@ -321,15 +338,15 @@ function TableView(_props) {
...
     });
   }
 
   const columnHelper = createColumnHelper();
-
+  const columns = useMemo(() => columnConfig.map(config => columnHelper.accessor(row => row, config)), [props.selectableItems]);
   const table = useReactTable({
     data: props.files,
-    columns: columnConfig.map(config => columnHelper.accessor(row => row, config)),
+    columns,
```
ChloeSartorelli, line 362: Memoising columns means the table doesn't get rebuilt unnecessarily. This was causing problems when selecting a checkbox with the keyboard - it would rebuild the whole table and that would cause the focus to be lost.

client/src/containers/TableView/TableView.js
```diff
@@ -347,8 +364,8 @@ function TableView(_props) {
             {headerGroup.headers.map(header => (
               <th
                 key={header.id}
-                onClick={() => handleSort(header)}
```
ChloeSartorelli, line 350: Click handler moved to the new button instead.

client/src/containers/TableView/TableView.js
```diff
@@ -347,8 +364,8 @@ function TableView(_props) {
...
                 className={getCellClassName(header, true)}
+                aria-sort={getSortOrder(header.column.id)}
```
ChloeSartorelli, line 381: `aria-sort` attribute is documented in:
- https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-sort

client/src/containers/TableView/TableView.js
```diff
@@ -360,6 +377,7 @@ function TableView(_props) {
         {table.getRowModel().rows.map(row => (
           <tr
             key={row.id}
+            id={`row-${row.id}`}
```
ChloeSartorelli, line 393: id added so it can be referenced in `aria-labelledby` attribute

client/src/containers/TableView/TableView.scss
ChloeSartorelli, line 1: CSS changes ultimately result in a nearly identical display before and after the relevant accessibility changes.

There is a slight difference in the size of the sort icons in the sortable header but fiddling with that looked like more effort than it's worth. IMO the bigger icons are better anyway.

tests/behat/features/accessibility-asset-admin.feature
```diff
@@ -0,0 +1,125 @@
+@assets @retry @job3
```
emteknetnz: As mentioned https://github.com/silverstripe/silverstripe-asset-admin/pull/1616#discussion_r2412116770 - rather than creating a new file naming structure, follow the existing pattern and just call this file `accessibility-asset-admin.feature`

tests/behat/src/FixtureContext.php
```diff
@@ -283,10 +301,10 @@ protected function getGalleryItem($name, $timeout = 3)
         if ($cell) {
             return $cell;
         }
-        // Find by row
+        // Find by row (tr which has a child td with a descendent span with the text - the tr also has a specific class)
         $row = $page->find(
             'xpath',
-            "//tr[contains(@class, 'gallery__table-row')]//div//span[contains(text(), '{$name}')]"
+            "//tr[td//span[contains(text(),'{$name}')] and contains(@class, 'gallery__table-row')]"
```
ChloeSartorelli, line 307: This used to return the `span` element, but we need the `tr` element so we can find child elements (specifically the checkbox)

client/src/containers/TableView/TableView.js
emteknetnz: Please add this info as a code comment

emteknetnz (changes requested):
I tab over each item twice, with a slightly different focus each time. In one of these states I can press space to select the checkboxes used to select multiple files, or press enter to edit the file, but in the the other one I cannot do either. There should only be one focus state.

Arrow keys (up, down, left, right) do nothing, I would expect to be able to use these. This is possibly too hard techncially due to the way that files will dynamically wrap based on screen width, though if it's doable it would be a really nice to have. I'd much rather use arrows than tab to navigate the files than tab

Also need to add changelog entry

ChloeSartorelli:
> I tab over each item twice, with a slightly different focus each time. In one of these states I can press space to select the checkboxes used to select multiple files, or press enter to edit the file, but in the the other one I cannot do either. There should only be one focus state.
>
> Arrow keys (up, down, left, right) do nothing, I would expect to be able to use these. This is possibly too hard techncially due to the way that files will dynamically wrap based on screen width, though if it's doable it would be a really nice to have. I'd much rather use arrows than tab to navigate the files than tab

Just to confirm.... were you in table view? Not grid/gallery view? There are some separate cards for the accessibility issues in grid view, this one is _only_ for table view.

My experience is that in table view with this PR each checkbox can be focused, and nothing else in the table row can gain focus. Pressing space checks the checkbox.  Pressing enter doesn't do anything. This is the normal browser behaviour for keyboard interaction with a checkbox.

The up and down arrow keys don't allow you to move focus up and down rows - if you can find some accessibility recommendations from a reliable/reputable source that indicates using arrow keys should move focus up and down rows in a table then I'll look into adding that functionality.

ChloeSartorelli:
> Also need to add changelog entry

Done.

tests/behat/features/accessibility-asset-admin.feature
ChloeSartorelli: Done though I dislike that pattern and it's gonna cause a merge conflict.

client/src/containers/TableView/TableView.js
ChloeSartorelli: Done

emteknetnz:
Opps, yeah I was on grid view, not table view :1st_place_medal: 

OK, different issue in table view, as you cannot tab to the thumbnail (and press enter on it) there is no way to edit an image. With a mouse you do this by clicking on the image

I don't think arrow keys nav is strictly required according to [wcag 2.4.3 focus order](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html) - and tab is probably sufficient.

They did give and example of `An example of keyboard navigation that is not the sequential navigation addressed by this success criterion is using arrow key navigation to traverse a tree component.` which is similar to what did on the [sitetree component](https://github.com/silverstripe/silverstripe-cms/issues/3101).  For the file table view it's more of a straight grid so is more sequential.  The file grid view though it's a bit more debatable as there the intuitive control seems like it should it would be arrow up/down/left/right. If we wanted to add arrows for grid view, then we'd also want it for table view.  Happy to split the arrow nav off as a new card.

ChloeSartorelli:
> OK, different issue in table view, as you cannot tab to the thumbnail (and press enter on it) there is no way to edit an image.

You can edit the image by selecting it, then navigating to the actions panel and activating the "edit" button. I'm not convinced tabbing to the thumbnail specifically is a good way to get to the edit functionality. It feels rather arbitrary and I'm not sure what the screen reader experience should be. Given there _is_ a way to edit the file, I'm inclined to not make anything else in the row interactive.

> With a mouse you do this by clicking on the image

With a mouse clicking _anywhere_ on the row will open the edit form but that's not really feasible with keyboard interaction - you can't make the whole row a button.

> They did give and example of `An example of keyboard navigation that is not the sequential navigation addressed by this success criterion is using arrow key navigation to traverse a tree component.` which is similar to what did on the https://github.com/silverstripe/silverstripe-cms/issues/3101. For the file table view it's more of a straight grid so is more sequential. The file grid view though it's a bit more debatable as there the intuitive control seems like it should it would be arrow up/down/left/right. If we wanted to add arrows for grid view, then we'd also want it for table view. Happy to split the arrow nav off as a new card.

Sounds like that should be in both or neither, so needs to be a separate card. Note that a table isn't a tree component, and I think the intention there was to say "things that aren't regular HTML structures may need custom keyboard interaction design" rather than "do whatever makes sense to you even if it's not standard"

ChloeSartorelli:
I think if we want a faster way to edit (or perform other actions on) files for keyboard users we may need some UX advice for what that is - in that case IMO this PR should still be merged as is (it's already a vast improvement) and I can open a new card for further improvements.

emteknetnz:
> You can edit the image by selecting it, then navigating to the actions panel and activating the "edit" button. I'm not convinced tabbing to the thumbnail specifically is a good way to get to the edit functionality.

I just tested this, in a subfolder with a single file in it, when I select the file, I need to shift-tab 8 times to get to the edit button. That's a lot. And if have a more regular scenario where I want to edit the middle most file in a list of 30, well that's over 20 shift-tabs just get to the edit button. That seems like super bad usability.

What this should be doing is when you tab, the focus was on the entire row, pressing space checked the checkbox, and pressing enter edited the file. That way it would be consistent with the file grid view which has the space to check + enter to edit keyboard nav (ignore the buggy double focus state that does nothing, on the correct focus state both space + enter work as expected)

ChloeSartorelli:
I'm not convinced the grid view interaction is necessarily correct, but I guess we can go with that. What should the screen reader experience be? How do we inform users that space selects the row and enter opens the edit form?
I guess I can add a `caption` to the top of the table with that information so it only gets read out once and doesn't interrupt understanding a given row.

I guess then the `tr` gets `role="button"` and `tab-index=0`? I'll have to make sure the button role doesn't interfere with it also being a table row for screen reader purposes... otherwise I guess the button could go _inside_ the row but I dunno if that's semantically correct HTML. I'll have to play around a bit. I'll have to add some way to tell screen readers when it's selected vs not selected too since we're not interacting directly with the checkbox anymore.

This rambling is mostly me gathering my thoughts but if you have any concerns with the above approach lemme know.

Note that the tab position for the actions toolbar will change in one of the other issues. Note also that we could add keyboard shortcuts if we wanted for either specific actions or for focusing on the actions toolbar.

Note also that for anything _other than editing_ (or maybe publishing? Is that available in the edit form?) a file you'll still have to get to the actions toolbar to do anything.

And even for editing a file, you'll still need to navigate to the edit form - unless you want enter to move focus into the edit form? I don't think it does that in the grid view right now but I could be wrong.

emteknetnz:
> What should the screen reader experience be?

Up to you though perhaps split the aria/screen-reader portion to its own card (I would). I find it much easier to separate the two so there's less going on.

I also wouldn't worry about the shortcut action keys and tab order of the action bar at this stage, perhaps just make a new card or two for that stuff

ChloeSartorelli:
I'm not splitting screen reader interaction into its own card - keyboard navigation and screen reader announcement of how that navigation works go hand-in-hand, and changing one may require changing the other. I don't like splitting out things that are interlinked like that. They're not really separate at all.

> I also wouldn't worry about the shortcut action keys and tab order of the action bar at this stage, perhaps just make a new card or two for that stuff

I'm definitely not doing that as part of this card. I'm just pointing out that the problem you indicated has multiple possible solutions and you've just picked one that isn't necessarily in scope for this card at all.

IMO the scope of this card is "make it so you can select and perform actions for files" which this PR does. But I'll change the interaction to work in the way you want it to to avoid ping pong, unless that change turns out to be a massive piece of work (which is very well might).

ChloeSartorelli:
Rather than do something custom and weird (space and enter activating different actions) which may be confusing for screen reader users, I'm opting to go with [your first suggestion](https://github.com/silverstripe/silverstripe-asset-admin/pull/1614#issuecomment-3383238903) of having a separate edit button in the row, so there are two interactive elements - a checkbox to select the row, and a button to open the edit form.

I'm still not convinced this is correct and would still prefer to split out working out the correct behaviour separately from this card, but I'll give this a go and you can see what you think when it's done.

ChloeSartorelli:
I have also opened https://github.com/silverstripe/silverstripe-asset-admin/issues/1619 to separately address what the behaviour should be when opening an edit form in both gallery and table view.

emteknetnz:
I don't think this implementation is the way to go, it's now increased the doubled the number of times people need to press tab in order to traverse the list of files. It's also inconsistent with how it works in the gallery currently where enter edits the file and space selects it (ignoring all the currenty bugginess of that where there's a redundant focus state and pressing space seems to mess up focuus)

> I'm opting to go with https://github.com/silverstripe/silverstripe-asset-admin/pull/1614#issuecomment-3383238903 of having a separate edit button in the row

That's not what I suggested, I suggested having a single focus point and space/enter do different things

If you look at the gallery view you can see that the focus point encompasses the image, the title and the checkbox, the equivalent to this in the table view would probably be to focus the entire row.

> Rather than do something custom and weird (space and enter activating different actions) which may be confusing for screen reader user

As far as I can tell space/enter are the standard keys for what we're doing - seems like the most relevant patterns would be:
- https://www.w3.org/WAI/ARIA/apg/patterns/listbox/ - mentions using space to select from a multi-select
- https://www.w3.org/WAI/ARIA/apg/patterns/grid/ - mentions using enter to edit in the grid, also mentions using ctrl-space/shift-space to select all in column/row (though does not mention selecting a single item)

We'd want to have consistent controls for both gallery and list view. I'm pretty sure that we'd also only want to have a tab a single time to navigate between files (I'm assuming the most commons operation will be to select a file from existing files to link to a DataObject), so having to make double the number of tabs to navigate would be pretty awful. That kind of leaves having to have enter/space do different things, which is what we currently do on the gallery view so seems like we should also do that in table view?

ChloeSartorelli:
> [!NOTE]
> This is long - tl;dr in https://github.com/silverstripe/silverstripe-asset-admin/pull/1614#issuecomment-3387728631

> I don't think this implementation is the way to go

Nor do I.

I propose (and have been proposing for a few review cycles now) that _for this card_ we go with the minimal improvement that at least makes this thing usable (i.e. use the checkbox to select items, and the actions panel to perform actions).

We can open a new card to improve things from there. The new card would likely need to look at _both_ the gallery and table view to find a suitable solution that is unified (or at least similar) between both of these views and reduces effort to perform actions.

This would mean:

1. Keyboard users can at least use the table view in the meantime, whether we come to a conclusion of what the best case scenario should be in time for 6.2 or not
2. We have as much time as we need to figure out a best case scenario, which allows us to more easily bring in other teams (UX and testing team) to help design the behaviour

> It's also inconsistent with how it works in the gallery currently where enter edits the file and space selects it (ignoring all the currently bugginess of that where there's a redundant focus state and pressing space seems to mess up focuus)

I don't think we should rely on what's happening in the gallery view - there are several issues open for fixing accessibility issues in that view, which will likely change how that interaction works.

>> Rather than do something custom and weird (space and enter activating different actions) which may be confusing for screen reader user
> 
> As far as I can tell space/enter are the standard keys for what we're doing - seems like the most relevant patterns would be:
> 
> - https://www.w3.org/WAI/ARIA/apg/patterns/listbox/ - mentions using space to select from a multi-select
> - https://www.w3.org/WAI/ARIA/apg/patterns/grid/ - mentions using enter to edit in the grid, also mentions using ctrl-space/shift-space to select all in column/row (though does not mention selecting a single item)

Listbox doesn't seem like it's appropriate for this. Grid _might_ be... though this isn't really a grid which would more likely be 2d (you would navigate individual cells) - our data is linear (you navigate entire rows) so presumably is more https://www.w3.org/WAI/ARIA/apg/patterns/table/ which doesn't give examples or recommendations of mixed space/enter interaction.

Note also that the grid link you provided does mention enter and `ctrl+space/shift-space` but _not in combination at the same time_ . The implication to me is that you can you enter in one cell, or you can use ctrl+space in one cell - but not both in the same cell. Though I may have misunderstood the recommendation.
It also says enter should "Disables grid navigation and: If the cell contains editable content, places focus in an input field ... If the cell contains one or more widgets, places focus on the first widget." which isn't what we want, because this would require having a button in every cell, and pressing enter twice to activate the button.

Specifically this quote tells me the grid pattern probably isn't appropriate here:

> In a grid, every cell contains a focusable element or is itself focusable, regardless of whether the cell content is editable or interactive. There is one exception: if column or row header cells do not provide functions, such as sort or filter, they do not need to be focusable. One reason it is important for all cells to be able to receive or contain keyboard focus is that screen readers will typically be in their application reading mode, rather than their document reading mode, when users are interacting with the grid.

Every cell is focusable with the only exception being header cells. I don't think we want that.

> We'd want to have consistent controls for both gallery and list view. I'm pretty sure that we'd also only want to have a tab a single time to navigate between files (I'm assuming the most commons operation will be to select a file from existing files to link to a DataObject), so having to make double the number of tabs to navigate would be pretty awful. 

Sounds like the best course of action for now is for me to revert this latest change so the select checkbox is the only interactive element, and we can improve it further later if/when we know more about how interactive elements in a table should behave for keyboard interaction for accessibility.

> That kind of leaves having to have enter/space do different things, which is what we currently do on the gallery view so seems like we should also do that in table view?

I don't think enter and space should do different things on the same interactive element. And again I don't think we should rely on how the gallery view works as a best practice to model this from.

IMO we need to know more about how table data and interactive cells inside a table should behave for accessibility purposes rather than just guessing ourselves. Therefore if we do the simple thing now, we can do the complex thing when we know more about how the complex thing should work. Otherwise we risk spending a lot of time on something that isn't best practice.

ChloeSartorelli:
tl;dr:

I propose I revert this latest change, and for this card we allow tabbing to the checkbox and then using the actions panel as a minimal way to interact with rows. This is better than the current state where you can't do anything with the keyboard at all.

I can then open a new card that discusses what the best case scenario is for both grid view and table view. We can get help from the UX team and testing team to discover what the best practices are for these scenarios. We can then implement based on established best practices as discovered by the cross-discipline group of UX, testers, and us - instead of you and I arguing about things that IMO neither of us knows enough about to reach a consensus on.

ChloeSartorelli:
One option we may have (which again would be discussed in a separate issue) would be to make the file table view more like a gridfield - there could be a separate actions `...` button in the final column which includes edit/delete/publish actions in a dropdown.

A benefit of this approach is that the experience would be familiar across different tables in the CMS - i.e. the assets table and gridfield tables would have similar interaction instead of needing to know two different ways to navigate interactive tables.

emteknetnz:
I'd say for now the extra tabbing between checkbox and image is much less bad having to use a checkbox, and then shift-tab back to the action bar just to select 'edit'

Though if you see this PR as a halfway state, I'd still prefer the shift/enter solution and focus goes on the whole row, that way it's consistent with how gallery view operates, so if we never get to "redoing the file manager" (kind of sounds more likely than not), then at least we have a consistent user experience between the gallery and table view

ChloeSartorelli:
enter/space doing different things to the same interactive element doesn't seem to be best practice anywhere I can see... I don't want to do that. If I make the whole row focusable it needs an aria role, which would probably be `button`. Pressing `space` and `enter` on a button should both do the same thing.

If it's between not doing anything until we figure out the best case scenario and temporarily making space/enter do the same thing, I'm leaning towards do nothing tbh. A non-standard experience that IMO is bad practice is not something I want to put my name on.

ChloeSartorelli:
I'll close this PR and rework the issue to be about what the best case scenario should be, and flag it to Jenn for inter-team time to be booked.

PR #1623, opened 2025-10-21
ENH Hide remaining font-icons from screen readers

ChloeSartorelli:
Excludes icons in chosen-js elements, because this proved too difficult.

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1957


composer.json
```diff
@@ -6,7 +6,7 @@
     "require": {
         "php": "^8.3",
         "silverstripe/framework": "^6.1",
-        "silverstripe/admin": "^3"
+        "silverstripe/admin": "^3.2"
```
GuySartorelli: Need to bump since we're using a new component in JS

emteknetnz (approved):

Issue #1625, opened 2025-10-23
Upload field crashes when publishing a file not in the current folder


UndefinedOffset:
### Module version(s) affected

3.1.0

### Description

In asset admin 3.1.0 (at least) when you open an upload field that has a default folder other than the folder the file is actually in when you save (or publish) that file upload field's js throws an error and the modal and upload field disappears. This appears to be because the actual folder shown in the popup is the default folder not the folder the file is contained in.

This did work as expected in Silverstripe 5, untested but it's probably this bug also existed in Silverstripe 6.0.0 as well.

### How to reproduce

1. Have any data object (page or other) with an upload field with a default folder of say `example`
2. Upload a file through the assets section to say the `Uploads` folder
3. In the upload field select the file from the `Uploads` folder
4. Save the page or reload the window
5. Re-open the insert media modal via the eye button on the upload field
6. Click details for the file
7. Publish the file, the UI should disappear and so should the upload field

### Additional Context

JS Error thrown (full stack truncated as it's not very helpful):

```
TypeError: can't access property "filter", t is null
    render _resources/vendor/silverstripe/asset-admin/client/dist/js/bundle.js?m=1759443602:1
```

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1627
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1629

ChloeSartorelli:
> Have any data object (page or other) with an upload field with a default folder of say example

Can you please provide a quick code snippet that does this?

UndefinedOffset:
```php
class Page extends SiteTree
{
    private static $has_one = [
        'TestFile' => File::class,
    ];

    private static $table_name = 'Page';

    private static $owns = [
        'TestFile',
    ];

    /**
     * Gets fields used in the cms
     * @return \SilverStripe\Forms\FieldList Fields to be used
     */
    public function getCMSFields()
    {
        $fields = parent::getCMSFields();

        $fields->addFieldToTab(
            'Root.Main',
            UploadField::create('TestFile', 'Test File')
                ->setFolderName('test-folder')
        );

        return $fields;
    }
}
```

emteknetnz:
[...]

PR #1627, opened 2025-10-29
FIX Ensure query object has keys

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1625


ChloeSartorelli (changes requested):
The `AssetAdmin` component's `refetchFolder()` function has functionality which changes depending on whether `Object.keys(query).length` is 0 or not which heavily implies an empty query is valid.

This proposed change will make that condition always true. 

There's also logic in places like `renderGallery()` which use `this.props.query &&` i.e. allowing empty queries.

I think instead of adding default values to the query, we should just correctly account for empty/missing queries.

PR #1628, opened 2025-10-29
ENH Use javascript default parameters instead of defaultProps

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2037


ChloeSartorelli (approved):
LGTM

PR #1629, opened 2025-10-31
FIX Don't pass null query

ChloeSartorelli:
Looks like this `null` was introduced when removing graphql in https://github.com/silverstripe/silverstripe-asset-admin/commit/c34ad4eaa91a742b8ff1400b9b7389ee570d349a#diff-70b6d894c2ee55707c729504a8617ac1037791097583446f07eeeb76ccfe06a5 - passing an empty object instead works fine.

I played with making `AssetAdmin` explicitly allow `null` queries (looks like a bunch of the code is already set up for that with things like `const sort = this.props.query && this.props.query.sort;`) - but just ran into _even more problems_ when trying to do that.

At some stage it might be sensible to clean up `AssetAdmin` to explicitly _not_ allow null/undefined queries instead - though empty ones are fine.

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1625


client/lang/src/en.json
```diff
@@ -85,5 +85,9 @@
     "AssetAdmin.SIZE": "Size",
     "AssetAdmin.TITLE": "Title",
     "AssetAdmin.TYPE": "File type",
-    "AssetAdmin.URL": "URL"
-}
\ No newline at end of file
+    "AssetAdmin.URL": "URL",
+    "File.DRAFT": "Draft",
+    "File.MODIFIED": "Modified",
```
ChloeSartorelli, line 90: These are already in `UploadFieldItem.js` so adding them here will make them updatable in transifex next translations run

client/lang/src/en.json
```diff
@@ -85,5 +85,9 @@
...
+    "File.REMOVE": "Remove",
+    "File.VIEW": "View"
```
ChloeSartorelli, line 92: These two are new

client/src/components/UploadField/UploadFieldItem.js
```diff
@@ -311,6 +313,8 @@ class UploadFieldItem extends Component {
       <button
         className={classes}
         onClick={this.handleView}
+        aria-label={i18n._t('File.VIEW', 'View')}
+        title={i18n._t('File.VIEW', 'View')}
```
ChloeSartorelli, line 317: Adding so I have a way to reference the button in behat - and also good for accessibility. Adding the remove above while I'm at it for consistency and accessibility.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -358,7 +358,7 @@ class AssetAdmin extends Component {
         } else if ((action === 'action_save' || action === 'action_publish')
           && this.getFolderId() !== response.record.parent.id) {
           // If the file was moved, open the folder containing the file was moved to
-          this.handleBrowse(response.record.parent.id, response.record.id, null);
+          this.handleBrowse(response.record.parent.id, response.record.id, {});
```
ChloeSartorelli, line 361: This is the fix

emteknetnz (approved):
Works well

PR #1631, opened 2025-11-03
ENH Add aria-sort to table view headers

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11860


client/src/containers/TableView/TableView.js
```diff
@@ -257,6 +257,17 @@ function TableView(_props) {
...
 
+  /**
+   * Returns the aria-sort value for a th table header
+   */
+  function getAriaSortValue(columnId) {
+    const [sortColumn, sortDirection] = props.sort.split(',');
+    if (sortColumn === columnId) {
+      return sortDirection === 'asc' ? 'ascending' : 'descending';
+    }
+    return 'none';
```
ChloeSartorelli: We should not have "none" - see https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/
Any header which is not explicitly sorting the column should not have an `aria-sort` attribute at all.

ChloeSartorelli (changes requested):

client/src/containers/TableView/TableView.js
emteknetnz: Updated

ChloeSartorelli (approved):
LGTM

PR #1635, opened 2025-11-13
ENH Add UnsavedChangesIndicator to file edit form

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1973


client/src/containers/Editor/Editor.scss
ChloeSartorelli, line 1: Why do we need all this extra CSS? Especially the hide/display stuff, which is surely already in admin?

ChloeSartorelli (changes requested):

client/src/containers/Editor/Editor.scss
emteknetnz: Asset-admin has a very dynamic 'file details' panel which takes up the entire viewport width on smaller break points, though gets smaller on large breakpoints where it will only show in thin right hand panel while the rest of the gallery takes up most of the viewport width

client/src/containers/Editor/Editor.scss
```diff
@@ -183,6 +183,39 @@
   }
 }
 
+.editor .unsaved-changes-indicator {
```
ChloeSartorelli: Instead of (or as well as) `.editor` use a selector that's more specific to asset-admin if possible.

ChloeSartorelli (changes requested):

client/src/containers/Editor/Editor.scss
emteknetnz: Changed to `.editor__details`

ChloeSartorelli (approved):
LGTM, works well locally

Issue #1636, opened 2025-11-20
Convert Gallery to functional components


emteknetnz:
There are react class component(s) for Gallery that needs to be converted to react functional component(s).

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1637

ChloeSartorelli:
PR merged

PR #1637, opened 2025-11-20
ENH Refactor Gallery from class to functional component

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1636


client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -38,35 +38,121 @@ const preventFocus = (event) => {
...
-    this.handleActivate = this.handleActivate.bind(this);
-    this.handleKeyDown = this.handleKeyDown.bind(this);
-    this.handleCancelUpload = this.handleCancelUpload.bind(this);
-  }
+  /**
+   * Check if this item has been saved, either in this request or in a prior one
+   *
+   * @return {Boolean}
+   */
+  const saved = () => item.id > 0;
```
emteknetnz, line 85: Moved saved() to hasError() methods up because they're called by other methods

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -86,60 +172,120 @@ class GalleryItem extends Component {
...
-  getErrorMessage() {
-    let message = null;
-    const { item, loadState } = this.props;
-
-    if (this.hasError()) {
-      message = item.message.value;
-    } else if (this.missing()) {
-      message = i18n._t('AssetAdmin.FILE_MISSING', 'File cannot be found');
+  const getErrorMessage = () => {
+    let errMessage = null;
```
emteknetnz, line 200: Renamed var from message because it had same name as prop

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -86,60 +172,120 @@ class GalleryItem extends Component {
...
     }
-  }
+  };
+
+  /**
+   * Gets a function that may be overloaded at the item level
+   * @param {string} functionName
+   * @returns {Function}
+   */
+  const getItemFunction = (functionName) => {
```
emteknetnz, line 181: Moved method up because it's called by another method below

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -86,60 +172,120 @@ class GalleryItem extends Component {
...
-  }
+  };
+
+  /**
+   * Determine that this record is an image, and the thumbnail is smaller than the given
+   * thumbnail area
+   *
+   * @returns {boolean}
+   */
+  const isImageSmallerThanThumbnail = () => {
```
emteknetnz, line 230: Moved method up because it's called by another method below

client/src/containers/Gallery/Gallery.js
```diff
@@ -39,85 +39,68 @@ const ACTION_TYPES = {
...
-
-      $select.off('change');
-    }
-
-    // Flush uploaded files on folder navigation
-    if (prevProps.folderId !== this.props.folderId) {
-      this.props.actions.queuedFiles.purgeUploadQueue();
-    }
-  }
+const Gallery = ({
```
emteknetnz, line 42: Added the following as they were previously referred to via this.props
- LoadingComponent
- onClearSearch
- onInsertMany
- noticeMessage
- queuedFiles

Removed the following props because they were unused:
- combinedFiles
- count
- highlightedFiles
- search
- sorters

client/src/containers/Gallery/Gallery.js
```diff
@@ -179,7 +162,23 @@ class Gallery extends Component {
...
     );
-  }
+  };
+
+  /**
+   * Gets items that can be selected, in order
+   *
+   * @return {Array}
+   */
+  const getSelectableFiles = () => {
```
emteknetnz, line 172: Moved method up because it's called by another method below

client/src/containers/Gallery/Gallery.js
```diff
@@ -373,36 +356,43 @@ class Gallery extends Component {
...
+  };
 
-  handleUploadProgress(file, progress) {
-    this.props.actions.queuedFiles.updateQueuedFile(file._queuedId, { progress });
-  }
+  const handleUploadProgress = (file, progress) => {
+    actions.queuedFiles.updateQueuedFile(file._queuedId, { progress });
+  };
+
+  const handleFailedUpload = (fileXhr, response) => {
```
emteknetnz, line 367: Moved method up because it's called by another method below

client/src/containers/Gallery/Gallery.js
```diff
@@ -413,56 +403,45 @@ class Gallery extends Component {
...
-   * Toggle concatenating selected items based on the key event
+   * Determines whether concat should happen
    *
    * @param {Event} event
+   * @return {boolean}
    */
-  toggleSelectConcat(event) {
-    this.props.actions.gallery.setConcatenateSelect(this.isConcat(event));
-  }
+  const isConcat = (event) => event.metaKey || event.ctrlKey || event.shiftKey;
```
emteknetnz, line 435: Moved method up because it's called by another method below

client/src/containers/Gallery/Gallery.js
```diff
@@ -610,104 +563,115 @@ class Gallery extends Component {
    * @param {Event} event - Event object.
    * @param {Object} item - The item being selected/deselected
    */
-  handleSelect(event, item) {
-    const maxFiles = this.props.maxFilesSelect;
-    const selectable = this.getSelectableFiles();
+  const handleSelect = (event, item) => {
```
emteknetnz, line 566: Moved method up because it's called by another method below

client/src/containers/Gallery/Gallery.js
```diff
@@ -610,104 +563,115 @@ class Gallery extends Component {
...
   /**
    * Generates the react components needed for the BulkActions part of this
    * component.
    *
    * @returns {XML}
    */
-  renderBulkActions() {
-    const { type, dialog, maxFilesSelect, files, selectedFiles, BulkActionsComponent, sectionConfig } = this.props;
-
+  const renderBulkActions = () => {
```
emteknetnz, line 674: Moved method up because it's called by another method below

client/src/containers/Gallery/Gallery.js
```diff
@@ -313,58 +296,58 @@ class Gallery extends Component {
...
 
   /**
    * Handles setting the pagination page number
    *
    * @param {number} page
    */
-  handleSetPage(page) {
-    this.props.onSetPage(page);
-  }
+  const handleSetPage = (pageParam) => {
```
emteknetnz, line 316: Rename param because it's the same name as a prop

client/src/containers/Gallery/Gallery.js
```diff
@@ -567,41 +543,18 @@ class Gallery extends Component {
...
+  };
 
   /**
    * Handles a user drilling down into a folder.
    *
    * @param {Event} event - Event object.
    * @param {Object} folder - The folder that's being activated.
    */
-  handleOpenFolder(event, folder) {
+  const handleOpenFolder = (event, folderParam) => {
```
emteknetnz, line 554: Rename param because it's the same name as a prop

client/src/containers/Gallery/Gallery.js
```diff
@@ -610,104 +563,115 @@ class Gallery extends Component {
...
-  handleMoveFiles(folderId, fileIds) {
-    const url = this.props.sectionConfig.endpoints.move.url;
+    onOpenFile(file.id, file);
+  };
+
+  const handleEnableDropzone = (enabled) => {
+    actions.gallery.setEnableDropzone(enabled);
+  };
+
+  const handleMoveFiles = (folderIdParam, fileIds) => {
```
emteknetnz, line 631: Rename param (folderId => folerIdParam) because it's the same name as a prop

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -38,35 +38,121 @@ const preventFocus = (event) => {
   event.preventDefault();
 };
 
-class GalleryItem extends Component {
-  constructor(props) {
-    super(props);
+const GalleryItem = ({
```
emteknetnz: This following where added, they were previously referred to via this.props
- maxSelected
- children
- actions (added via redux)

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -553,6 +617,7 @@ GalleryItem.propTypes = {
   // Whether the item should be enlarged for more prominence than "highlighted"
   isDropping: PropTypes.bool,
   isDragging: PropTypes.bool,
+  maxSelected: PropTypes.bool,
```
emteknetnz, line 557: This should be a bool - see ThumnailView.js logic

client/src/components/GalleryToolbar/Buttons/UploadButton.js
```diff
@@ -8,25 +7,22 @@ import PropTypes from 'prop-types';
...
+  >
+    <span className="font-icon-upload btn__icon" aria-hidden="true" />
+    <span className="btn__text btn__title">{i18n._t('AssetAdmin.DROPZONE_UPLOAD')}</span>
+  </button>
+);
 
-UploadButton.defaultProps = {
-  canEdit: PropTypes.func.isRequired,
+UploadButton.propTypes = {
+  canEdit: PropTypes.bool.isRequired,
```
emteknetnz, line 25: This was wrong before - they code refers to it as a bool

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -38,35 +38,121 @@ const preventFocus = (event) => {
...
+    if (shouldLoadImage({
+      item,
+      sectionConfig,
+    })) {
+      actions.imageLoad.loadImage(
+        item.thumbnail,
+        sectionConfig.imageRetry
+      );
+    }
+  }, [item, sectionConfig, actions]);
```
ChloeSartorelli:
```suggestion
  }, [item, sectionConfig, actions.imageLoad]);
```
We don't need to fire this effect if unrelated actions change.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -38,35 +38,121 @@ const preventFocus = (event) => {
...
+   */
+  const canBatchSelect = () => selectable && item.canEdit;
+
+  /**
+   * Checks if the component has an error set.
+   *
+   * @return {boolean}
+   */
+  const hasError = () => {
+    let hasErrorFlag = false;
```
ChloeSartorelli, line 137: `hasErrorFlag` was renamed with no stated reason, but doesn't seem to affect anything so okay to leave as is.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -86,60 +172,120 @@ class GalleryItem extends Component {
...
+      return item[functionName];
+    }
+    // Return the default function based on the function name
+    const functionMap = {
+      updateStatusFlags,
+      updateProgressBar,
+      updateErrorMessage,
+    };
+    return functionMap[functionName] || (() => null);
+  };
```
ChloeSartorelli, line 192: Looks like this was completely refactored? There's no `functionMap` variable in the original implementation.
Was this just a way around the fact that we dereferenced `props` when the original hadn't?

What's the new `|| (() => null)` for? Why's that needed now but wasn't before?

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -86,60 +172,120 @@ class GalleryItem extends Component {
...
+        onSelect,
+        onCancelUpload,
+        onRemoveErroredUpload,
+        badge,
+        updateStatusFlags,
+        updateProgressBar,
+        updateErrorMessage,
+        children,
+        actions,
+      });
```
ChloeSartorelli: This is kinda ugly (especially the `message: {value: errMessage}` part), and happens multiple times in this file. Might be better to just accept `props` in the component function, and then dereference afterwards? Then we can still use the `props` variable for stuff like this. 

i.e.
```js
const GalleryItem = (props) => {
  const {
    // ... all of the dereferencing goes here
  } = props;
```

If left as is, I need to go through everywhere in the props chain (i.e. wherever these are passed through directly, and then anywhere they gets passed from there on) to double check the list is correct. It is also a wall of text which feels like it'll be mildly frustrating any time we need to touch it.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -86,60 +172,120 @@ class GalleryItem extends Component {
...
     } else if (loadState === IMAGE_STATUS.FAILED) {
-      message = i18n._t('AssetAdmin.FILE_LOAD_ERROR', 'Thumbnail not available');
+      errMessage = i18n._t('AssetAdmin.FILE_LOAD_ERROR', 'Thumbnail not available');
     }
 
-    if (message !== null) {
-      const updateErrorMessage = this.getItemFunction('updateErrorMessage');
-      message = updateErrorMessage(message, this.props);
+    if (errMessage !== null) {
+      const updateErrorMessageFn = getItemFunction('updateErrorMessage');
```
ChloeSartorelli, line 211: `updateErrorMessageFn` renamed for no stated reason. Doesn't seem to cause problems but please do state reasons for changes in these refactor PRs.

This is not the first PR like this where I've asked this.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -154,52 +300,38 @@ class GalleryItem extends Component {
...
    * Retrieves class names for the item
    *
    * @returns {string}
    */
-  getItemClassNames() {
-    const category = this.props.item.category || 'false';
-    const selected = this.props.selectable && (this.props.item.selected || this.props.isDragging);
+  const getItemClassNames = () => {
+    const category = item.category || 'false';
+    const isSelected = selectable && (item.selected || isDragging);
```
ChloeSartorelli, line 291: Not clear why this was renamed but seems to be fine.
Note that everything that's renamed adds additional review burden because I then have to check all usages to make sure it got renamed consistently.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -215,22 +347,42 @@ class GalleryItem extends Component {
...
+      onSelect,
+      onCancelUpload,
+      onRemoveErroredUpload,
+      badge,
+      updateStatusFlags,
+      updateProgressBar,
+      updateErrorMessage,
+      children,
+      actions,
+    });
```
ChloeSartorelli: Refer to https://github.com/silverstripe/silverstripe-asset-admin/pull/1637#discussion_r2554303146

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -257,289 +409,201 @@ class GalleryItem extends Component {
...
```
ChloeSartorelli: Refer to https://github.com/silverstripe/silverstripe-asset-admin/pull/1637#discussion_r2554303146

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -215,22 +347,42 @@ class GalleryItem extends Component {
         });
       }
     }
-    const updateStatusFlags = this.getItemFunction('updateStatusFlags');
-    flags = updateStatusFlags(flags, this.props);
+    const updateStatusFlagsFn = getItemFunction('updateStatusFlags');
```
ChloeSartorelli, line 329: `updateStatusFlags` renamed - seems fine, leave as is.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -257,289 +409,201 @@ class GalleryItem extends Component {
...
+    if (!hasError() && uploading() && !complete()) {
       progressBar = (
         <div className="gallery-item__upload-progress">
           <div {...progressBarProps} />
         </div>
       );
     }
-    const updateProgressBar = this.getItemFunction('updateProgressBar');
-    progressBar = updateProgressBar(progressBar, this.props);
+    const updateProgressBarFn = getItemFunction('updateProgressBar');
```
ChloeSartorelli, line 393: `updateProgressBar` renamed - seems fine, leave as is.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -257,289 +409,201 @@ class GalleryItem extends Component {
...
-      hasError = this.props.item.message.type === 'error';
-    }
-
-    return hasError;
-  }
+  };
 
   /**
-   * Wrapper around this.props.onActivate
+   * Wrapper around onActivate
```
ChloeSartorelli:
```suggestion
   * Wrapper around onActivate prop
```
The context that this is wrapping a prop seems important to keep

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -257,289 +409,201 @@ class GalleryItem extends Component {
...
-      this.props.onActivate(event, this.props.item);
+    if (typeof onActivate === 'function' && saved()) {
+      onActivate(event, item);
     }
-  }
+  };
 
   /**
-   * Wrapper around this.props.onSelect
+   * Wrapper around onSelect
```
ChloeSartorelli:
```suggestion
   * Wrapper around onSelect prop
```
The context that this is wrapping a prop seems important to keep

client/src/components/GalleryItem/GalleryItem.js
ChloeSartorelli: A link would be handy next time but not strictly necessary.

client/src/components/GalleryToolbar/Buttons/UploadButton.js
```diff
@@ -8,25 +7,22 @@ import PropTypes from 'prop-types';
...
-        disabled={!canEdit}
-      >
-        <span className="font-icon-upload btn__icon" aria-hidden="true" />
-        <span className="btn__text btn__title">{i18n._t('AssetAdmin.DROPZONE_UPLOAD')}</span>
-      </button>
-    );
-  }
-}
+const UploadButton = ({
+  canEdit = true,
```
ChloeSartorelli: Where is the default value coming from?

client/src/components/GalleryToolbar/GalleryToolbar.js
```diff
@@ -68,90 +76,76 @@ class GalleryToolbar extends Component {
...
    *
    * @returns {Array} buttons
    */
-  renderViewChangeButtons() {
+  const renderViewChangeButtons = () => {
     const views = ['tile', 'table'];
-    return views.map((view) => {
-      const icon = (view === 'table') ? 'list' : 'thumbnails';
+    return views.map((viewButton) => {
+      const icon = (viewButton === 'table') ? 'list' : 'thumbnails';
```
ChloeSartorelli, line 89: Renamed, presumably to avoid conflict with the prop

client/src/containers/Gallery/Gallery.js
```diff
@@ -39,85 +39,68 @@ const ACTION_TYPES = {
...
+  actions,
+  selectedFiles,
+  files,
+  folderId,
+  fileId,
+  folder,
+  queuedFiles,
+  sort,
+  page = 1,
+  limit = 15,
```
ChloeSartorelli, line 52: These didn't have defaults in `Gallery.defaultProps`

client/src/containers/Gallery/Gallery.js
```diff
@@ -870,133 +819,160 @@ class Gallery extends Component {
...
+      window.removeEventListener('keyup', toggleSelectConcat);
+    };
+  }, []);
+
+  useEffect(() => {
+    initSortDropdown();
+    if (prevFolderIdRef.current !== folderId) {
+      initFlushUploadFiles({ folderId: prevFolderIdRef.current });
+      prevFolderIdRef.current = folderId;
+    }
```
ChloeSartorelli, line 855: What's this all about? There was nothing like this conditional block in the `componentDidUpdate()` implementation.

If this is a way to handle `prevprops.folderId` in `initFlushUploadFiles()` - presumably this new logic should live in that function instead? Looks like if `view` changes on its own, this new logic might prevent the first part of that function from working as expected.

ChloeSartorelli: We can remove the need for the weird `prevFolderIdRef` altogether by splitting `initFlushUploadFiles` up into two separate effects:

```js
useEffect(() => {
  initSortDropdown();
  // turn off chosen.js
  if (view !== 'tile') {
    const $select = getSortElement();
    $select.off('change');
  }
}, [view]);

useEffect(() => {
  actions.queuedFiles.purgeUploadQueue();
}, [folderId]);
```

Note that the above assumes `initSortDropdown` is only needed when `view` changes.

client/src/containers/Gallery/Gallery.js
```diff
@@ -39,85 +39,68 @@ const ACTION_TYPES = {
...
 
   /**
    * Compose the search critia into a human readable message
    *
    * @param {object} search
    * @returns {string}
    */
-  getSearchMessage(filters) {
+  // eslint-disable-next-line no-unused-vars
+  const getSearchMessage = (filters) => {
```
ChloeSartorelli, line 103: the disable line is for `getSearchMessage()` I assume? Should we just remove that if it's not used?

client/src/containers/Gallery/Gallery.js
```diff
@@ -471,32 +450,29 @@ class Gallery extends Component {
...
-  }
+  const hasOpenedItem = () => !!fileId;
 
-  handleClearSearch(event) {
-    this.props.onClearSearch(event);
-  }
+  // eslint-disable-next-line no-unused-vars
+  const handleClearSearch = (event) => {
+    onClearSearch(event);
+  };
```
ChloeSartorelli, line 465: I assume the disable line is for `handleClearSearch`? Should we remove it if it's not used?

client/src/containers/Gallery/Gallery.js
```diff
@@ -718,7 +682,7 @@ class Gallery extends Component {
       ? action => action.value !== ACTION_TYPES.DELETE
       : action => action.value !== ACTION_TYPES.ARCHIVE;
 
-    const actions = CONSTANTS.BULK_ACTIONS
+    const bulkActionsList = CONSTANTS.BULK_ACTIONS
```
ChloeSartorelli, line 685: Renamed - presumably to avoid collision with a prop

client/src/containers/Gallery/Gallery.js
```diff
@@ -764,49 +728,44 @@ class Gallery extends Component {
...
    * @returns {XML}
    */
-  renderGalleryView() {
-    const GalleryView = (this.props.view === 'table') ? TableView : ThumbnailView;
-    const files = this.props.files.map((file) => {
-      const selected = this.itemIsSelected(file.id);
-      const highlighted = this.itemIsHighlighted(file.id);
+  const renderGalleryView = () => {
+    const GalleryView = (view === 'table') ? TableView : ThumbnailView;
+    const mappedFiles = files.map((file) => {
```
ChloeSartorelli, line 761: Renamed - presumably to avoid collision with a prop

client/src/containers/Gallery/Gallery.js
```diff
@@ -870,133 +819,160 @@ class Gallery extends Component {
...
+        </div>
+      );
+    }
+    if (loading) {
       return (
         <div className="flexbox-area-grow">
-          <div className="editor__file-preview-message--file-missing m-t-3">
-            {i18n._t('Admin.UNKNOWN_ERROR', 'An unknown error has occurred')}
-          </div>
+          <LoadingComponent />
```
ChloeSartorelli, line 874: Used to be `Loading` but is now `LoadingComponent`?

Edit: Had to chase this for a minute - this is the prop. Used to be `const Loading = this.props.LoadingComponent` so this change is correct.

ChloeSartorelli (changes requested):

client/src/components/GalleryItem/GalleryItem.js
emteknetnz: Updated

client/src/components/GalleryItem/GalleryItem.js
emteknetnz: We cannot go `this.props[functionName]` as there is no `this.props` as we're using deconstructed parameters now. This is roughtly equivalent.

There were 3x calls to this method before:
- `this.getItemFunction('updateStatusFlags')`
- `this.getItemFunction('updateProgressBar')`
- `this.getItemFunction('updateErrorMessage')`

The `|| (() => null)` is to satisfy the `@returns {Function}` in case a `functionName` not equally to one of the 3x above was returned

I think this code is correct, though if you want I'll remove the `|| (() => null)`

client/src/components/GalleryItem/GalleryItem.js
emteknetnz: Renamed var because it clashes with the `updateErrorMessage()` method

client/src/components/GalleryItem/GalleryItem.js
emteknetnz: I'll changed to something of a middle ground. I don't want to simply have (props) in the constructor as that's inconsistent with the the why I'm doing the constructors everywhere else.

Instead what I've done is created getPropsForFunction() which essentially just returns this.props, though with some code duplication

There's also some risk that we we add a new prop to the component and for some reason it breaks something because it's not being passed through, though if that happens during some work it's probably quite easy to debug, so I think this is fine

client/src/components/GalleryItem/GalleryItem.js
emteknetnz: Renamed because select is a prop

client/src/components/GalleryItem/GalleryItem.js
emteknetnz: Updated

client/src/components/GalleryItem/GalleryItem.js
emteknetnz: Updated

client/src/components/GalleryToolbar/Buttons/UploadButton.js
emteknetnz: Removed

client/src/components/GalleryToolbar/GalleryToolbar.js
emteknetnz: Correct

client/src/containers/Gallery/Gallery.js
emteknetnz: They were in

```
const sharedDefaultProps = {
  page: 1,
  limit: 15,
};
```

Which got merged in to `Gallery.defaultProps`

client/src/containers/Gallery/Gallery.js
emteknetnz: This method was not called, I did not want to simply remove it though as it would make a big difference in the diff.

I'd prefer to keep to minimise scope of changes, though let me know if you would like me to just remove the method instead

client/src/containers/Gallery/Gallery.js
emteknetnz: This method was not called, I did not want to simply remove it though as it would make a big difference in the diff.

I'd prefer to keep to minimise scope of changes, though let me know if you would like me to just remove the method instead

client/src/containers/Gallery/Gallery.js
emteknetnz: Correct

client/src/containers/Gallery/Gallery.js
emteknetnz: Correct

client/src/containers/Gallery/Gallery.js
emteknetnz: I tried the refactor as recommended, though a unit test that was present in the class component `Gallery componentDidUpdate() should not call purgeUploadQueue when receiving same files` will fail

The reason is the that new `useEffect()` with the `[folderId]` dependency will still be called on mount.

I've tidied up the useEffect() added in this PR as well as initFlushUploadFiles() and put a note in there to explain why we're using prevFolderIdRef.current to track the previous value of folderId

client/src/components/GalleryItem/GalleryItem.js
ChloeSartorelli: This code is fine, thank you for explaining. I just needed to know what the reasoning was for the changes - sounds like they're for good reasons.

client/src/containers/Gallery/Gallery.js
ChloeSartorelli: It's fine to keep, but also fine to remove. A big chunk removed with a "this was removed because it's not used anywhere" comment is very little review burden (even less than functions that are moved with a stated reason) so I'm totally fine with you removing these if you want to going forward.

It's mostly when there's no stated reason for a change that it becomes a burden.

client/src/containers/Gallery/Gallery.js
ChloeSartorelli: as per https://github.com/silverstripe/silverstripe-asset-admin/pull/1637#discussion_r2554501093

client/src/components/GalleryItem/GalleryItem.js
ChloeSartorelli: This still requires me to now go find the three functions that are currently passed in as props (from anywhere they might be passed from - hopefully there's only one for each function) and make sure all the necessary props are passed in to each of those. And if they then pass the props on somewhere else..... that's a lot of review work that just moving the dereferencing down a line cuts out entirely.

emteknetnz: Switched to using props in constructor and then immediately dereferencing. Have removed `getPropsForFunction()`

client/src/containers/Gallery/Gallery.js
emteknetnz: Will keep for now

client/src/containers/Gallery/Gallery.js
emteknetnz: Keeping

ChloeSartorelli (approved):
LGTM

Issue #1638, opened 2025-11-24
Convert some asset-admin class components to functional components


emteknetnz:
There are react class component(s) that needs to be converted to react functional component(s).

Convert the following components:
- BackButton.js
- BulkActions.js
- InsertEmbedModal.js
- ~InsertMediaModal.js~ was unable to get this converted without behat failure (unexpected alert)
- MoveModal.js
- PreviewImageField.js

Note there are plenty of other class components within asset-admin that need to be converted, only doing six of them here to limit the scope of peer review required at once

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1639

ChloeSartorelli:
PR merged

PR #1639, opened 2025-11-24
ENH Refactor components from class to functional

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1638


client/src/components/InsertEmbedModal/InsertEmbedModal.js
```diff
@@ -9,36 +9,27 @@ import { joinUrlPaths } from 'lib/urls';
...
+  fileAttributes = {},
+  onClosed,
+  className = '',
+  actions,
+  schemaUrl,
+  targetUrl,
+  onLoadingError,
+  FormBuilderModalComponent = FormBuilderModal
+}) => {
+  const clearOverrides = () => {
```
emteknetnz, line 30: Moved method up because it's called by another method below

client/src/containers/InsertMediaModal/InsertMediaModal.js
```diff
@@ -13,72 +13,89 @@ import PropTypes from 'prop-types';
...
-  getModalProps() {
-    const { onHide, onInsert, sectionConfig, schemaUrl, className, ...props } = this.props;
-    return {
-      ...props,
-      className: classnames('insert-media-modal', className),
-      size: 'lg',
-      showCloseButton: false
-    };
-  }
+  const renderToolbarChildren = () => (
```
emteknetnz: Moved method up because it's called by something below

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -13,168 +13,130 @@ import PropTypes from 'prop-types';
...
-    this.handleSending = this.handleSending.bind(this);
-    this.handleUploadProgress = this.handleUploadProgress.bind(this);
-    this.handleUploadComplete = this.handleUploadComplete.bind(this);
-    this.handleCancelUpload = this.handleCancelUpload.bind(this);
-    this.handleRemoveErroredUpload = this.handleRemoveErroredUpload.bind(this);
-    this.canFileUpload = this.canFileUpload.bind(this);
-    this.updateFormData = this.updateFormData.bind(this);
-    this.cacheBustUrl = this.cacheBustUrl.bind(this);
-  }
+const PreviewImageField = ({
```
emteknetnz, line 16: Have removed the prop formid because it was unused

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -13,168 +13,130 @@ import PropTypes from 'prop-types';
...
 
   /**
    * Handles when a file is added to this field.
    *
    * @param {object} data
    */
-  handleAddedFile(data) {
-    this.props.actions.previewField.addFile(this.props.id, data);
-  }
+  const handleAddedFile = (dataParam) => {
```
emteknetnz, line 115: Renamed param because it was same as a prop

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -187,89 +149,80 @@ class PreviewImageField extends Component {
...
   /**
    * Build the preview URL
    * @param {string} category
    * @param {object} upload
-   * @param {object} data
+   * @param {object} dataParam
    * @returns {string}
    */
-  preview(category, upload, data) {
+  const preview = (category, uploadParam, dataParam) => {
```
emteknetnz, line 220: Renamed params because they were the same as props

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -187,89 +149,80 @@ class PreviewImageField extends Component {
       'Are you sure you want upload a file with a different extension?'
     );
 
-    return this.props.confirm(message);
-  }
-
-  preventDefault(e) {
```
emteknetnz, line 193: Deleted because unused

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -187,89 +149,80 @@ class PreviewImageField extends Component {
...
+    actions.previewField.updateStatus(id, { status });
+  };
+
+  /**
+   * Append a vid parameter to the URL to bust the cache
+   * @param {string} url
+   * @param {string} versionId
+   * @return string
+   */
+  const cacheBustUrl = (url, versionId = '') => {
```
emteknetnz, line 202: Moved method up because it's called by something below

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -345,32 +298,70 @@ class PreviewImageField extends Component {
...
+  const getDropzoneProps = () => {
+    const endpoint = data.uploadFileEndpoint;
+    const options = {
+      url: endpoint && endpoint.url,
+      method: endpoint && endpoint.method,
+      paramName: 'Upload',
+      clickable: true,
+      maxFiles: 1,
+    };
+    const previewObj = {
```
emteknetnz, line 324: Renamed var because preview is the name of a method

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -345,32 +298,70 @@ class PreviewImageField extends Component {
         {messageBox}
       </div>
     );
-  }
+  };
 
-  render() {
-    const { AssetDropzoneComponent } = this.props;
-    const dropzoneProps = this.getDropzoneProps();
+  const getDropzoneProps = () => {
```
emteknetnz, line 315: Moved method down because there are heaps of methods that it refers to that are now above it

client/src/components/InsertEmbedModal/InsertEmbedModal.js
```diff
@@ -85,60 +124,24 @@ class InsertEmbedModal extends Component {
...
-      case 'action_cancel': {
-        this.props.onClosed();
-        break;
-      }
-      default: {
-        // noop
-      }
+  useEffect(() => {
+    if (isOpen) {
+      setOverrides({ schemaUrl, fileAttributes });
```
emteknetnz: schemaUrl and fileAttributes were the only two props that were used in setOverrides()

the logic in setOverrides() is pretty goofy as we're passing in props as a param and we could just remove the param, but I want to keep the code as similar as possible without refactoring

client/src/components/InsertEmbedModal/InsertEmbedModal.js
```diff
@@ -9,39 +9,31 @@ import { joinUrlPaths } from 'lib/urls';
 
 const sectionConfigKey = 'SilverStripe\\AssetAdmin\\Controller\\AssetAdmin';
 
-class InsertEmbedModal extends Component {
-  constructor(props) {
-    super(props);
-
-    this.handleSubmit = this.handleSubmit.bind(this);
-  }
+const InsertEmbedModal = (props) => {
```
emteknetnz, line 12: props is passed through in getModalProps() so using props as constructor arg

client/src/components/InsertEmbedModal/InsertEmbedModal.js
```diff
@@ -9,39 +9,31 @@ import { joinUrlPaths } from 'lib/urls';
...
+  const {
```
emteknetnz, line 13: sectionConfig is was unused so does not appear here, though it is passed through in getModalProps()

client/src/components/InsertEmbedModal/InsertEmbedModal.js
```diff
@@ -132,13 +81,53 @@ class InsertEmbedModal extends Component {
...
+    );
+    delete modalProps.sectionConfig;
+    delete modalProps.onInsert;
+    delete modalProps.fileAttributes;
+
+    return modalProps;
+  };
+
+  useEffect(() => {
+    setOverrides({ schemaUrl, fileAttributes });
```
emteknetnz, line 132: schemaUrl and fileAttributes were the only two props that were used in setOverrides()

the logic in setOverrides() is pretty goofy as we're passing in props as a param and we could just remove the param, but I want to keep the code as similar as possible without refactoring

client/src/components/BulkActions/BulkActions.js
```diff
@@ -1,41 +1,42 @@
...
 import PropTypes from 'prop-types';
 
-class BulkActions extends Component {
-  constructor(props) {
-    super(props);
-    this.handleChangeValue = this.handleChangeValue.bind(this);
-    this.renderChild = this.renderChild.bind(this);
-  }
-
+const BulkActions = ({
```
emteknetnz, line 10: total and totalReachedMessage props are not included here as they are unused

client/src/containers/InsertMediaModal/InsertMediaModal.js
```diff
@@ -89,37 +73,59 @@ class InsertMediaModal extends Component {
...
+  });
+
+  /**
+   * Generates the properties for the modal
+   * @returns {object}
+   */
+  const getModalProps = () => {
+    // Remove some props that are not relevant to the modal
+    const modalProps = { ...props };
+    delete modalProps.onHide;
```
emteknetnz: Expanded to manually delete each props that we don't want to prevent unused variable eslint warning

client/src/components/InsertEmbedModal/InsertEmbedModal.js
```diff
@@ -9,39 +9,31 @@ import { joinUrlPaths } from 'lib/urls';
...
+    schemaUrl,
+    targetUrl,
+    onLoadingError,
+    FormBuilderModalComponent = FormBuilderModal
+  } = props;
 
-  componentDidMount() {
-    this.setOverrides(this.props);
-  }
+  const clearOverrides = () => {
```
ChloeSartorelli, line 30:
```suggestion
  /**
    * Clear any overrides that may be in place
    */
  const clearOverrides = () => {
```
was there fore the original

client/src/components/InsertEmbedModal/InsertEmbedModal.js
```diff
@@ -9,39 +9,31 @@ import { joinUrlPaths } from 'lib/urls';
...
+    actions.schema.setSchemaStateOverrides(schemaUrl, null);
+  };
 
-  componentDidUpdate(prevProps) {
-    if (this.props.isOpen && !prevProps.isOpen) {
-      this.setOverrides(this.props);
+  const setOverrides = (obj) => {
```
ChloeSartorelli, line 40:
```suggestion
  /**
    * Compares the current properties with received properties and determines if overrides need to be
    * cleared or added.
    *
    * @param {object} props
    */
  const setOverrides = (obj) => {
```
Was there for the original

client/src/components/InsertEmbedModal/InsertEmbedModal.js
```diff
@@ -52,78 +44,35 @@ class InsertEmbedModal extends Component {
...
 
-  /**
-   * Handler for when loading the form returns an error
-   *
-   * @param error
-   */
-  handleLoadingError(error) {
-    if (typeof this.props.onLoadingError === 'function') {
-      this.props.onLoadingError(error);
+  const handleLoadingError = (error) => {
```
ChloeSartorelli, line 65: Also missing its doc comment

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -13,168 +13,141 @@ import PropTypes from 'prop-types';
...
-      height: CONSTANTS.THUMBNAIL_HEIGHT,
-      width: CONSTANTS.THUMBNAIL_WIDTH,
+    prevDataRef.current = {
+      url: data.url,
+      version: data.version,
+      isInitial: false
     };
-    const securityID = this.props.securityID;
+    prevIdRef.current = id;
+  }, [id, data.url, data.version, actions.previewField]);
```
ChloeSartorelli:
```suggestion
  }, [data.url, data.version]);
```
We don't want this effect to fire if `id` or `actions.previewField` have changed, since the condition explicitly only calls `actions.previewField.removeFile` if `data.url` or `data.version` have changed.

If `id` changes without one of the two things in the condition changing, nothing will happen - ditto with `actions.previewField`.

Having only the two things that we care about as the dependencies also should mean we don't need the condition or the weird ref, since anything else changing won't trigger the effect.

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -187,89 +160,80 @@ class PreviewImageField extends Component {
...
+
+    const parsedUrl = urlLib.parse(url);
+    const parsedQs = { ...qs.parse(parsedUrl.query), vid };
+    return urlLib.format({ ...parsedUrl, search: qs.stringify(parsedQs) });
+  };
 
   /**
    * Build the preview URL
    * @param {string} category
    * @param {object} upload
```
ChloeSartorelli:
```suggestion
   * @param {object} uploadParam
```

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -13,168 +13,141 @@ import PropTypes from 'prop-types';
...
-    this.props.actions.previewField.failUpload(this.props.id, response);
-  }
+  const handleFailedUpload = (file, response) => {
+    actions.previewField.failUpload(id, response);
+  };
 
   /**
    * Handles when a file is added to this field.
    *
    * @param {object} data
```
ChloeSartorelli:
```suggestion
   * @param {object} dataParam
```

tests/behat/features/single-file-upload-field.feature
```diff
@@ -17,8 +17,9 @@ Feature: Single file Upload field
       And I go to "/admin/test/SilverStripe-FrameworkTest-Model-Employee"
       And I click "Allen" in the "#Form_EditForm_SilverStripe-FrameworkTest-Model-Employee" element
       And I should see an ".uploadfield" element
-
+@sboyd
   Scenario: I can select an existing file
+Given I take a screenshot after every step
```
ChloeSartorelli:
```suggestion

  Scenario: I can select an existing file
```

ChloeSartorelli (changes requested):
Mostly inconsequential comments below but one thing that does matter with a `useEffect`.

This was way simpler than I expected for an asset-admin PR :+1:

tests/behat/features/single-file-upload-field.feature
emteknetnz: Updated

client/src/components/InsertEmbedModal/InsertEmbedModal.js
emteknetnz: Updated

client/src/components/PreviewImageField/PreviewImageField.js
emteknetnz: I'm pretty sure we need `id` as a dependency as this is the block that updates `prevIdRef` with `prevIdRef.current = id;`

We're calling `actions.previewField.removeFile(prevIdRef.current);` within the block so `prevIdRef` needs to be kept up to date

I've removed `actions.previewField` since that will never change as it's that just a thing set by redux via function mapDispatchToProps() and shouldn't ever change

client/src/components/PreviewImageField/PreviewImageField.js
emteknetnz: Updated

client/src/components/PreviewImageField/PreviewImageField.js
emteknetnz: Updated

client/src/components/PreviewImageField/PreviewImageField.js
ChloeSartorelli: Just put `id` in a separate effect to update `prevIdRef` if that's needed?

client/src/components/InsertEmbedModal/InsertEmbedModal.js
emteknetnz: Updated

client/src/components/InsertEmbedModal/InsertEmbedModal.js
emteknetnz: Updated

client/src/components/PreviewImageField/PreviewImageField.js
emteknetnz: I'd rather keep it in the useEffect() it's in, because the order matters. The `prevIdRef.current = id;` needs to happen after `actions.previewField.removeFile(prevIdRef.current);`, and that's easier to manage in a single function. Otherwise someone may swap the order of the useEffect() calls. Also it's just more lines of code for no real benefit

ChloeSartorelli: > Also it's just more lines of code for no real benefit

Removing the weird ref removes lines of code - and also makes it easier to tell what's going on which is the benefit.

But given we can't rely on the order of effects I'll let it stay - didn't realise it was used in this effect directly. Needs a comment though or I'll just want to change it again later lol.

client/src/components/PreviewImageField/PreviewImageField.js
```diff
@@ -13,168 +13,141 @@ import PropTypes from 'prop-types';
...
+      url: data.url,
+      version: data.version,
+      isInitial: false
     };
-    const preview = {
-      height: CONSTANTS.THUMBNAIL_HEIGHT,
-      width: CONSTANTS.THUMBNAIL_WIDTH,
-    };
-    const securityID = this.props.securityID;
+    prevIdRef.current = id;
```
ChloeSartorelli, line 56:
```suggestion
    // This must be set after it is used - do not move this into its own separate effect.
    prevIdRef.current = id;
```

ChloeSartorelli (changes requested):

client/src/components/PreviewImageField/PreviewImageField.js
emteknetnz: Updated

ChloeSartorelli (approved):
LGTM, appears to work as expected

Issue #1618, opened 2025-10-08
Accessibility: Arrow key nav in file manager (gallery view)


emteknetnz:
Follow on from https://github.com/silverstripe/silverstripe-asset-admin/pull/1614#issuecomment-3383238903

In the file manager grid and list views, currently you can navigate with tab and shift-tab. You should be able to navigate with the arrows keys. 

This should be easy to implement on the list view and it's just up/down, though it may be harder to implement on the grid view as the files will dynamically wrap based on the the browser width, so there will need to be some dynamic way to work out the correct file to navigate based on the number of files in the row.

## Related

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1652

## Acceptance criteria

- [x] The gallery ~~and table are both~~ is a tab island
- [x] Up/down/left/right arrow keys can be used for navigation in the file grid view
- [ ] ~~Up/down arrow keys can be used for navigation in the file list view~~

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1643
- https://github.com/silverstripe/silverstripe-admin/pull/2098
- https://github.com/silverstripe/developer-docs/pull/867

ChloeSartorelli:
As discussed in the linked issue, if we can't have arrow navigation in gallery view we probably shouldn't in table view either. Note also that arrow key navigation to move between table rows doesn't seem to be a standard practice so may not be appropriate.

emteknetnz:
> Note also that arrow key navigation to move between table rows doesn't seem to be a standard practice so may not be appropriate

I wouldn't focus too much on "it's a table row", and more on "it's a file manager"

In Dolphin (default file manager for kubuntu which I'm using). The keys it used are:

Gallery/Grid view
- Tab - move between different sections of the (Location, Tabs, Files)
- Up/down/left/right - move between different files
- Space - enter selection mode (for multiple files)
- Enter - open file

List/Table view
- Tab - move between different sections of the (Location, Tabs, Files)
- Up/down - move between different files
- Left/right - close/open sub folders
- Space - enter selection mode (for multiple files)
- Enter - open file

ChloeSartorelli:
> I wouldn't focus too much on "it's a table row", and more on "it's a file manager"

HTML is semantic, and screen readers especially use semantic HTML for a lot of their guidance. The fact that we're using a table is extremely relevant, unless there are aria roles we can use to say "this is a file manager" to a screen reader, which I don't think there are.

[...]

PR #1643, opened 2026-01-09
ENH Use arrow keys to navigate asset gallery

ChloeSartorelli:
- Turns the gallery into a tab island, i.e. a single tab to get into the gallery and then another tab to get back out
- Removes partial keyboard accessibility of draggable element, which wasn't correctly implemented and was only getting in the way. Moving files is the only functionality dragging provides, and that's available by selecting the file(s) and using the actions panel
- Adds appropriate keyboard events once tabbed into the gallery
    - Arrow keys navigate around the gallery, including between the file/folder boundary.
    - Home/End keys go to the start/end of the row - hold ctrl to go the the first/last item in the whole grid
    - Space key selects/deselects the currently focused grid item
    - Enter key:
        -  opens edit form for files
        - navigates to the folder for folders
        - cancels in-progress uploads
        - removes failed uploads
- Adds a `focus-within` outline around the grid container to give a hint that arrow-key navigation is available
- Provides a good screen reader experience when tabbing into, navigating around, and selecting items within the grid

## Resources used

> [!NOTE]
> No W3C pattern examples were directly 100% applicable for our use case:
> - we're using a flex grid (i.e. it updates the rows/columns during runtime)
> - we can select cells
> - we can "activate" cells to "navigate" (sometimes navigating away to view a folder, sometimes just opening an edit form inside the same page).

Here's the main resources used to piece this together:

- The [W3C aria roving tabindex guide](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex)
- keyboard support suggestions from [the W3C grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/).
    - I have opted not to include advanced selection hotkeys for now (e.g. selecting a whole row or column, selecting all, and moving while selecting) but can add those if it's deemed worthwhile.
    - I'm not using page up or page down
- General setup as defined by [the W3C grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/#wai-ariaroles,states,andproperties) as far as it's applicable
    - Note that when this refers to "content editing capabilities" it's talking about editing the content that exists in the cell directly, not editing with a separate interface like we do.
- some attributes from [the W3C layout grid examples](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/layout-grids/#rps_label) as applicable
- some of the accessibility feature suggestions from [the W3C nagivation treeview example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/#accessibilityfeatures) since none of the grid examples include activating a grid item to navigate like we're doing.
    Mostly about `aria-current`, moving tabindex to the "current" item, and adding a border around the container (in this case the grid itself).
- Feedback from the UX team - see https://github.com/silverstripe/silverstripe-asset-admin/pull/1643#issuecomment-3781423666

## ~~Things I'm unsure about~~

These have been resolved, see https://github.com/silverstripe/silverstripe-asset-admin/pull/1643#issuecomment-3781423666

<details>
<summary><s>When reviewing, please read and give your opinion on each of these.</s></summary>

1. The "editable" (or "not editable" if the user can't edit that folder/file) audible label might not be useful as in most cases all files and folders will be editable. Should this be ommitted, and only "not editable" is announced if the file/folder isn't editable? But then again we need to make sure the user knows they can "activate" the cell using the enter key somehow.
    - Note that regardless of the decision on this, I want to keep "selectable" since that's not immediately obvious without visually seeing the checkbox
    - Should we have some other audible label for folders? Like "openable"? (since pressing enter on a file opens the edit form, but pressing enter on a folder navigates to that folder)
1. I provided a count of folders and files as an aria description of the grid, which is useful since non-visual users can't just glance and see roughly how many folders there are for example. But maybe this text should be visible? It might be useful for visual users as well.
1. Some (maybe all?) screen readers don't announce that there's pagination for the grid until you tab to the pagination element. Should we add the number of pages and current page to the grid's description? I couldn't find much information about grids with non-scrollable pagination.
1. Currently when there's an error (e.g. validation error when uploading an item) there is a visual `x` element that, when clicked, removes the file from the gallery view. This isn't currently keyboard accessible nor mentioned at all via the screen reader.
    - Should we announce this is available (e.g. by announcing "removable") and allow activating this button via keyboard?
    - If so, should this be activated with Space, Enter, or both?
    - Note that anything beyond this for errors is out of scope - see https://github.com/silverstripe/silverstripe-asset-admin/issues/1651
1. Similarly to above, currently there's no indication to non-visual users when the actual file is missing from the filesystem. For visual users (at least for images - probably not for documents) this is immediately obvious. Should we announce this?
    - My instinct is that we spin off a separate card for this, since even visual users don't have a great indicator of this for non-image files.
    
</details>

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1618


tests/behat/files/assets-pagination-limit-10.yml
ChloeSartorelli, line 1: Renamed (and updated references to it) because I now need a different specific number of items per page.
The new name makes a clear distinction between the two config files.

tests/behat/features/accessibility-asset-admin.feature
```diff
@@ -3,12 +3,14 @@ Feature: Accessibility asset-admin
   As a user with accessibility needs
   I want to be able to use asset-admin with accessibile technology
 
+  Background:
+    And the "group" "EDITOR" has permissions "Access to 'Files' section" and "FILE_EDIT_ALL"
+    And I am logged in as a member of "EDITOR" group
+
   Scenario: Tab navigation on react edit form using arrow keys loads focused tab
-    # Note - othres react edit forms e.g. linkfield will use the same tab nav code
+    # Note - other react edit forms e.g. linkfield will use the same tab nav code
```
ChloeSartorelli, line 11: Quick typo fix

tests/behat/features/accessibility-asset-admin.feature
```diff
@@ -3,12 +3,14 @@ Feature: Accessibility asset-admin
...
     # so only testing this here
     Given a "image" "assets/file1.jpg"
-    And the "group" "EDITOR" has permissions "Access to 'Files' section" and "FILE_EDIT_ALL"
-    And I am logged in as a member of "EDITOR" group
```
ChloeSartorelli, line 11: Moved to background since all scenarios will want to use these same permissions

tests/behat/features/accessibility-asset-admin.feature
ChloeSartorelli, line 1: Using behat instead of jest to test keyboard navigation because the way we're mocking etc makes it difficult to test in jest, since the key events aren't triggered on the same component that handles them.

client/lang/src/en.json
ChloeSartorelli, line 1: I think I caught all of the new i18n keys but please keep an eye out for any I missed, since text collection doesn't work in JS.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -559,10 +602,6 @@ GalleryItem.propTypes = {
   item: fileShape,
   loadState: PropTypes.oneOf(Object.values(IMAGE_STATUS)),
   bustCache: PropTypes.bool,
-  // Can be used to highlight a currently edited file
-  highlighted: PropTypes.bool,
-  // Styles according to the checkbox selection state
-  selected: PropTypes.bool,
```
ChloeSartorelli, line 565: These aren't actually props - these are properties on the `item` object (which itself is passed as a prop)

client/src/components/GalleryItem/draggable.js
ChloeSartorelli, line 1: We need to disable all of the accessibility attributes. Dragging and dropping these isn't really feasible with the keyboard and these were just confusing the screen reader.

Done in an optional way to prevent breakage if anyone is doing wild things with these components.

client/src/components/GalleryItem/draggable.js
```diff
@@ -2,18 +2,47 @@ import React from 'react';
...
+    }
+
+    const setRef = (el) => {
+      setNodeRef(el);
+      // The ref used for keyboard navigation needs to be set on this element,
+      // because the margin applied to it factors into the items per row calculation.
+      if (typeof props.draggableSizeRef === 'function') {
+        props.draggableSizeRef(el);
+      }
+    };
```
ChloeSartorelli, line 38: This is used here for files and in droppable for folders

client/src/components/GalleryItem/tests/GalleryItem-test.js
ChloeSartorelli, line 1: New tests here broadly mirror the tests for Enter and Space in this file.

client/src/containers/Gallery/Gallery.js
```diff
@@ -797,6 +795,7 @@ const Gallery = ({
       sectionConfig,
       canDrag: type === ACTION_TYPES.ADMIN,
       maxFilesSelect,
+      hasOpenedItem: hasOpenedItem(),
```
ChloeSartorelli: We need to know if an item is open to determine whether tabindex stays on the first item or moves to the currently active item.

client/src/containers/Gallery/Gallery.js
```diff
@@ -761,9 +761,7 @@ const Gallery = ({
     const mappedFiles = files.map((file) => {
       const selected = itemIsSelected(file.id);
       const highlighted = itemIsHighlighted(file.id);
-      const key =
-        (file.queuedId ? `queueId${file.queuedId}` : `id${file.id}`) +
-        (selected ? '--selected' : '');
+      const key = (file.queuedId ? `queueId${file.queuedId}` : `id${file.id}`);
```
ChloeSartorelli, line 764: The key MUST NOT change when selecting the item! This was causing the component to be completely thrown away and reconstructed whenever the selected status changed, which broke focus.

client/src/containers/Gallery/Gallery.scss
```diff
@@ -47,6 +47,12 @@
   }
 }
 
+.gallery__main-view--tile:has(.gallery-item:focus-visible) {
+  outline: var(--focus-outline-width) var(--focus-outline-style) var(--focus-outline-color);
+  // Add additional room for the outline of the items inside (assuming they use the same variable values).
+  outline-offset: calc(var(--focus-outline-offset) * 2 + var(--focus-outline-width));
+}
```
ChloeSartorelli, line 54: As per W3C advice, we add an outline to the whole gallery to visually indicate that arrow keys can be used to navigate.

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -60,14 +118,125 @@ class ThumbnailView extends Component {
...
+          folders.length - 1
+        );
+      } else {
+        nextIndex = folders.length - 1;
+      }
+    }
+
+    // Prevent default scroll behavior and move focus if there's something to move to
+    if ((allowSameIndex || nextIndex !== index) && nextIndex >= 0 && nextIndex < items.length) {
+      event.preventDefault();
```
ChloeSartorelli: It could be argued that we should prevent default regardless of whether we're changing focus - I couldn't find any advice on this. Gemini said to do it only when changing focus, and that feels natural to me when using it (e.g. pressing down when on the last item scrolls down a bit instead of just nothing happening)

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -86,8 +255,37 @@ class ThumbnailView extends Component {
       selectedFiles,
       badge,
       canDrag: this.props.canDrag,
+      onNavigateKeyDown: (event) => this.handleNavigateKeyDown(event, index, item.type),
+      tabIndex: item.highlighted ? 0 : -1,
```
ChloeSartorelli: Tab Index is "0" on the first item and -1 on everything else.
Unless an item is open, and then the open item has the "0" value instead of the first item.

This seems consistent with all of the W3C examples.

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -109,22 +308,71 @@ class ThumbnailView extends Component {
...
+    />;
   }
 
   render() {
+    // filter files and folders
+    const folders = this.props.files.filter(this.folderFilter);
+    const files = this.props.files.filter(this.fileFilter);
     const className = 'gallery__main-view--tile';
+    const numFoldersText = i18n.inject(i18n._t('AssetAdmin.NUM_FOLDERS', '{count} folders'), { count: folders.length || '0' });
+    const numFilesText = i18n.inject(i18n._t('AssetAdmin.NUM_FILES', '{count} files'), { count: files.length || '0' });
```
ChloeSartorelli: ` || '0'` is necessary here - otherwise if there are 0 items it doesn't do a replacement and we get literally "{count} folders" in the DOM

emteknetnz (changes requested):
Double border around all the files and the border around the currently "active" file is kind of weird and confusing as it current is.. Since both borders and the same widths it's not clear which element is "active". Maybe this would be better if the the border around all the files was less prominent (either a thinner semi-transparent border, or change the background color of the container? Probably though just get rid of it since we don't use that pattern anywhere else (and I'd rather not put it everywhere else e.g. groups of tabs, model admin list once that's done, etc).

If I am not editing a file, and I move the active file using arrow keys, and then shift-tab, the it changes the active file to the first file. Instead what should happen is that focus shifts to the previous tab island, and if I tab back into the file tab-island, the active file should be the one I previously used arrow keys to get to

If I press enter to select the third file, then I alt-tab from the detail form a few times back to the file manager, I'll end up on the file I select, which is good. However if I then press the left arrow key, then tab again, I end up on the file that is being edited. Instead I should go the next tab-island.

Home/End go the the start/end of the row, which is good. However pressing them a second time then changes the behavior of those keys and goes back to browser defaults, which is to go to the top/bottom of the document, causing the file manager to scroll. Instead they keys should do essentially nothing at that point.

ChloeSartorelli:
> When reviewing, please read and give your opinion on each of these.

Please find the above text in the PR description and give your opinion about those items.

ChloeSartorelli:
> If I am not editing a file, and I move the active file using arrow keys, and then shift-tab, the it changes the active file to the first file. Instead what should happen is that focus shifts to the previous tab island, and if I tab back into the file tab-island, the active file should be the one I previously used arrow keys to get to

> If I press enter to select the third file, then I alt-tab from the detail form a few times back to the file manager, I'll end up on the file I select, which is good. However if I then press the left arrow key, then tab again, I end up on the file that is being edited. Instead I should go the next tab-island.

Is there an example you're getting these expected behaviours from? If not we should probably get the UX team to weigh in since our expectations differ here. I was loosely basing the roving tab index behaviour on [the W3C nagivation treeview example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) in the absence of a navigation grid example.

That example doesn't have a case where there is no open page so it's unclear how your first scenario would react, but it does behave identically to this PR for your second scenario.

> Home/End go the the start/end of the row, which is good. However pressing them a second time then changes the behavior of those keys and goes back to browser defaults, which is to go to the top/bottom of the document, causing the file manager to scroll. Instead they keys should do essentially nothing at that point.

This was intentional, see https://github.com/silverstripe/silverstripe-asset-admin/pull/1643#discussion_r2692619127
As noted there, I couldn't find any advice one way or another, so I did what both felt natural and was suggested by Gemini AI.

Since our expectations differ here the UX team should probably weigh in on this, too.

Once you have opined on the items I'm unsure about from the PR description I'll make any changes that come out of that, and then if needed I'll deploy this to our testing environment so the UX team can give suggestions for the above.

ChloeSartorelli:
> Double border around all the files and the border around the currently "active" file is kind of weird and confusing as it current is.. Since both borders and the same widths it's not clear which element is "active". Maybe this would be better if the the border around all the files was less prominent (either a thinner semi-transparent border, or change the background color of the container? Probably though just get rid of it since we don't use that pattern anywhere else (and I'd rather not put it everywhere else e.g. groups of tabs, model admin list once that's done, etc).

I'm not sure what you mean by "double border around all the files" - there is a pre-existing border that was always there, and there's now an outline on the focused item. The border disappears when the outline is present.

I also don't know what you mean by "active" - is that a file that is open for editing? Or the currently focused item? Here they are side-by-side:

image

And here's an open file that has focus:

image

Each only has a single border/outline at any given moment. The focus outline is visually distinct from any other border, and is consistent with our focus outline that is used everywhere else in the CMS now.

emteknetnz:
I don't have any source / best-practice for my feedback, it's just that I found the experience to be unpredictable/unexpected
- the active file changing when I tab in and out of the tab-island is unexpected
- the tab key being used to move to a different file is unexpected (and really just a bug)
- the home/end keys changing what they do based on the currently active file is unexpected

> Double border

This is the double border I'm referring to - the border around the whole file container

image

> Please find the above text in the PR description and give your opinion about those items.

1. I'd say don't accounce. A parallel here is we only show the 'protected' file icons on protected files (those with a level of permission on them), we don't show icons for regular unprotected files.

2. Probably not? I think it would get confusing with pagination since that aria description only shows how many files are on the current pagiginated page, not how many files there are in total which is what I'd expect as a user. And also a "total" file count may undercount due to how Silverstripe handles canView() checks i.e. to get an accurate count you have to perform canView() checks on ALL files in that folder which could be thousands. Given that probably makes sense to update the aria-description to say that the page count is only for the current page, not the total.

3. Probably makes sense to announce on the grid to provide context

4. Yes announce, and yes should be keyboard accessible, activate with "enter"

5. Yes and spin off a card

GuySartorelli:
**WIP** because github decided to ignore my "review in progress" I made to hide comments until I'm ready.

To confirm:

1. Right now in this PR the current item to have focus will always have `tabindex=0`. Tabbing away and tabbing back will focus that same item again. This is true even if a file is open for editing.
    An alternative approach would be that when you tab off the gallery and then shift+tab back into the gallery, if a file is open that file in the gallery gains focus even if it wasn't the previously focused item.
    - Should we focus on the file that is open if there is one? Or retain the current behaviour?
    - This kinda falls under "If the design calls for a specific element to be focused the next time the user moves focus into the composite" from [the roving tabindex guide](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex) so I guess the question is: does this design call for the file currently open for editing to be focused the next time the user moves focus into the gallery?"
1. This PR reverts to default browser behaviour on key inputs for arrow keys and home/end keys if pressing those keys doesn't move focus. There's no advice that I can find about whether this is appropriate or not.
    - Should we do this, or always ignore browser default behaviour of these key presses inside the tab island?
1. This PR provides an outline around the whole tab island when anything inside the tab island has focus. This follows advice from [the W3C nagivation treeview example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/#accessibilityfeatures) - though this is the only place I have seen this advice.
    - Should we keep this outline, or remove it?
    - If keeping it, should the styling be adjusted at all or left as is?

Plus the "things I'm unsure about"

ChloeSartorelli:
I've made the following changes:

1. When using the arrow keys or home/end to navigate within the gallery, the tabindex is per the [W3C aria guide for roving tabindex](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex).
1. Pagination information is provided in the roving index container for screen readers
1. Files in the process of being uploaded can be cancelled with the enter key
1. Failed uploads can be removed with the enter key

ChloeSartorelli:
Things for the UX team to advise on:

1. Right now in this PR the current item to have focus will always have `tabindex=0`. Tabbing away and tabbing back will focus that same item again. This is true even if a file is open for editing.
    An alternative approach would be that when you tab off the gallery and then shift+tab back into the gallery, if a file is open that file in the gallery gains focus even if it wasn't the previously focused item.
    - Should we focus on the file that is open if there is one? Or retain the current behaviour?
    - This kinda falls under "If the design calls for a specific element to be focused the next time the user moves focus into the composite" from [the roving tabindex guide](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex) so I guess the question is: "does this design call for the file currently open for editing to be focused the next time the user moves focus into the gallery?"
1. Right now if you add new items (i.e. upload new files) the tabindex doesn't change - whatever item had focus in the gallery last will have focus again when you tab into the gallery.
    - Should tabindex reset to the first folder (or file if there's no folders) when uploading files?
1. Right now mouse events don't change which item will recieve focus when you tab into the gallery.
    - Should they? If so, which events, when?
1. This PR reverts to default browser behaviour on key inputs for arrow keys and home/end keys if pressing those keys doesn't move focus. There's no advice that I can find about whether this is appropriate or not.
    - Should we do this, or always ignore browser default behaviour of these key presses inside the tab island?
1. This PR provides an outline around the whole tab island when anything inside the tab island has focus. This follows advice from [the W3C nagivation treeview example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/#accessibilityfeatures) - though this is the only place I have seen this advice.
    - Should we keep this outline, or remove it?
    - If keeping it, should the styling be adjusted at all or left as is?
1. The gallery container has a screen reader helper that gives pagination information, e.g "Page 1 of 5, 3 folders and 47 files on this page". The page information is visually available but the number of folders/files isn't.
    - Should we add the "x folders and y files on this page" information visually to the page somewhere? If so where and with what exact text/styling?
    - Any other suggestions around this?
1. Every file and folder that can be edited has a screen reader "editable" label, and non-editable ones have "not editable". The aim of this was to indicate that pressing enter to "activate" the editable item will allow editing, but this is only true of files. For folders pressing enter will navigate to that folder.
    - How can we make it more obvious what actions are available for screen reader users?
    - Should we omit the "editable" label? Or maybe keep it, but replace it with "openable" on folders?
    - Should we add a screen reader label to the gallery as a whole that says something like "Use arrow keys to navigate the gallery. Enter on a file edits the file. Enter on a folder navigates to the folder. Space on either a file or folder selects the item"?
        - How much information do we want cram into something like that?
    - What about uploads which can be cancelled with "enter" and failed uploads which can be removed with "enter"?

ChloeSartorelli:
Answers to the above discussed with Cassie and Steve:

1. If you've used arrow keys to focus on an item, that item should get focus if you move out of and back into the container (even if you are editing a different item)
2. If adding items to the container (e.g. upload file), move focus to the new item - that item gets the roving index
3. Mouse events which would ordinarily set focus to a `focusindex=0` element should update the item that has the roving index to the item that was interacted with
4. Navigation keyboard events that would normally change focus should not revert to browser default behaviours
5. dashed line for container outline
6. Number of items per page not rendered for sighted users - but keep for screen reader
7. 
    - Remove "editable" for folder (it's misleading, you don't edit a file from here you open it)
    - Remove "editable" for file (it's redundant - all files are editable unless explicitly stated otherwise)
    - On upload button provide information about cancelling uploads
    - "press enter to remove" for failed uploads

tests/behat/files/invalid.file
ChloeSartorelli, line 1: Renamed (and updated references to it) so I can make assertions about a file named "invalid" instead of about a file named "file" to make things a bit more explicit, and to avoid problems with fuzzy matching (e.g. "file" might match for "file01" in some cases)

Also has the nice side-effect that in screenshots you'll clearly see the word "invalid" instead of just "file"

tests/behat/features/accessibility-asset-admin.feature
```diff
@@ -27,3 +29,202 @@ Feature: Accessibility asset-admin
     And I should not see "Who can view this file?"
     When I press the "Tab" key globally
     Then the "#Form_fileEditForm_Title" element should have focus
+
+  Scenario: Keyboard navigation within the gallery view
```
ChloeSartorelli, line 33: I don't really like making a single massive scenario like this, but the alternatives are:

1. Multiple scenarios in this file, but they all repeat the fixture setup since moving that to "Background" would mess with the above scenario
2. Using a different file, but we've agreed on having a single "accessibility" feature file

I don't really like the single "accessibility" feature file approach (for exactly this reason) so I'd be okay with moving these into a separate file, but leaving them here for now as this approach is what we previously agreed on.

client/lang/src/en.json
```diff
@@ -32,6 +32,12 @@
     "AssetAdmin.BULK_ACTIONS_SELECTED": "%s selected",
     "AssetAdmin.BULK_ACTIONS_SELECT_ALL": "Select all",
     "AssetAdmin.CANCEL": "Cancel",
+    "AssetAdmin.CATEGORY_archive": "archive",
```
emteknetnz: Looks weird being mixed case, should be `"AssetAdmin.CATEGORY_AUDIO"` etc

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -536,13 +603,20 @@ const GalleryItem = (_props) => {
         {getStatusFlags()}
         {getStatusIcons()}
       </div>
+      <span className="visually-hidden">
+        {i18n._t(`AssetAdmin.CATEGORY_${item.category}`, item.category)},&nbsp;
```
emteknetnz:
```suggestion
        {i18n._t(`AssetAdmin.CATEGORY_${item.category}`, item.category.toUpperCase())},&nbsp;
```

To match change in en.json to remove mixed case keys

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -1,17 +1,213 @@
...
 import Paginator from 'components/Paginator/Paginator';
+import ResizeAware from 'components/ResizeAware/ResizeAware';
 import PropTypes from 'prop-types';
 
+// Generates a unique ID each time this function is called,
+// even between components.
+// When ThumbnailView is converted to a functional component, use
+// the useId hook instead.
+let idCounter = 0;
+function generateUniqueId(prefix) {
```
emteknetnz:
```suggestion
const generateUniqueId = (prefix) => {
```

tests/behat/src/FixtureContext.php
```diff
@@ -538,4 +538,31 @@ public function stepDragTheFileToTheBackButton(string $fileName): void
...
+    /**
+     * Example: Then the file named "file1" should have focus
+     *
+     * @Then /^the (?:file|folder) named "([^"]+)" should (not |)have focus$/
+     */
+    public function theItemShouldHaveFocus(string $name, string|bool $not): void
+    {
+        $file = $this->getGalleryItem($name)?->getParent();
+        Assert::assertNotNull($file, "File named {$name} could not be found");
+        $xpath = $file->getXpath();
```
emteknetnz, line 551:
```suggestion
        $xpath = json_encode($xpath);
```
This will automatically quote strings and for weird edge cases like newlines or unicode

emteknetnz (changes requested):
When i upload a file with an invalid file extension I don't get the dashed line around the container, or the regular focus state as if I just used arrow keys to navigate to it, though it does get a thin border around it - the same border you see around files that are currently being edited

> Mouse events which would ordinarily set focus to a focusindex=0 element should update the item that has the roving index to the item that was interacted with (from https://github.com/silverstripe/silverstripe-asset-admin/pull/1643#issuecomment-3781423666)

(I assume you meant tabindex=0) I don't think this is working correctly - when I use arrow keys to got file B, and then I click on file C, and I shift-tab back to the file container, I'm on file B, I would expect to be on file C

tests/behat/src/FixtureContext.php
ChloeSartorelli: I assume you mean `json_encode($file->getXpath())` since there was no `$xpath` assigned before this.
Done

client/src/containers/ThumbnailView/ThumbnailView.js
ChloeSartorelli: Done to avoid ping pong

client/src/components/GalleryItem/GalleryItem.js
ChloeSartorelli: Done (though as part of the key, not the value)

client/lang/src/en.json
ChloeSartorelli: Done

ChloeSartorelli:
> When i upload a file with an invalid file extension I don't get the dashed line around the container, or the regular focus state as if I just used arrow keys to navigate to it, though it does get a thin border around it - the same border you see around files that are currently being edited

If you were using your mouse, this is because the browser has decided the mouse interaction shouldn't create a `focus-visible` state. Focus has been set, though.

If you use your keyboard, the browser will set a `focus-visible` state and you'll see those outlines.

tests/behat/src/FixtureContext.php
ChloeSartorelli: Nevermind, reverting this - it causes failures.

> Then the folder named "folder01" should not have focus # SilverStripe\AssetAdmin\Tests\Behat\Context\FixtureContext::theItemShouldHaveFocus()
      javascript error: Invalid or unexpected token

If we run into a scenario where the xpath has unexpected characters we can deal with it then. I think the chances of that are very low.

ChloeSartorelli:
>> Mouse events which would ordinarily set focus to a focusindex=0 element should update the item that has the roving index to the item that was interacted with (from https://github.com/silverstripe/silverstripe-asset-admin/pull/1643#issuecomment-3781423666)
>
> (I assume you meant tabindex=0) I don't think this is working correctly - when I use arrow keys to got file B, and then I click on file C, and I shift-tab back to the file container, I'm on file B, I would expect to be on file C

Ahh yup I'll look into that.
Edit: Fixed and added a test for it.

emteknetnz (approved):

Issue #1652, opened 2026-01-14
Accessibility: Arrow key nav in file manager (list view)


ChloeSartorelli:
Follow on from https://github.com/silverstripe/silverstripe-asset-admin/pull/1614#issuecomment-3383238903

In the file manager grid and list views, currently you can navigate with tab and shift-tab. You should be able to navigate with the arrows keys. 

This should be easy to implement on the list view and it's just up/down.

## Related

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1618

## Acceptance criteria

- [ ] The table is a tab island
- [ ] Up/down arrow keys can be used for navigation in the file list view
- [ ] Appropriate W3C suggestions are applied where applicable

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1614
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1661

ChloeSartorelli:
PR(s) merged. This will be included in the next minor release.

PR #1661, opened 2026-02-03
ENH Use arrow keys to navigate asset table view

ChloeSartorelli:
> [!IMPORTANT]
> This implementation sets a precedent for other data grids such as gridfield and campaign admin, both for development and for user expectations. Keep those in mind during review.

Broadly, this PR does the following:

- Turns the table view into a tab island (roving tabindex section)
- Allows focusing on the sortable header buttons and activating them with the keyboard
- Allows focusing on table rows and activating/selecting them with the keyboard
- Provides appropriate aria attributes for a good screen reader experience within the table

## Resources used

- https://github.com/silverstripe/developer-docs/issues/866
- [W3C treegrid example](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/) for allowing the row as a whole to be focused
- [W3C grid pattern](https://www.w3.org/WAI/ARIA/apg/patterns/grid/)
- [W3C data grid examples](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/)
- [W3C roving tabindex guidelines](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex)

### space/enter within a cell

I've opted to not let space/enter work when focused on a cell - instead you must go back to focus on the row as a whole. This is for a couple of reasons:
1. It allows us to add interactive elements to cells later on without altering expectations
2. It is consistent with how other data grids, such as gridfields, will need to be. e.g. gridfield has a cell that contains actions for the row, and can contain arbitrary interactable elements in any cell.

### focusing on sortable header cells

I've opted to have the sortable header cells remain in the tab order instead of being part of the roving tabindex because:
1. Having two extra tabs doesn't increase much friction
1. Having these in the tab order makes them arguably easier to access, e.g. if you're already halfway down the table you can just shift+tab back to them rather than going up or remembering to hit home
1. It keeps the home/end functionality more predictable - you always go to the first/last _record_ in the table - regardless of whether that table has sortable header cells or not
1. If the header cells were part of the roving tabindex we suddenly have some new expectation questions to ask, such as:
    - does the whole header row need to be focusable? That's not useful to do and may be confusing
        - if not, what happens if you're focused on the top row and hit up? Do you go up to an arbitrary cell in the header row?
        
I've also opted not to include non-sortable header cells in the tab order since there's no action to be performed on them (so useless for sighted users to focus on them) and the header information is presented when focusing on data cells anyway. This follows the (admittedly loosely worded) [information from W3C's grid pattern](www.w3.org/WAI/ARIA/apg/patterns/grid/#datagridsforpresentingtabularinformation):
> if column or row header cells do not provide functions, such as sort or filter, they do not need to be focusable.
        

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1652


client/src/components/GalleryItem/GalleryItem.scss
```diff
@@ -183,10 +183,12 @@ $gallery-item-label-height: 40px;
...
-  .gallery-item--selectable .gallery-item__checkbox-label:hover .gallery-item__checkbox-icon:before {
-    border: 1px solid $component-active-border-color;
-  }
+// This is a bit of an ugly selector but it's all needed. It boils down to:
+// - Selectable items only
+// - Hovering in the checkbox label (but only if it's not disabled)
+// - select the :before pseudo-element in the checkbox icon
+.gallery-item--selectable .gallery-item__checkbox-label:not(.gallery-item__checkbox-label--disabled):hover .gallery-item__checkbox-icon:before {
+  border: 1px solid $component-active-border-color;
 }
```
ChloeSartorelli, line 192: This is technically a bug fix, saw this while I was getting the checkbox to work for the table view.
The hover styling just wasn't working at all for the gridview checkbox, and this fixes it.

IMO not really worth fixing on its own, but small enough of a change to bundle in here.
If it's adding too much peer review friction lemme know and I'll split it out

client/src/containers/Gallery/Gallery.js
```diff
@@ -440,7 +440,10 @@ const Gallery = ({
    * @param {Event} event
    */
   const toggleSelectConcat = (event) => {
-    actions.gallery.setConcatenateSelect(isConcat(event));
+    // Tile view doesn't use the click-to-drag selection functionality
+    if (view === 'tile') {
+      actions.gallery.setConcatenateSelect(isConcat(event));
+    }
```
ChloeSartorelli, line 448: This was needed to avoid the whole table re-rendering whenever the shift key was held down, which confused the screen reader on shift+tab

client/src/containers/TableView/TableView.js
```diff
@@ -16,11 +17,460 @@ import {
...
+      return false;
+    }
+    // Check if there are any files in one array which aren't present in the other.
+    // If there's no descrepencies, we return true.
+    return arrA.every(itemA => arrB.some(itemB => filesAreIdentical(itemA, itemB)));
+  }
+
+  // IMPORTANT
+  // A lot of the logic here is similar to logic in ThumbnailView.
+  // If a change is needed (e.g. to resolve a bug), make sure you check both views!
```
ChloeSartorelli, line 84: I did think about abstracting this out into a `lib/roving-tabindex-lifecycle.js` or similar - but IMO while the logic is largely the same, the way indexes and state are managed are different enough that having a single shared function would be messier than just having them separate.

client/src/containers/TableView/TableView.js
```diff
@@ -64,13 +522,6 @@ function TableView(_props) {
     props.onSetPage(page);
   }
 
-  /**
-   * Avoids the browser's default focus state when selecting an item.
-   */
-  function preventFocus(evt) {
-    evt.preventDefault();
-  }
```
ChloeSartorelli, line 72: Was only applied to the `<input type="checked">` element which we are removing (see below)

client/src/containers/TableView/TableView.js
```diff
@@ -170,24 +621,30 @@ function TableView(_props) {
...
-        onMouseDown: (evt) => preventFocus(evt),
-      };
-      const maxSelected = (
-        ![null, 1].includes(props.maxFilesSelect) &&
-        props.selectedFiles.length >= props.maxFilesSelect
-      );
-      if (maxSelected && !rowData.selected) {
-        checkboxProps.disabled = true;
-      }
-      return <input {...checkboxProps} />;
```
ChloeSartorelli, line 188: An actual semantic `input` element is counterproductive here.
- Screen reader users will get information about whether the row is selected from the `<tr>` element being focused. They aren't able to focus on the `input` element at all.
- Sighted users don't care what the underlying markup is

Beyond this, when I kept the `input` element here, the visual display of whether the checkbox was checked wouldn't update when clicking the checkbox directly, even though it updated when clicking the cell. It's not clear why - the `rowData.selected` property _was_ updating, and rendering that here in any other way visually updates as expected.

tl;dr the `<input>` isn't needed and was getting in the way. Having a fake checkbox resolves the problem and is fine because the accessibility is provided by the grid pattern.

client/src/containers/TableView/TableView.js
```diff
@@ -199,7 +656,13 @@ function TableView(_props) {
...
     }
     moment.locale(i18n.detectLocale());
-    return <span>{ moment(rowData.lastEdited).format('L LT') }</span>;
+    const momentInstance = moment(rowData.lastEdited);
+    return <time
+      aria-label={momentInstance.format('LLL')}
+      dateTime={momentInstance.format('YYYY-MM-DD[T]HH:mm')}
+    >
+      {momentInstance.format('L LT')}
+    </time>;
```
ChloeSartorelli, line 673: This improves the semantics, as well as giving screen reader users a better experience.
For screen reader users the output is now something like "February eleven, two thousand and twenty six two colon fourty four PM" (and is localised via moment.js).

It used to be something more like "zero two slash eleven slash two thousand and twenty six two colon fourty four PM" which was just unintelligible.

client/src/containers/TableView/TableView.js
```diff
@@ -209,42 +672,100 @@ function TableView(_props) {
...
-        styles.backgroundImage = `url("${url}")`;
+    const thumbnailClassNames = classNames({
+      'gallery__table-image': true,
+      'gallery__table-image--error': errorMsg,
+      [`gallery__table-image--${category}`]: !errorMsg,
+    });
+    if (!errorMsg && category === 'image' && url) {
+      styles.backgroundImage = `url("${url}")`;
+    }
+    return <div title={errorMsg} aria-label={errorMsg} className={thumbnailClassNames} style={styles} />;
```
ChloeSartorelli, line 696: Technically unrelated refactoring but I added the `classNames` import to make variable classname adding a little easier for some new classes, and it made sense to tidy up bits like this while I was at it to also use the import.

client/src/containers/TableView/TableView.js
```diff
@@ -209,42 +672,100 @@ function TableView(_props) {
...
+      'gallery__table-row--selected': rowData.selected,
+    });
   }
 
   /**
    * Returns the CSS class for a cell, for both header and body cells
    */
   function getCellClassName(cell, isHeader) {
-    const ret = [];
+    const ret = ['gallery__table-column'];
```
ChloeSartorelli, line 773: Added for CSS purposes

client/src/containers/TableView/TableView.js
```diff
@@ -253,19 +774,34 @@ function TableView(_props) {
       ret.push('gallery__table-column--image');
     } else if (columnId === 'title') {
       ret.push('gallery__table-column--title');
+    } else if (columnId === 'status') {
+      ret.push('gallery__table-column--status');
+    } else if (columnId === 'size') {
+      ret.push('gallery__table-column--size');
+    } else if (columnId === 'lastEdited') {
+      ret.push('gallery__table-column--modified');
```
ChloeSartorelli, line 795: Added for behat purposes

client/src/containers/TableView/TableView.js
```diff
@@ -209,42 +672,100 @@ function TableView(_props) {
...
     if (isHeader) {
-      const [sortColumn, sortDirection] = props.sort.split(',');
-      if (sortColumn === columnId) {
-        ret.push(sortDirection === 'asc' ? 'sort-ascending' : 'sort-descending');
+      const sortOrder = getSortOrder(columnId);
+      if (sortOrder) {
+        ret.push(`sort-${sortOrder}`);
+      }
+      if (columnIsSortable(columnId)) {
+        ret.push('gallery__table-column__header--sortable');
```
ChloeSartorelli, line 781: Added for CSS purposes

client/src/containers/TableView/TableView.js
```diff
@@ -253,19 +774,34 @@ function TableView(_props) {
...
+      ret.push('gallery__table-column--modified');
     }
     return ret.join(' ');
   }
 
   /**
-   * Returns the aria-sort value for a th table header
+   * Returns the aria role for the table cell
    */
-  function getAriaSortValue(columnId) {
```
ChloeSartorelli, line 263: This got replaced with `getSortOrder` which was moved higher up to be before the functions that use it

client/src/containers/TableView/TableView.js
```diff
@@ -276,10 +812,10 @@ function TableView(_props) {
     if (columnIsSortable(header.column.id)) {
       // The "font-icon-" class is intentional. It adds the necessary CSS for adding an icon.
       // The actual icon itself is added through explicit CSS depending on how the column's sorted.
-      return <>
+      return <button onClick={() => handleSort(header)} className="btn gallery__table-column__sort-button">
```
ChloeSartorelli, line 831: Added semantically correct markup for these

client/src/containers/TableView/TableView.js
```diff
@@ -288,7 +824,25 @@ function TableView(_props) {
...
+    let editableText = rowData.canEdit ? null : `${i18n._t('AssetAdmin.NOT_EDITABLE', 'not editable')},`;
+    let uploadText = null;
+    const uploading = rowIsUploading(rowData);
+    if (uploading) {
+      editableText = null;
+      uploadText = rowHasError(rowData)
+        ? i18n._t('AssetAdmin.UPLOAD_FAILED_INSTRUCTION', 'upload failed, press enter to remove')
+        : i18n._t('AssetAdmin.UPLOADING_INSTRUCTION', 'uploading, press enter to cancel');
+      uploadText += ',';
+    }
```
ChloeSartorelli, line 853: Screen reader context, very similar to what you get in the gallery view

client/src/containers/TableView/TableView.js
```diff
@@ -349,19 +906,54 @@ function TableView(_props) {
...
+      onKeyDown={(event) => handleKeyDown(null, null, event)}
+    >
+      <span id={gridDescriptionID} className="visually-hidden">{`${paginationText}, ${thisPageText}`}</span>
       <thead>
         {table.getHeaderGroups().map(headerGroup => (
           <tr key={headerGroup.id}>
             {headerGroup.headers.map(header => (
               <th
                 key={header.id}
-                onClick={() => handleSort(header)}
```
ChloeSartorelli, line 362: Click handler moved to the button which is inside the cell

client/src/containers/TableView/TableView.scss
```diff
@@ -3,8 +3,16 @@ th.sort--disabled>span:after {
...
 .gallery__table {
+  --focus-outline-offset: var(--focus-outline-offset-inset);
+
   margin-bottom: -$panel-padding-y;
 
+  &:has(.gallery__table-row:focus-visible, .gallery__table-column:focus-visible) {
+    outline: var(--focus-outline-width) dashed var(--focus-outline-color);
+    // Add additional room for the outline of the items inside (assuming they use the same variable values).
+    outline-offset: calc(var(--focus-outline-offset) * 2 + var(--focus-outline-width));
+  }
```
ChloeSartorelli: We have to add explicit declarations for `outline` and `outline-offset` here because we're not dealing directly with `:focus-visible`

client/src/containers/TableView/TableView.scss
ChloeSartorelli, line 1: All of the checkbox-related styling here closely resembles the styling for the checkbox in the gallery view, but the selectors are so specific and the styling differs a bit so it was easier to just recreate it here.

ChloeSartorelli:
CI failure is pre-existing on the `3` branch: https://github.com/silverstripe/silverstripe-asset-admin/actions/runs/21963947888

Looks like there's a problem with the `NavigationBlocker` component in admin, after playing with it for a bit.

emteknetnz (changes requested):
I haven't reviewed code yet, so far I've just noticed several UX quirks from manual testing which will need to be resolved

- Tabbing through sortable header, though being able to then press down and go to the tab island with dash border, but then not being able to press up to get back to the header is disconcerting. The dashed outline surronds the sortable headers so it's expect arrows keys to work here. Probably include the sortable headers as part of the tab-island and give the first header tabindex=0 if no images are being edited.
- Likewise not being able to use left/right keys on the sortable header feels wrong
- The dashed border surrounds the pagination, but I cannot use the down arrow key to move to the pagination
- The left/right arrows to focus on individual cells feels pretty off because I expect the entire row to be selected, not individual cells. I think normal file manager behavior in the "details" view is for left/right to act the same as up/down and highlight the entire row, at least it does that in my file manager (dolphin). I assume you implemented this as a thing aimed at screen readers to save having to read out the entire row when you only care about a single cell.
- If I have three files (pic-a, pic-b, pic-c), and I press down arrow twice to focus on pic-c, and then shift-tab back to the sortable header, and press space to reverse sort, when i tab back to the tab-island, I am now focused on pic-a which is now in the 3rd row where pic-c used to be. I'd expect to be focused on pic-c in row 1. I also saw this behavior with a folder that was being edited (I had 2x folders abc and def). Seems like the issue is the the row index is being used instead the actual file. I don't see this behavior in the gallery view where after sort it'll auto focus on the current file whether or not it's being edited.

ChloeSartorelli:
Regarding sortable headers, before I change anything there, I just want to make sure you read the PR description where I explained why I made those choices? Just want to make sure that context is factored in since you didn't reference it in your comment.

Note that if I add them to the roving tab index, I'll also add the columns that _aren't_ interactive (can't be used to sort) and remove the ability to tab to the sortable header cells.

It would also be weird to have the full header _row_ focused, so I'm not sure what you'd want to happen if you're focused on a row and press the up arrow key... does it focus on the first header cell or the last? And wouldn't pressing down from the header cells then focus on a cell rather than the full row of the first record?

I've also needed to move the dotted outline to be around the `tbody` to avoid having it around the pagination. If the table header is part of the roving tabindex area I can't do that.

ChloeSartorelli:
> The left/right arrows to focus on individual cells feels pretty off because I expect the entire row to be selected, not individual cells. 

Please refer to the resource links in the PR description. Focusing on cells is a mandatory part of the data grid W3C pattern, and even the treegrid pattern which focuses rows like we do _requires_ that individual cells are also focusable.

ChloeSartorelli:
> If I have three files [...] and [...] reverse sort [...] I am now focused on [...] the 3rd row where pic-c used to be. I don't see this behavior in the gallery view where after sort it'll auto focus on the current file whether or not it's being edited.

I'll reset the roving tabindex on sort, and open a card to do the same for the gallery view. Sorting can bring files from other pages, so we shouldn't be retaining the roving tabindex after adjusting the sort order. Good catch.

I'll also add a behat test for that.

I don't know what you mean by "the current file whether or not it's being edited" in the gallery view but again there'll be a separate card for that.

ChloeSartorelli:
> The dashed border surrounds the pagination, but I cannot use the down arrow key to move to the pagination

I've moved the outline to only be on the `tbody` element. IMO it's a lot less obvious that it exists now and is a bit confusing with only two rows - but you're probably right that it's more correct.

emteknetnz (changes requested):
>> Sortable headers being tabbable

I re-read your PR description and it makes sense. The "what happens when you press up from the top row?" problem seems a good reason to keep the regular tab order in the header. Probably the previous dashed border impled they're part of the tab-island. Now the dashed outline is moved to `tbody` that makes things clearer

>> Cell-level left/right navigation

Fair enough - my view came from comparing it to non-web file managers, but as you say we're following web accessibility patterns here, which don't always align with native behaviors

> I don't know what you mean by "the current file whether or not it's being edited" in the gallery view but again there'll be a separate card for that.

If you have selected a file and pressed enter to edit / clicked on it. Whether or not you've done that doesn't change the behaviour.

--

Retesting this with the latest changes:

I've noticed that you can press down when focused on a sortable header and you go to the first file, but if you press up you don't go back to the sortable header. We should remove the ability to press down from the header to make it clear that it's not part of the roving tab index on the files themselves.

I've also noticed that when you use left/right to navigate between cells, you loose the dashed border around the files. Since we're still inside the "use arrow keys to navigate" section, we should retain the dashed border.

ChloeSartorelli:
> I've also noticed that when you use left/right to navigate between cells, you loose the dashed border around the files.

Fixed - accidentally broke than when I changes what element owned the outline.

> I've noticed that you can press down when focused on a sortable header and you go to the first file, but if you press up you don't go back to the sortable header. 

Fixed, that was unintentional. I do kinda like it but you're right that it's not correct behaviour.

client/src/containers/TableView/TableView.js
```diff
@@ -349,19 +922,49 @@ function TableView(_props) {
...
+      // we skip the thumbnail column as it's purely presentational
+      // we skip the checkbox column as that's activated on the row as a whole.
+      aria-colcount={4}
+      aria-multiselectable="true"
+      aria-describedby={gridDescriptionID}
+      tabIndex={-1}
+      className="gallery__table table table-hover"
+      onKeyDown={(event) => handleKeyDown(null, null, event)}
+    >
+      <span id={gridDescriptionID} className="visually-hidden">{`${paginationText}, ${thisPageText}`}</span>
```
emteknetnz:
```suggestion
      <caption id={gridDescriptionID} className="visually-hidden">{`${paginationText}, ${thisPageText}`}</caption>
```
`<span>` isn't a valid child of `<table>` so browsers may do something unexpected with it, though `<caption>` is valid use that instead

client/src/containers/TableView/TableView.js
```diff
@@ -16,11 +17,464 @@ import {
...
+          setRovingRowIndex(0);
+        } else {
+          gridRef.current.focus();
+          setDoSetFocus(false);
+        }
+      }
+    }
+
+    prevPropsRef.current = getPropsToTrack();
+  }, [props.page, props.folderId, props.files, rovingColIndex, rovingRowIndex, doSetFocus]);
```
emteknetnz:
```suggestion
  }, [
    props.page,
    props.folderId,
    props.files,
    props.openFileId,
    props.totalCount,
    rovingColIndex,
    rovingRowIndex,
    doSetFocus
  ]);
```
You also need deps that are used in getPropsToTrack() i.e. props.openFileId and props.totalCount

client/src/containers/TableView/TableView.js
emteknetnz:
```suggestion
      <caption id={gridDescriptionID} className="visually-hidden">{`${paginationText}, ${thisPageText}`}</caption>
```
`<span>` isn't a valid child of `<table>` so browsers may do something unexpected with it, though `<caption>` is valid use that instead

client/src/containers/TableView/TableView.js
```diff
@@ -16,11 +17,464 @@ import {
...
+        }
+      }
+    }
+
+    // If a file edit form gets closed, focus back inside the grid.
+    if (prevProps.openFileId && !props.openFileId) {
+      setDoSetFocus(true);
+      // If we don't have the roving index set, make sure it's set
+      // in a sensible place.
+      if (!rovingRowIndex) {
```
emteknetnz: Making this a falsey conditional makes it a little hard to know what's going on. The default value is null, it's set to null in some places, but below it's set to 0.  Is 0 and null supposed to be the same? Would be better if it were consistently set to 0 or null and then `===` here

tests/behat/src/FixtureContext.php
```diff
@@ -50,13 +50,24 @@ public function stepICheckTheGalleryItem($name)
...
         $item = $this->getGalleryItem($name);
         Assert::assertNotNull($item, "File named $name could not be found");
-        $checkboxLabel = $item->find('css', 'label.gallery-item__checkbox-label:not(.gallery-item__checkbox-label--disabled)');
-        Assert::assertNotNull($checkboxLabel, "Could not find checkbox label for file named {$name}");
-        $checkboxLabel->click();
+        $page = $this->getMainContext()->getSession()->getPage();
+        if ($page->find('css', '.gallery__table')) {
+            // On the table view we fake a checkbox, because the actual row is responsible for handling
+            // events and the screenreader doesn't need a semantic HTML checkbox.
+            $item = $item->find('xpath', "/ancestor-or-self::tr[contains(@class, 'gallery__table-row')]");
```
emteknetnz, line 57: Possibly the leading slash on the xpath means it'll search from the root of the document rather than relative to the item. Could you test this without the leading slash?

tests/behat/src/FixtureContext.php
```diff
@@ -50,13 +50,24 @@ public function stepICheckTheGalleryItem($name)
...
+            // One of the children should have this error class
+            $selectCell = $item->find('css', '.gallery__table-column--select');
```
emteknetnz, line 60: Assert::assertNotNull($item, ....) again like it does above since item is being re assigned, otherwise it'll have a hard fail rather than have a nice output if not found

client/src/containers/Gallery/Gallery.js
emteknetnz: Add that context as an inline comment

emteknetnz (changes requested):
Keyboard nav seems good now

Have left some comments on code

Tested with orca screen reader, some issues:

When an item is not selected, it will say "not selected". When I press space it will say "space", when I press down and then up to re focus the file, it will still say "not selected"

It calls the first item "row 2", it should be "row 1" or "item 1" or similar. It calls the sortable header "row 1", though this seems wrong as it's a very different type of row from a regular data row. 

When using the sortable header, it gives no indication of what direction the sort is in, it just say something like "Return" (i.e. the enter key) when I activate it

Navigating to an empty folder gives no audible feedback

ChloeSartorelli:
> When an item is not selected, it will say "not selected". When I press space it will say "space", when I press down and then up to re focus the file, it will still say "not selected"

Yeah I noticed that as well, but I couldn't find a solution to that and I seem to be using all the right attributes. If you can identify a best practice I'm not following I'll happily follow it, but right now I'd be tempted to call this a problem with Orca.

Unfortunately from what I can tell W3C doesn't currently have an example to validate against for this, though.

ChloeSartorelli:
> It calls the first item "row 2", it should be "row 1" or "item 1" or similar. It calls the sortable header "row 1", though this seems wrong as it's a very different type of row from a regular data row.

Please review the `W3C` examples which also call the header row "row 1" and the first row in the body "row 2". That's just how Orca has decided to interact with tables.

ChloeSartorelli:
> When using the sortable header, it gives no indication of what direction the sort is in, it just say something like "Return" (i.e. the enter key) when I activate it

To the best of my knowledge I have applied all the appropriate attributes which are being updated correctly. If you can identify a best practice I'm not following I'll happily follow it but as with the above this seems more like a problem with Orca.

You'll notice the same thing happens in the W3C "Example 2: Sortable Data Grid With Editable Cells" data-grid example.

ChloeSartorelli:
> Navigating to an empty folder gives no audible feedback

Resolved by giving the "no items found" p block a `role="region"`

client/src/containers/TableView/TableView.js
ChloeSartorelli: Done

client/src/containers/TableView/TableView.js
ChloeSartorelli: Oops dunno why this was falsey instead of checking `=== null` explicitly. Fixed.

tests/behat/src/FixtureContext.php
ChloeSartorelli: This was essentially copies wholesale from the existing `iShouldSeeAnErrorMessageOnTheFile()` method.

If it was searching from the root of the document, then "ancestor or self" would fail, since the document itself isn't a `tr` element, and it has no ancestors to check.

From what I can tell, a single `/` at the _start_ of an xpath will indeed search from the root of the document, but if it's in the _middle_ of an xpath (which this one is - it gets appended to the xpath of the `$item` element) it basically means "select the direct element"

e.g. `/html/body/div` means "From the root of the document, get the `html` element. Then get its direct child `body` element, and then get the direct child `div` element.

This is in contrast to a double slash which says "as a child anywhere"

e.g. `//div//span` would mean "Look for any `div` in the document, and then look for any `span` inside the `div`.

In this case I don't think it matters whether it's there or not, theoretically - but since I've copied this from a pre-existing method in this class which has been working since who knows when, I'd prefer not to mess with it.

tests/behat/src/FixtureContext.php
ChloeSartorelli: The `gallery__table-row` is explicitly part of the path to find the gallery item in the first place so this will never fail - but added to avoid ping pong.

client/src/containers/Gallery/Gallery.js
ChloeSartorelli: I think the existing comment is already the correct amount of context for a code comment.
The PR comment is just saying why this change was required as part of this PR (so the context is the PR), while the code comment is saying why the condition is required in the code (i.e. the context is the code).

That said, I'll add additional context to avoid ping pong.

client/src/containers/TableView/TableView.js
ChloeSartorelli: Done

client/src/containers/TableView/TableView.js
```diff
@@ -16,11 +17,464 @@ import {
 
 function TableView(_props) {
   const props = { ...galleryViewDefaultProps, ..._props };
+  // The roving state determines which item should have the roving tab index.
+  // Focus doesn't actually move unless doSetFocus is set to true.
+  const [rovingRowIndex, setRovingRowIndex] = useState(null);
+  const [rovingColIndex, setRovingColIndex] = useState(null);
+  const [doSetFocus, setDoSetFocus] = useState(false);
+  const gridRef = useRef([]);
```
emteknetnz:
```suggestion
  const gridRef = useRef(null);
```
May as well use null since it's for a element reference, an array is never used

client/src/containers/TableView/TableView.js
```diff
@@ -336,9 +907,11 @@ function TableView(_props) {
       cell: (info) => renderSelect(info.getValue()),
     });
   }
+  // Memoising column config prevents the table from being completely rebuilt when unrelated props/state update.
+  // Notably, this fixes issues with race conditions for screen readers when updating focus via keyboard events.
+  columnConfig = useMemo(() => columnConfig, [props.selectableItems]);
```
emteknetnz:
```suggestion
  columnConfig = useMemo(() => columnConfig, [props.selectableItems, props.selectableFolders, props.maxFilesSelect, props.selectedFiles]);
```
Those are all used in `renderSelect()`.

Also just use `const columnConfig = useMemo( all-the-things )` instead of reassigning.

client/src/containers/Gallery/Gallery.js
```diff
@@ -440,7 +440,12 @@ const Gallery = ({
    * @param {Event} event
    */
   const toggleSelectConcat = (event) => {
-    actions.gallery.setConcatenateSelect(isConcat(event));
+    // Tile view doesn't use the click-to-drag selection functionality
```
emteknetnz:
```suggestion
    // Tile view uses click-to-drag selection functionality
```
I think it's this way around?

emteknetnz (changes requested):

client/src/containers/TableView/TableView.js
ChloeSartorelli: Oops yup good call. Done.

client/src/containers/Gallery/Gallery.js
ChloeSartorelli: Have updated to say "Table view doesn't use"

client/src/containers/TableView/TableView.js
ChloeSartorelli: Adding `props.selectedFiles` as a dependency causes the same problem we're trying to avoid by memoising - it rebuilds the entire table too often. Specifically it rebuilds it any time you select any item.

Doing this with the screen reader on results in two observable problems:
1. After selecting an item with the "Space" key, there is a lag before your next key input will be processed
2. If you press "Space" to select and item and then just wait, it will re-announce the item you're focused on, because it's an entirely new DOM element so to the screen reader it thinks your focus was moved.

For that reason I haven't added that dependency, though I did add `props.maxFilesSelect` and `props.selectableFolders` (which shouldn't change in the lifetime of this component anyway) and moved the call to `useMemo` as requested.

emteknetnz: Fair enough, just add an inline comment explaining why the dependency array is the way it is and point at it's know that things like props.selectedFiles aren't included in the dependency array even though they're consumed inside the useMemo() block

ChloeSartorelli: Done

emteknetnz (approved):

ChloeSartorelli:
Was approved and is green. Merging.

PR #1673, opened 2026-02-16
FIX Resolve race condition when navigating to new folder

ChloeSartorelli:
Fixes https://github.com/silverstripe/silverstripe-installer/actions/runs/21940009833/job/63692331702
> Then I should not see the file named "file1" in the gallery  # SilverStripe\AssetAdmin\Tests\Behat\Context\FixtureContext::iShouldNotSeeTheGalleryItem()
      File named file1 was found when it should not be visible

This was caused by a race condition, where two separate effects triggered a refetch - one when the folder ID changes, and one when forceRefetch is set to true.
This meant we would attempt to both refetch the current folder as well as fetch data for the new folder. Whichever of those finished LAST would win, because the second state to be set is the state that is used to render the component.

## Issue

- https://github.com/silverstripe/.github/issues/476


ChloeSartorelli:
Yikes it's even failing when we have a dynamic wait (which it passes locally so it's not that I have the wrong selector or something like that)... I'm gonna revert back to what I was saying in https://github.com/silverstripe/.github/issues/476#issuecomment-3893438783 and explore that angle.

ChloeSartorelli:
The problem can be seen very easily by performing the following steps:

1. Create a folder in asset admin
2. Create a file in asset admin (NOT inside the folder)
3. Refresh (to be sure we have a clean slate)
4. Click the file to open the edit form
5. Type anything in the title field. Do NOT save
6. Click the x in the top right of the form
7. Wait for a few seconds, and then press "Cancel" in the dialog box
8. Repeat steps 6 and 7 again
9. Click on the folder to navigate to the folder
10. Wait for a few seconds, and then press "Ok" in the dialog box - behold the URL updated but you're still in the wrong place
11. Click on the folder again, and this time it will actually navigate.

Waiting while the dialog box is open seems important - which heavily implies this is caused by some race condition (which explains why it was sporadically failing earlier, and why adding a wait hid the problem for a while, though it doesn't explain why it's reliably failing again now).

9 times out of 10 you won't actually end up navigating to the folder, despite the URL updating.
In firefox when I do this I get a "Should not already be working" error from react (sometimes twice) but that error doesn't occur in chromium-based browsers so should be treated as a red herring.


https://github.com/user-attachments/assets/4266ea8a-a212-4239-bb21-810e5b14da38


(note that the above was after commenting out `destroyOnUnmount: false,` in `vendor/silverstripe/admin/client/src/containers/Form/Form.js` - without that your step 11 will prompt you _again_ which is a [separate issue](https://github.com/silverstripe/silverstripe-admin/issues/2116))

ChloeSartorelli:
The props in `AssetAdmin` update as expected, i.e. `folderId` is the ID of the folder you clicked on.
The network request to load the files for the new folder _does_ get fired off and returns as expected.
The STATE is still old state, though. `folder.id` will be the ID of the folder you were in (`0` by default).

ChloeSartorelli:
~~Somehow, while the `refetchFolder()` function in the `AssetAdmin` component is only run ONCE, the async function passed into the `backend.get().then()` function triggers _TWICE_.
Once with the new folder details
And then instantly with the OLD folder details, which overrides the new.~~

Okay I was wrong about this - in that `refetchFolder()` does get called twice - once with the old folder ID and once with the new.
The call for the old ID happens first, but sometimes returns last. This is where the race condition lies.

Now to figure out why it's calling twice. I suspect it's because of the blocked navigation that leads to `/admin/assets` - the navigation being blocked should have cleared it out entirely but I suspect something is remembering that the navigation was meant to happen (and doesn't get cleared when the navigation is cancelled), which is what's causing this.

client/src/containers/AssetAdmin/AssetAdmin.js
```diff
@@ -127,7 +128,7 @@ const AssetAdmin = ({
    */
   const handleBrowse = (browseFolderId, browseFileId, browseQuery) => {
     if (typeof onBrowse === 'function') {
-      const shouldForceRefetch = browseFolderId !== folderId || browseQuery !== query;
+      const shouldForceRefetch = !isEqual(browseQuery, query);
```
ChloeSartorelli, line 131: Two changes here:
1. remove `browseFolderId !== folderId` entirely. There's a whole separate effect to handle that case.
1. Do a proper deep equality check on `browseQuery` vs `query`, since e.g. `{} === {}` will return false.

tests/behat/features/file-details.feature
```diff
@@ -58,7 +58,6 @@ Feature: File details
     When I click on the file named "folder1" in the gallery
     Then I see the text "Are you sure you want to navigate away from this page?" in the alert
     When I confirm the dialog
-    And I wait for 2 seconds
```
ChloeSartorelli, line 61: We shouldn't need this anymore - this was just hiding the race condition.

client/src/containers/AssetAdmin/AssetAdmin.js
ChloeSartorelli: Note the condition was originally added in https://github.com/silverstripe/silverstripe-asset-admin/pull/1603 - but it was fine at the time because there was a single lifecycle event (`componentDidUpdate`) managing both the folderID change and the force refetch change.

Now that we're using a functional component with two separate effects we need to be more careful about changing both of those values at the same time.

emteknetnz (changes requested):
Code looks fine, just need to change the commit message from MNT to either FIX or ENH

emteknetnz (approved):

Issue #1655, opened 2026-01-26
Breadcrumbs in asset admin are invalid links


ChloeSartorelli:
In the asset admin (`/admin/assets`), the breadcrumb links are invalid.

Left clicking works as expected because of react event handlers, but middle mouse clicking or selecting "open link in new tab" in the context menu takes you to an invalid URL. For example for the `Files` breadcrumb the URL is `assets` instead of the relative `/admin/assets`.

Found in the `3` branch, haven't tested if it's in `3.1` or `2.4` yet.

## Kitchen sink CI (admin PR, not asset-admin PR)
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/22375158556

## PRs (choose one and close the other)
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1675
- https://github.com/silverstripe/silverstripe-admin/pull/2131

ChloeSartorelli:
PR(s) merged. This will be automatically tagged by GitHub Actions.

Reassigned to Steve to make sure this will work in CMS 6.2 in the context of [the work to deprecate base tag](https://github.com/silverstripe/silverstripe-admin/pull/2112) - i.e:
1. does it work as expected with the base tag removed?
2. does it work as expected with/without the base tag when the app is installed in a subfolder of the web root?

emteknetnz:
Please raise a new card for that, I'm not going to do CMS 7 work on this issue right now.

I though we'd drop support for 2. as well

ChloeSartorelli:
This has nothing to do with CMS 7, it has to do with new configuration that is added in CMS 6.2

emteknetnz:
OK ... didn't even know this was a thing. Please raise a new card

ChloeSartorelli:
I guess I'll validate it then. It's not a new card, it's part of this card. But it's not worth arguing about.

ChloeSartorelli:
Merged up and resolved conflicts. Confirmed works with both valid values of the new base tag config.

PR #1675, opened 2026-02-24
FIX Breadcrumb links in asset admin

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1655


ChloeSartorelli (commented):
If the react route path is incorrect, shouldn't we fix that in PHP instead of adjusting it in multiple places in JavaScript code?

emteknetnz:
Replaced by https://github.com/silverstripe/silverstripe-admin/pull/2131

Issue #1667, opened 2026-02-09
In gallery view, focus swaps too soon when navigating folders with keyboard


ChloeSartorelli:
https://github.com/silverstripe/silverstripe-asset-admin/pull/1643 added new keyboard navigation behaviour to the grid view.

Part of this is that when you navigate to a new folder (either by pressing enter on a folder, or by pressing enter on the "navigate up a level" button), focus moves to the first item in the grid.

Unfortunately the focus changes _before_ navigation occurs. This results in focusing on the first item in the grid, _then_ navigating to the new folder.

Visually this presents as a slightly odd focus jump, but there's a worse problem with screen readers which will announce the item that got focused first - the item from the page you're not on anymore.

## Recommended solution

1. Delay focus until after the navigation has completed (not sure how to do that but that's important to do)
2. Consider focusing on the _grid_ rather than focusing on the first _item_ in the grid. This might be a bit odd visually, but it gives the benefit that the screen reader will announce the updated grid information, such as pagination details and how many folders/files are on this page.

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1676

ChloeSartorelli:
It seems that this bug stems from an order of operations problem. The lifecycle looks something like this:

1. User interacts with the app to trigger a navigation
2. React router state is updated to change the folder ID, which updates the `folderId` prop in `AssetAdmin` (and its children)
3. 
    - `AssetAdmin`'s `componentDidUpdate` fires, which notices the change in folder ID and forces a refetch of the folder
    - In the same cycle, `ThumbnailView`'s `componentDidUpdate()` fires, which notices the change in folder ID and resets refs etc.
4. The refetch request is still pending. Meanwhile focus gets set on the first item in the grid
5. The refetch request finishes and the grid content gets updated.
6. Focus gets moved to the new first item in the grid (though the screen reader doesn't announce this, since no user interaction forced the change in focus and it was still busy reading the previously focused item)

Ultimately that means this part of the `componentDidUpdate` function isn't actually doing what we expect it to:
https://github.com/silverstripe/silverstripe-asset-admin/blob/cccbd20142e68c273591134c95df43e970847423/client/src/containers/ThumbnailView/ThumbnailView.js#L54-L62

It is technically empting those ref arrays, but since we still have the same files list they just get filled with the exact same refs again.

ChloeSartorelli:
[...]

PR #1676, opened 2026-02-26
FIX Defer updating focus until navigation is complete

ChloeSartorelli:
When navigating to a new page or folder, we need to defer updating the focus state in the gallery view until after navigation is complete, or we'll temporarily focus on the wrong item. This is visually weird and a bad screen reader experience.

Also opted to focus on the full grid rather than just the first item, matching how the table view works.

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1667


tests/behat/features/accessibility-asset-admin.feature
```diff
@@ -214,7 +214,7 @@ Feature: Accessibility asset-admin
     When I press the "Enter" key globally
     Then I should not see the file named "invalid" in the gallery
     And the ".gallery__main-view--tile" element should have focus
-    # Navigating back to the folder doens't retain the old tabindex
+    # Navigating back to the folder doesn't retain the old tabindex
```
ChloeSartorelli, line 217: Unrelated typo fix

tests/behat/features/accessibility-asset-admin.feature
```diff
@@ -232,8 +232,9 @@ Feature: Accessibility asset-admin
     # Clicking to select an item sets the roving tabindex to the clicked item
     When I check the file named "file03" in the gallery
     Then the file named "file03" should have focus
-    When I press the "Arrow_Right" key globally
-    Then the file named "file04" should have focus
+    When I press the "Tab" key globally
+    And I press the "Shift-Tab" key globally
+    Then the file named "file03" should have focus
```
ChloeSartorelli, line 237: I'm not sure why, but for some reason this PR started failing behat in this spot (see https://github.com/silverstripe/silverstripe-asset-admin/actions/runs/22423861164/job/64927586367?pr=1676)

I couldn't reproduce locally, but thinking about it this was testing the wrong thing anyway - what's important is where the roving tabindex is, so checking that file03 actually has the roving tabindex is what matters here. Turns out it _didn't_ have it.

client/src/components/GalleryItem/GalleryItem.js
```diff
@@ -518,7 +518,13 @@ const GalleryItem = (_props) => {
...
   if (selectable) {
     if (canBatchSelect()) {
-      action = handleSelect;
+      action = (event) => {
+        if (typeof onClick === 'function') {
+          // Needed to make sure clicking the checkbox updates the roving tabindex
+          onClick(event);
+        }
+        handleSelect(event);
+      };
```
ChloeSartorelli, line 527: This change is so that the item which is selected with the mouse has roving tabindex (see note in the behat feature)
Ideally this would have been in a separate PR but the behat test test was failing with this PR and not without it, so I needed to fix it here.

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -54,12 +55,21 @@ class ThumbnailView extends Component {
...
+        if (!this.state.forceResetRefs) {
+          this.setState({ forceResetRefs: true });
+        }
+        return;
+      }
       this.gallerySizeRef.current = null;
       this.folderRefs.current = [];
       this.fileRefs.current = [];
+      // Explicitly focus on the grid itself to announce changes to page etc.
+      this.gridRef.current.focus();
```
ChloeSartorelli, line 70: Matches behaviour in the table view, and when there's no items in the gallery.

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -54,12 +55,21 @@ class ThumbnailView extends Component {
...
     // If we changed page or are looking at a different folder, throw away the old refs
     // and reset focus, and skip the rest of the logic in this lifecycle event.
-    if (oldProps.page !== this.props.page || oldProps.folderId !== this.props.folderId) {
+    if (this.state.forceResetRefs || oldProps.page !== this.props.page || oldProps.folderId !== this.props.folderId) {
+      // If the files arrays are still identical, the navigation hasn't finished yet, so defer changes for now.
+      if (this.fileArraysAreIdentical(oldProps.files, this.props.files)) {
+        if (!this.state.forceResetRefs) {
+          this.setState({ forceResetRefs: true });
+        }
+        return;
```
emteknetnz, line 64: Seems like the return here could things to not progress past this point under certain conditions?

client/src/containers/ThumbnailView/ThumbnailView.js
emteknetnz: Should we check if this.gridRef.current is truthy first?

emteknetnz (changes requested):

client/src/containers/ThumbnailView/ThumbnailView.js
ChloeSartorelli: No need to - this lifecycle event happens after rendering, so we'll always have a gridref by this point.

client/src/containers/ThumbnailView/ThumbnailView.js
ChloeSartorelli: Yes, which is exactly what it's there for. This skips most of this lifecycle event already by virtue of being in the condition covered by the "and skip the rest of the logic in this lifecycle event." comment.

If we hit this part of the code we also need to skip the part immediately following it - that's what we're deferring for later. If we didn't return, we'd just end up right where we were before this PR.

emteknetnz: I think the return statement might be in the wrong place? Should it be in the `if (!this.state.forceResetRefs) {` block?

Otherwise you might end in a state where either page or the folderId prop changed, `this.fileArraysAreIdentical(oldProps.files, this.props.files)` is true, but `this.state.forceResetRefs` is `false` (so it's not set to `true`).  Then execution stops and because forceResetRefs is never set to true?

emteknetnz (changes requested):

client/src/containers/ThumbnailView/ThumbnailView.js
ChloeSartorelli: If `this.fileArraysAreIdentical(oldProps.files, this.props.files)` resolves true, we should return, because we're deferring. That condition determines if we should defer or not.

So the defer process is like this:

1. The `page` or `folderId` prop has changed, signalling that we're going to navigate.
2. The `files` prop has _not_ changed, signalling that we haven't finished navigating yet.
3. If we haven't already set `forceResetRefs` to true, it gets set true.
4. We return out of the life cycle event
5. Eventually the `files` prop changes, and the lifecycle event fires again because it's updating the component.
6. `forceResetRefs` is true, signalling that a navigation occurred.
7. Since the `files` prop changed, we don't defer this time.
8. We reset all the refs, focus on the grid, set the focused item to null and (importantly) set `forceResetRefs` back to false.

> you might end in a state where either page or the folderId prop changed, this.fileArraysAreIdentical(oldProps.files, this.props.files) is true, but this.state.forceResetRefs is false (so it's not set to true)

In this case we've gone through 1 and 2 above, and since `forceResetRefs` is currently false it gets set to true (3) and then we return (4), and it continues through from there. Unless I've misunderstood something, what you're describing is exactly the intended flow of events.

emteknetnz (changes requested):
Merge conflict

ChloeSartorelli:
Rebased

emteknetnz (approved):

Issue #1669, opened 2026-02-09
Gallery view focus doesn't move to new folder when added


ChloeSartorelli:
When adding new items to the grid, this should move focus. It works correctly with uploads, but in the gallery view focus does _not_ move to the new folder in the grid when a new folder is created.

One of these should happen:

1. Focus stays in the form but the roving tabindex is moved to the folder, so the next time we focus on the grid the folder will have focus, or
1. Focus and roving tabindex both immediately move to the folder

Probably the first option is more appropriate.

Edit: Looks like this is only a problem when we're not on the last page - tracking `totalCount` instead of `files.length` will resolve the main problem here. That said, it would still be best if focus didn't move (i.e. only move the roving tab index)

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1677

PR #1677, opened 2026-02-26
FIX Ensure focus moves to folder when adding new folder.

ChloeSartorelli:
The issue says "it would still be best if focus didn't move (i.e. only move the roving tab index)" - however unfortunately due to the fact that the `GalleryItem` components are responsible for setting focus themselves, and that they get this directive _before_ the state switch for `allowedToSetFocus` being set false, I couldn't reliably change it to stop focusing on the folder after creation without it instead focusing on whatever item was previously focused.

This could be resolved by completely redoing the way focus is handled so that the `ThumbnailView` component handles focusing on the elements instead, passing down a ref so it knows what element to focus on - and in fact my original implementation of all of this work originally did things that way, but it got complex and had its own timing issues. It's a big can of worms that isn't worth opening IMO.

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1669


client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -117,20 +117,22 @@ class ThumbnailView extends Component {
...
-      this.setState({
-        allowedToSetFocus: true,
-        focusedItem: this.getFocusDataFromItem(newItem)
-      });
+      if (newItem) {
+        this.setState({
+          allowedToSetFocus: true,
+          focusedItem: this.getFocusDataFromItem(newItem)
+        });
+      }
```
ChloeSartorelli, line 140: New defensive programming as this PR uncovered a scenario where new item was undefined

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -176,6 +178,9 @@ class ThumbnailView extends Component {
    * @param {object} item A file or folder from props.files
    */
   getFocusDataFromItem(item) {
+    if (!item) {
+      return null;
+    }
```
ChloeSartorelli, line 202: New defensive programming as this PR uncovered a scenario where item was undefined

emteknetnz (approved):

emteknetnz (changes requested):
Is a behat test here possible? Would probably want an extension to limit the number of items per page so can test on both first page and last page.

ChloeSartorelli:
It would be easier if we didn't have to do all accessibility behat testing in the same feature file, but yes it is doable. Will add.

emteknetnz (changes requested):
Behat failure

ChloeSartorelli:
Yes, it's still assigned to me lol

client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -68,13 +68,15 @@ class ThumbnailView extends Component {
       this.fileRefs.current = [];
       // Explicitly focus on the grid itself to announce changes to page etc.
       this.gridRef.current.focus();
+      const newState = { forceResetRefs: false };
       if (this.state.focusedItem) {
-        this.setState({ focusedItem: null, forceResetRefs: false });
+        newState.focusedItem = null;
       }
+      this.setState(newState);
```
ChloeSartorelli, line 75: Missed this in a previous PR - if there is no focused item, we weren't setting `forceResetRefs` back to false. We *must* set this back to false every time, so that we don't go back to deferring a reset. Otherwise when we add or remove an item we'll reset focus state instead of focusing on the correct item.

This was causing a behat failure in this PR so needs to be fixed here.

emteknetnz (approved):

Issue #1668, opened 2026-02-09
Gallery view focus race condition when uploading new files


ChloeSartorelli:
When uploading a new file, focus moves to the temporary rendering of the file. Once it has finished uploading, two things happen:

1. If no edit form was already open, we navigate to the edit form for that file. **The first input field gets focus.**
2. The temporary rendering of the file is replaced with one from the server (since we now have a database entry for it). The file is sorted appropriately (unless it belongs on a different page, then it stays put). **If it still had focus, focus moves to the new element representing the file.**

There's some network requests included in this flow, so the order of what happens differs a bit depending on which request finishes first. This means it's not predictable which item will get focus - the form field or the item in the grid.

## Proposed solution

If no edit form is open, do not attempt to move focus onto the uploaded file in the grid after upload is complete. This means when the edit form opens for this file we only have one focus event - focusing on the first input in the form.
Note however that the roving tabindex _does_ need to be updated.

If an edit form is already open when the upload is complete, move focus to the new element representing the item in the grid. This will again be the only focus event, and allows users to immediately open the form if they want to by pressing enter.

## PRs

- https://github.com/silverstripe/silverstripe-asset-admin/pull/1678

PR #1678, opened 2026-02-26
FIX Avoid focus race condition when opening form for new file

ChloeSartorelli:
There was a race condition between setting focus on the newly-opened form and setting it on the newly-sorted file in the gallery. The form should always win this race.

The easiest way to reproduce the problem is to throttle your network speed in the browser developer tools. I found "regular 3G" to be a good setting for this.

## Issue

- https://github.com/silverstripe/silverstripe-asset-admin/issues/1668


client/src/containers/ThumbnailView/ThumbnailView.js
```diff
@@ -133,7 +133,14 @@ class ThumbnailView extends Component {
     if (this.state.focusedItem?.queuedId && !this.state.focusedItem.id && this.props.files.length === oldProps.files.length) {
       const current = this.props.files.find((item) => this.focusItemsAreIdentical(this.state.focusedItem, item));
       if (current && current.id !== this.state.focusedItem.id) {
-        this.setState({ focusedItem: this.getFocusDataFromItem(current) });
+        const newState = { focusedItem: this.getFocusDataFromItem(current) };
+        // It is important that focus change only happens if a different file already has its form open.
+        // Otherwise, we have a race condition between moving focus into the newly opened form for the new item
+        // and moving focus with the newly uploaded file.
```
emteknetnz:
```suggestion
        // and moving focus to the newly uploaded file.
```
I think this is what it's supposed to be?

emteknetnz (changes requested):

client/src/containers/ThumbnailView/ThumbnailView.js
ChloeSartorelli: "with" is correct. The newly uploaded file already has focus when we hit this bit of code - but that file is moving where it is in the grid and its key is changing because now it's stored in the database rather than just being in the DOM. This means the key is now its database ID and the queued ID, not just the queued ID, which means react builds a new DOM element for it.

In order to keep focus on the same file, we need to move focus _with_ the file - i.e. when the file moves, focus needs to move as well.

I'll update this to say something like "keeping focus on the newly uploaded file" instead of "moving focus with the newly uploaded file" to remove this confusion.

ChloeSartorelli: Also updated the same comment in `TableView.js`.

emteknetnz (approved):

Issue #1692, opened 2026-07-20
Basic image editing


emteknetnz:
Create a basic editing feature in the CMS

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1693
- https://github.com/silverstripe/developer-docs/pull/895

blueo:
Follow-up docs PR aligning the image editor documentation with the developer-docs writing style: silverstripe/developer-docs#896 (tone/clarity only, on top of the original docs PR silverstripe/developer-docs#895).

emteknetnz:
All merged

PR #1693, opened 2026-07-21
NEW Basic image editing

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1692


client/src/containers/ImageEditorModal/ImageEditorModal.js
```diff
@@ -0,0 +1,1027 @@
...
+    ? { width: parseDimension(widthField), height: parseDimension(heightField) }
+    : null;
+
+  // Route validation errors to the polite live region as well as the red field.
+  useEffect(() => {
+    const message = widthError || heightError;
+    if (message) {
+      setAnnouncement(message);
+    }
+  }, [widthError, heightError]);
```
blueo, line 448: Minor a11y: this effect only ever *sets* the polite live-region announcement and never clears it. On open there's a one-render window where `dimensionsKnown` has become true but `widthField` is still `''` (the `seedResizeFields` effect hasn't run yet), so `widthError` is transiently truthy and this announces *"Enter a whole number of pixels, 1 or more"* even though the fields are valid and Apply is enabled. I confirmed the stale text is left sitting in the live region after opening the editor in a live test. Suggest `setAnnouncement(message || '')` so it resets to empty when valid again.

client/src/containers/ImageEditorModal/ImageEditorModal.js
```diff
@@ -0,0 +1,1027 @@
...
+          color="secondary"
+          onClick={handleCancel}
+          disabled={loading}
+        >
+          {i18n._t('AssetAdmin.IMAGE_EDITOR_CANCEL', 'Cancel')}
+        </Button>
+        <Button
+          color="primary"
+          onClick={handleApply}
+          disabled={loading || !cropValid || !resizeValid}
```
blueo: Apply is enabled even when nothing has changed — `hasTransform` gates the Reset button just above but isn't applied here. Because the backup checkbox defaults to on, opening the editor and clicking Apply with no edits silently re-encodes the image (lossy for JPEG) and writes a `-vN` backup copy for a pure no-op. Consider `disabled={loading || !hasTransform || !cropValid || !resizeValid}` to mirror Reset. (No behat scenario covers the zero-transform Apply path.)

code/Controller/AssetAdmin.php
```diff
@@ -551,6 +569,86 @@ public function apiCreateFile(HTTPRequest $request): HTTPResponse
...
+        }
+        // The edit replaces the source's own bytes, so canView() is not enough.
+        if (!$file->canEdit()) {
+            $this->jsonError(403);
+        }
+
+        // Permission to create the backup copy in the source's folder.
+        $folder = $file->ParentID ? $file->Parent() : null;
+        $member = Security::getCurrentUser();
+        if (!Image::singleton()->canCreate($member, ['Parent' => $folder])) {
```
blueo: This create-in-folder check runs unconditionally, even when the request sets `backupOriginal: false` and no backup will be written. The effect is that a member who can *edit* an image but can't *create* in its folder can't use the editor at all. Same unconditional check in `FileFormFactory::getEditImageAction()` (~line 448). It's intentional and documented, but you could gate it on `backupOriginal` so declining the backup doesn't require create rights.

code/Controller/AssetAdmin.php
```diff
@@ -754,6 +852,18 @@ protected function getPostedJsonValue(HTTPRequest $request, string $key): mixed
         return $value;
     }
 
+    /**
+     * As getPostedJsonValue(), but returns null for an absent key rather than rejecting the request.
+     */
+    protected function getOptionalPostedJsonValue(HTTPRequest $request, string $key): mixed
+    {
+        $data = json_decode($request->getBody() ?? '', true);
```
blueo, line 863: Micro nit: `apiEditImage()` reads the payload via the get*PostedJsonValue helpers six times, and each call re-runs `json_decode($request->getBody())`. Negligible for a payload this small, but decoding once and reading keys off the array would avoid the repeated parse. (Pre-existing pattern, so take it or leave it.)

code/Model/ImageEditor.php
```diff
@@ -0,0 +1,348 @@
...
+     * Apply flip -> rotate -> crop -> resize in canonical order to a clone of the source resource
+     * and return the encoded bytes at the platform's configured image quality.
+     */
+    private function renderImage(Image $source, array $transforms): string
+    {
+        $backend = $source->getImageBackend();
+        if (!$backend instanceof Image_Backend) {
+            throw new RuntimeException('The image could not be rendered');
+        }
+        $resource = $backend->getImageResource();
```
blueo, line 240: `getIsImage()` accepts GIF, so animated GIFs reach the renderer and will be flattened to a single frame on re-encode under the GD driver (the class docblock already notes EXIF/ICC/XMP loss under GD). Might be worth either excluding GIF up front, or a doc note that animation is dropped, so authors aren't surprised.

code/Model/ImageEditor.php
```diff
@@ -0,0 +1,348 @@
...
+        $resource->crop($width, $height, $left, $top);
+
+        // Resize last so it only changes the size, never re-frames the box the author drew.
+        if ($transforms['resize'] !== null) {
+            [$resizeWidth, $resizeHeight] = $this->calculateResizeSize($width, $height, $transforms['resize']);
+            $resource->resize($resizeWidth, $resizeHeight);
+        }
+
+        return (string) $resource
+            ->encodeByExtension($source->getExtension(), quality: $backend->getQuality())
```
blueo: My main thought here is that its relatively difficult to extend or customise the transformations. It would be great if there was an option to tweak these as they went through especially the quality setting.  

Auto generated comments on this below:

The render pipeline is effectively sealed for customisation: `renderImage()` and its helpers are all `private`, and the only extension hooks (`onBefore/AfterReplaceOriginal` etc.) fire on bytes that are already rendered *and* encoded. That makes tweaking the render itself awkward — output quality here comes straight from `$backend->getQuality()`, so a dev who wants editor-specific quality either changes the backend quality **globally** (which also affects every resampled thumbnail) or re-encodes the finished bytes in `onBeforeReplaceOriginal`, a second lossy encode. Adding or swapping a transform (sharpen, a different resize filter) has no hook at all; the only route is an `Injector` class swap, but because the methods are `private` a subclass can't override a single step without reimplementing the public `editImage()`.

Could we open the pipeline up a little without exposing the validated core? A few small, additive options, roughly in order of value:

1. **A hook just before the encode**, exposing the rendered resource, the transforms, and the quality (by reference — `Extensible::extend()` passes args by ref). This covers both "change quality" and "add a transform" with no subclass and no global change:
   ```php
   $quality = static::config()->get('output_quality') ?? $backend->getQuality();
   $this->extend('updateRenderedImage', $resource, $transforms, $quality);
   return (string) $resource
       ->encodeByExtension($source->getExtension(), quality: $quality)
       ->toString();
   ```
2. **An `output_quality` config** (`null` → fall back to the backend) so editor quality is a discoverable one-liner rather than a global backend change.
3. **`protected` instead of `private`** on the pipeline methods, so an `Injector`-swapped subclass can override a single step as the escape hatch for anything deeper.

A dev could then do either of these, no class swap and nothing global:
```php
protected function updateRenderedImage(ImageInterface $resource, array $transforms, int &$quality): void
{
    $resource->sharpen(10); // add a transform
    $quality = 90;          // or set editor-specific quality
}
```
```yml
SilverStripe\AssetAdmin\Model\ImageEditor:
  output_quality: 90
```

blueo (changes requested):
generally really solid and works well. 

I've commented on extensibility on the ImageEditor class. 

usability wise: I found myself trying to drag the crop around before I had selected the tool which meant the image seemed to 'move'. Maybe activate the crop tool by default and/or auto select it when you change the aspect ratio dropdown. It could also help to mark the image as not clickable/selectable when the crop isn't active (maybe like https://stackoverflow.com/questions/55325687/how-can-i-disable-the-image-dragging-efect)

a couple of minor things too in the comments

emteknetnz:
>I found myself trying to drag the crop around before I had selected the tool which meant the image seemed to 'move'. Maybe activate the crop tool by default and/or auto select it when you change the aspect ratio dropdown. It could also help to mark the image as not clickable/selectable when the crop isn't active

I've made it it so that the click+drag on the image when the image when crop isn't selected does nothing

I've also made it so that changing the value of the crop drowndown aspect ratio will activate the crop transform

client/src/containers/ImageEditorModal/ImageEditorModal.js
emteknetnz: Done

client/src/containers/ImageEditorModal/ImageEditorModal.js
emteknetnz: Done, you now need to have at least one transform before apply button is active, and if crop is the only transform activated then it needs to be a smaller size that the original image

code/Controller/AssetAdmin.php
emteknetnz: I'll added the check for `backupOriginal` in AssetAdmin simply for correctness, though blocking the ability to open the modal on canCreate as is, otherwise you might end up in a situation where the user can edit the image, but not create a backup. This means that if they mess up the edit they have no way to restore it, unlike the default behaviour which will create a backup.

I think it's better to simply block them from editing the image (in `FileFormFactory::getEditImageAction()`) in this scenario otherwise the editor becomes a potential footgun because we'd essentially need to remove the 'create backup image' checkbox (other option is have the checkbox greyed out since they cannot activate it, which would be a frustrating UX experience)

code/Controller/AssetAdmin.php
emteknetnz: Leaving as is, any performance increase would be virtually non-existant so not worth adding complexity to optimise this

code/Model/ImageEditor.php
emteknetnz: Have added a new "Metadata and animation" section in the developer-docs saying the GD driver flattens an animated GIF to its first frame

I don't want to exclude animated gifs as you can transform them just fine using imagemagick which plenty of setups use

code/Model/ImageEditor.php
emteknetnz: Have gone for option 2 and added config to set the output quality

code/Controller/AssetAdmin.php
blueo: ok, seems reasonable to keep it simple ✅

blueo (approved):

PR #1695, opened 2026-08-19
TLN Update translations

emteknetnz:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)


client/lang/en.js
```diff
@@ -78,6 +78,41 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "AssetAdmin.FILTER_DATE_DESC": "newest",
     "AssetAdmin.FILTER_TITLE_ASC": "title a-z",
     "AssetAdmin.FILTER_TITLE_DESC": "title z-a",
+    "AssetAdmin.IMAGE_EDITOR_APPLY": "Apply",
```
emteknetnz, line 81: These don't technically need to be here because this feature isn't released until CMS 6.3, but it's also harmless

blueo (approved):
checked the Italian change and looks intentional. As mentioned editor ones are superfluous but not going to break anything

